PR #2, opened 2021-11-16
initial mono repo context

romainmenke:
- add config files for various tools
- add github workflows (these might fail until migration is done)

Notes :

`.github/bin/install-and-test-all-packages.sh`

This compensates for a lack of workspaces support in node 12 and 14.


.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
+name: test
+on:
+  push:
+
+jobs:
+  test_with_workspaces_support:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # All currently supported LTS versions + latest with workspaces support
```
Antonio-Laguna, line 10: Consider getting rid of this comment given it's a tad misleading with the same comment being below. Could you add some quick explanation on what this `--workspaces` means to the PR?

.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+      - run: npm run build --workspaces --if-present
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+
+      - run: bash ./.github/bin/install-and-test-all-packages.sh
+
+  # Run cross platform test of the basic CLI.
+  test_cli:
```
Antonio-Laguna, line 45: 🧑🏼‍🍳💋

package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@csstools/postcss-preset-env-mono-repo",
```
Antonio-Laguna, line 2: probably needs to update the name to be postcss-plugins

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "typescript": "^4.4.4"
+  },
+  "scripts": {
+    "build": "npm run build --workspaces --if-present",
+    "test": "npm run test --workspaces --if-present"
+  },
+  "volta": {
+    "node": "16.13.0",
+    "npm": "8.1.3"
+  }
```
Antonio-Laguna, line 35: Love this, can we use the engines too and nvmrc too?

Antonio-Laguna (commented):
Nice work as usual @romainmenke 

Left some comments. I also see a bunch of check annotations below.  Mostly with semicolons which consistently are left over on csstools plugins. Can you update eslint to handle this?

.github/workflows/test.yml
romainmenke: I'll remove both comments (they are not entirely the same).
But I'll add a `MAINTENANCE.md` doc to explain how these files should be updated when future node versions are released.

Once we only support 16,18,20 they will be different and simpler.

romainmenke: Any command run with `--workspaces` will run the command against individual packages but in a predictable order.

order is the same as the order in which workspaces are declared :
1. packages
2. plugins
3. plugin-packs

This ensures that dependencies for plugins are build first and that all plugins are build before handling `postcss-preset-env`

`--if-present` is added because some things might not have a script.
But I think it would be better to remove those flags in the future.
With them it is possible to forget `scripts.build` in a `package.json` and it would go undetected in CI.

package.json
romainmenke: Yes!

package.json
romainmenke: Indeed :)

romainmenke:
>Left some comments. I also see a bunch of check annotations below. Mostly with semicolons which consistently are left over on csstools plugins. Can you update eslint to handle this?

Maybe I am pushing my own opinions too much here :)
I always configure eslint to enforce `;` as I think it improves readability.
Improving readability speeds up refactors and review.

I will defer to your judgment here.

Any linting issues will be resolved later once initial setup is complete.
At the moment too many packages have different configs.

Antonio-Laguna (approved):

PR #3, opened 2021-11-16
shared rollup config

romainmenke:
- add a shared rollup config for both javascript and typescript.
- this should support `esm`, `cjs` and cli's.

Notes :

1) Some dependencies have circular imports and rollup nags about these.
In practice they seem to work fine so I silence them after testing.

2) Typescript support is added early here as I think it is an important tool.
There is no plan to migrate everything to Typescript now. It should just be possible to use in this repo.

### DIFF
--- a/rollup/default.js
+++ b/rollup/default.js
@@ -1,0 +1,91 @@
+import babel from '@rollup/plugin-babel';
+import commonjs from '@rollup/plugin-commonjs';
+import path from 'path';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import { terser } from 'rollup-plugin-terser';
+
+export default [
+	{
+		input: 'src/index.js',
+		output: [
+			{ file: 'dist/index.cjs', format: 'cjs', sourcemap: true, exports: 'auto' },
+			{ file: 'dist/index.mjs', format: 'esm', sourcemap: true, exports: 'auto' },
+		],
+		onwarn: (warning) => {
+			// Silence circular dependency warning for postcss-values-parsers package
+			if (
+				warning.code === 'CIRCULAR_DEPENDENCY' &&
+				warning.importer.indexOf('node_modules/postcss-values-parser/lib') > -1
+			) {
+				return;
+			}
+
+			console.warn(`(!) ${warning.message}`);
+		},
+		plugins: [
+			babel({
+				babelHelpers: 'bundled',
+				exclude: 'node_modules/**',
+				presets: [
+					['@babel/preset-env', {
+						corejs: 3,
+						loose: true,
+						modules: false,
+						targets: { node: 12 },
+						useBuiltIns: 'usage',
+					}],
+				],
+			}),
+			terser(),
+		],
+	},
+	{
+		input: 'src/cli.js',
+		output: [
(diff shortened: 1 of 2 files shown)


rollup/default.js
```diff
@@ -0,0 +1,91 @@
...
+		plugins: [
+			babel({
+				babelHelpers: 'bundled',
+				exclude: 'node_modules/**',
+				presets: [
+					['@babel/preset-env', {
+						corejs: 3,
+						loose: true,
+						modules: false,
+						targets: { node: 12 },
```
Antonio-Laguna, line 34: Are targets correct here? The current LTS is 16

rollup/default.js
```diff
@@ -0,0 +1,91 @@
...
+			terser(),
+		],
+	},
+	{
+		input: 'src/cli.js',
+		output: [
+			{ file: 'dist/cli.mjs', format: 'esm', sourcemap: false },
+		],
+		onwarn: (warning) => {
+			// Silence circular dependency warning for postcss-values-parsers package
```
Antonio-Laguna, line 48: Can you outline where the circular dependency is? I'd love to get rid of this workaround if doable

rollup/default.ts.js
```diff
@@ -0,0 +1,86 @@
+import babel from '@rollup/plugin-babel';
+import commonjs from '@rollup/plugin-commonjs';
+import path from 'path';
+import typescript from '@rollup/plugin-typescript';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import { terser } from 'rollup-plugin-terser';
+
+export default [
+	{
+		input: 'src/index.ts',
```
Antonio-Laguna, line 10: Is there a way for us to reuse all of this that's repeated here so it's common and easier to maintain?

Antonio-Laguna (commented):
This is great and I love you've opened the door to TS globally :) 

Some comments, happy to discuss. Most of this would clash with `;` for example on eslint

rollup/default.ts.js
romainmenke: Definitely but it might be too early for that.
I first want to see that these configs hold up for every plugin.
Only did some spot tests so I don't yet have a good view of all edge cases.

Having everything explicit makes it easier now to handle the unexpected.

I'll make an issue to clean this up later.

rollup/default.js
romainmenke: I think so, but now starting to doubt it.
I thought we declare node 12 here as that is the lowest version we actively support.
Otherwise the build output might not work in node 12|14

https://babeljs.io/docs/en/babel-preset-env#targetsnode

rollup/default.js
romainmenke: I opened an issue to outline this later once the general setup is complete : https://github.com/csstools/postcss-plugins/issues/8

At the moment I can't run it to trigger the warning again.

rollup/default.js
Antonio-Laguna: Makes sense

rollup/default.js
Antonio-Laguna: Thanks for doing that!

rollup/default.ts.js
Antonio-Laguna: Perfect!

PR #5, opened 2021-11-16
base cli

romainmenke:
- add a shared package to build cli's
- heavily inspired by [postcss-cli](https://github.com/postcss/postcss-cli) (not internally, but in usage as a CLI)

```ts
import plugin from '@csstools/postcss-base-plugin';
import { cli, helpTextLogger } from '@csstools/base-cli';

cli(
	plugin,
	['color', 'another_option'],
	helpTextLogger(
		'postcss-base-plugin-cli',
		'Base Plugin',
		'An example plugin CLI',
		{
			color: 'A CSS color',
			another_option: true,
		},
	),
);
```


packages/base-cli/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+
+			return writeFile(output, result.css + (argo.inlineMap ? '\n' : '')).then(() => {
+				console.log(`CSS was written to "${path.normalize(output)}"`);
+			});
+		});
+	})).catch((error) => {
+		console.error(argo.debug ? error : error.message);
+
+		process.exit(1);
+	});
```
romainmenke, line 150: I am not really happy that we have all this duplication here.

But each case is very different and splitting it this way was easiest to have no bugs.

packages/base-cli/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+			return result.css;
+		}).then((result) => {
+			process.stdout.write(result + (argo.inlineMap ? '\n' : ''));
+
+			process.exit(0);
+		}).catch((error) => {
+			console.error(argo.debug ? error : error.message);
+
+			process.exit(1);
+		});
```
Antonio-Laguna, line 40: I'm not strongly against promises but I reckon that given we already return a promise, it'd be more legible if we go for `async`/`await`?

packages/base-cli/src/index.ts
```diff
@@ -0,0 +1,212 @@
...
+		let output = argo.output;
+		if (!output && argo.outputDir) {
+			output = path.join(argo.outputDir, 'output.css');
+		}
+
+		return getStdin().then((css) => {
+			if (!css) {
+				helpLogger();
+				process.exit(1);
+			}
```
Antonio-Laguna, line 53: All of this logic feels rather similar to the above, should we extract or is this a ca see of keeping this as is to see where it could deviate?

Antonio-Laguna: Just read your comment below, disregard this one. However I feel like some of the things could be extracted into their own functions?

packages/base-cli/src/index.ts
romainmenke: `async`/`await` would be much better!
I haven't focussed on this aspect and got caught up in the argument parsing.

Thank you for mentioning it!

romainmenke: Switched too `async/await` and removed some code which has nicer native API's.
(the starting point of this was old CLI code from one of the plugins)

Going to give it another look tomorrow to remove some duplication of split it into smaller functions.

packages/base-cli/src/index.ts
romainmenke: I split up the code from `index.ts` into 4 separate files with one function each.
This does not remove much of the duplication.
But each case is different in subtle ways.

Just splitting it up and using `async`/`await` makes each code path a bit more readable (imho).

packages/base-cli/src/index.ts
romainmenke: As answered in the other comment I chose to split first.
I think this is ready for another view :)

Antonio-Laguna (approved):
So much clearer now! Great work

PR #6, opened 2021-11-16
cmd generate-test-cases

romainmenke:
- a command to generate test cases
- generated test cases do not replace existing tests in plugins
- cases can be added and all plugins using these can re-generate the tests

Similar issues often appear in plugins and this can cause add extra friction as it is easy to forget a plugin when testing for a new known issue.

Having a tool to generate test cases for known issues for all relevant plugins can help.

Currently supports selectors and values, properties and declarations.

-------

I considered a fuzzer but found that we currently have more need to easily test known issues before we start testing for unknowns.

------

Notable fixes : 

https://github.com/csstools/postcss-logical/issues/37

MDN data is used to check if properties being transformed are animatable.
If they are a `@keyframes` test is added.

This is less automatic than it sounds. The property is manually declared when generating test cases.

_MDN data might also be useful in other ways.
Like generating tests for known idents.
Something like `:dir(diagonal)` becoming a thing (just an example) could be caught this way._

------

Note : 

I also update the `lint` command to be even more uniform and reusable.

`eslint src/**/*.ts src/**/*.js --no-error-on-unmatched-pattern`

->

`eslint ./src --ext .js --ext .ts --ext .mjs --no-error-on-unmatched-pattern`

This makes it easier to have these 3 types of source code :

- plugin code in typescript intended to be passed to rollup (`.ts`).
- plugin code in javascript intended to be passed to rollup (`.js`).
- utility code in javascript intended to be used locally (`.mjs`).


Antonio-Laguna (approved):
Amazing job! 😮‍💨

This is very thorough and boosts our confidence level moving forward while also helping to be more forgiving in some cases!

This is very impressive 👏

PR #7, opened 2021-11-16
postcss-base-plugin

romainmenke:
A base plugin.

This serves both as an example and a plugin to run tests against.
(We don't have to test the full CLI features for all plugins if we just test it here)

A new plugin can be started by duplicating this one and changing relevant bits.

This can also be scripting to have a command `npm run new-plugin postcss-foo` that does all the replacing for you.

-----------------

Notes : 

The contents of `package.json` are especially important here as they will form the basis for all other packages migrated to the mono repo.


plugins/postcss-base-plugin/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "peerDependencies": {
+    "postcss": "^8.3"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/csstools/postcss-plugins.git",
+    "directory": "plugins/postcss-base-plugin"
+  }
+}
```
romainmenke, line 44: In `postcss-nesting` there is an `exports` field.
Do we need or want this going forward?

https://github.com/csstools/postcss-nesting/blob/main/package.json#L18-L24

Antonio-Laguna: I don't think we do need that given that's it's pretty limited what we send to `npm` when packed and published. Do you want to keep it?

Antonio-Laguna (approved):

PR #17, opened 2021-11-20
Enable full CI for CLI

romainmenke:
integration after `base-cli` and `postcss-base-plugin` were merged.

### DIFF
--- a/packages/base-cli/src/io-fs-to-fs.ts
+++ b/packages/base-cli/src/io-fs-to-fs.ts
@@ -1,4 +1,4 @@
-import fsp from 'fs/promises';
+import { promises as fsp } from 'fs';
 import path from 'path';
 import { Arguments } from './args';
 import postcss, { Plugin } from 'postcss';
--- a/plugins/postcss-base-plugin/test/cli/test.sh
+++ b/plugins/postcss-base-plugin/test/cli/test.sh
@@ -9,6 +9,7 @@ postcss-base-plugin ./test/cli/basic.css --output ./test/cli/basic.result.css
 # Check result
 if [[ "$OSTYPE" == "win32" ]]; then
 	# CRLF on Windows makes it hard to test with base64 encoded sourcemaps
+	echo 'did not compare actual output assuming exit 0 means everything is ok'
 else
 	git --no-pager diff --no-index --word-diff ./test/cli/basic.expect.css ./test/cli/basic.result.css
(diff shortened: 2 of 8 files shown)


romainmenke:
@Antonio-Laguna any opinions on squash commits?
Would give a cleaner history on `main` after PR's that have lot's of smaller (sometimes bogus) commits.

Antonio-Laguna:
@romainmenke I completely agree. I've been avoiding that for now unless it was enforced since I wasn't sure if it affected the monorepo migration. I can enable squash as the only option here

packages/base-cli/src/io-fs-to-fs.ts
```diff
@@ -1,4 +1,4 @@
-import fsp from 'fs/promises';
+import { promises as fsp } from 'fs';
```
Antonio-Laguna, line 1: I often just do this

```suggestion
import fs from 'fs/promises';
```

But feel free to leave it like this if you want to be specific

plugins/postcss-base-plugin/test/cli/test.sh
```diff
@@ -1,13 +1,20 @@
...
 # Zero out result file
 echo '' > ./test/cli/basic.result.css;
 
 # Test with long flag
 postcss-base-plugin ./test/cli/basic.css --output ./test/cli/basic.result.css
 
 # Check result
-git --no-pager diff --no-index --word-diff ./test/cli/basic.expect.css ./test/cli/basic.result.css
+if [[ "$OSTYPE" == "msys" ]]; then
+	# CRLF on Windows makes it hard to test with base64 encoded sourcemaps
```
Antonio-Laguna, line 13: Nice one

Antonio-Laguna (approved):
This LGTM. Feel free to merge whenever you want!

packages/base-cli/src/io-fs-to-fs.ts
romainmenke: I mostly use the specific naming. This is more a reminder for myself that I imported the promises API further down in the code. :)

romainmenke:
@Antonio-Laguna 

> I completely agree. I've been avoiding that for now unless it was enforced since I wasn't sure if it affected the monorepo migration. I can enable squash as the only option here

Enforcing it would be fine I think.
I don't expect any issues with this during the migration :)

Issue #9, opened 2021-11-17
Enable CI


romainmenke:
Undo these changes in test.yml :
https://github.com/csstools/postcss-plugins/pull/2/commits/1607a255981ac9f2648cff2838f7f50637f6ffd4

PR #27, opened 2021-11-21
enable full CI

romainmenke:
Checks :
- eslint
- postcss-tape
- CLI cross platform
- deno
- multiple node versions

Caveats :

Node 12|14 don't have workspaces support.
For small changes this wont matter as dependencies are pulled from npm.

CI will fail when there is a breaking change in a plugin that affects `postcss-preset-env` 
To handle this the node 12|14 checks are run in a separate workflow.

This PR itself is currently failing because of this.
-> some plugins have breaking changes in how they need to be imported.

This will require a human to inspect the output and see if an error is expected or not.
Once everything has been released CI will be green again.

fixes : https://github.com/csstools/postcss-plugins/issues/9


Antonio-Laguna:
@romainmenke this is erroring as you can see above and also there seems to be a canceled run, let me know if you want to discuss this.

romainmenke:
@Antonio-Laguna yeah, it's failing because some plugins in this repo have breaking changes.
`postcss-preset-env` has been updated to match.

In older node versions those plugins are pulled from npm instead of the workspace.
So in node 12 and 14 we have a postcss-preset-env that was updated and expects new releases of for example css-blank-pseudo. But it downloads the current version of css-blank-pseudo

This is not something that we can (or should) solve.
The issue will go away over time when all node LTS versions have workspaces support.

CI will become green once we release packages and update versions in `postcss-preset-env`.

The skipped test is because it's a workflow with a matrix strategy.
Once there is a single failure the others are cancelled.

Antonio-Laguna:
Sounds sensible!

Antonio-Laguna (approved):

PR #31, opened 2021-11-22
Migrate css-blank-pseudo

romainmenke:
- remove `browser.js` and keep `browser-legacy.js` ([both were also broken](https://unpkg.com/css-blank-pseudo@2.0.0/browser-legacy.js))
- use `postcss-selector-parser`, used to be a regexp and was broken in some cases (e.g. `[foo=":blank"]`)
- rest follows work done in `css-has-pseudo`
- add warning about `:not(:blank)`

--------

### remove `browser.js` and keep `browser-legacy.js`

Keeping both adds a lot of complexity with little benefits (±200 bytes less).
Having just one that works in as many places as possible is best going forward imho.


Antonio-Laguna (approved):
Impressive!

PR #34, opened 2021-11-27
Migrate css-prefers-color-scheme

romainmenke:
- migrate `css-prefers-color-scheme`
- found and fixed a bug in how `browser-global.js` works


plugins/css-prefers-color-scheme/src/browser-global.js
```diff
@@ -0,0 +1,6 @@
+/* global self */
```
Antonio-Laguna, line 1: TIL 😂

Antonio-Laguna (approved):
Nice one. Learned about self on window context too 👍

PR #37, opened 2021-11-27
Migrate postcss-image-set-function

romainmenke:
- the original still used `Once`.
- converting it to `Declaration` resulted in an infinite loop.
- migrated to typescript to aid in debugging.

The end result should work exactly the same, even if the generated CSS is a bit different.
The previous version generated extra (unneeded) `Rule`'s for individual `Declaration`'s


Antonio-Laguna (approved):
LGTM!

PR #38, opened 2021-11-27
fork postcss-values-parser

romainmenke:
This is a fork of `postcss-values-parser`.
The only difference is that I completely removed the `registerWalker` stuff.

This does not pose a problem in actual plugins and fixes the issues reported completely from our perspective.

The package must be published on npm if this is a solution we like.

-----

I also added a minimal test setup so that we can easily debug and fix issues.
The original package uses [ava](https://github.com/avajs/ava) but I don't want to add that as a dependency here.


romainmenke:
Not going in this direction.
Superseded by : https://github.com/csstools/postcss-plugins/pull/53

PR #49, opened 2021-11-29
Feedback

### LINKED ISSUE #48: Issue templates feedback
Maybe we need 2 flows for bugs?

1. general code bugs (something throws an exception)
2. for CSS where the output is unexpected.

1) Would ask for things like error logs, and place emphasis on reproduction

2) Would have extra fields :
- Actual Before
- Actual After
- Expected Before
- Expected After

We could also nudge reporters to open a PR with a small CSS snippet in the test suite of the affected plugin (e.g. a failing test).

--------

I am wondering if we need to place `Actual` before `Expected`?
This is a more linear flow where reporters start with the bug description and behaviour.

---------

`Can you reproduce it with npx <plugin-name> minimal-example.css?`

Maybe we need a text field so that reporters can provide a reproduction in this style?

----------

>Reproduction link
Please provide a link to a repo that can reproduce the problem you ran into or a minimum reproducible case for your issue. A reproduction is required unless you are absolutely sure that the issue is obvious and the provided information is enough to understand the problem.

This states that it's optional, but I can't submit an issue without providing something.
Maybe best not to make it required?
Else we risk pushing people into entering "example.com" style bogus links.

The comment could maybe also be a bit more welcoming to persons who are just starting out :)
We don't want to take on the burden of teaching them everything, but we could explain better why a reproduction is so helpful.

Maybe something like this?

Reproduction link
Please provide a link to a repo that can reproduce the problem you ran into or a minimum reproducible case for your issue. These plugins are used in ways we can not always predict and showing us what you did helps us in tracking down any bugs.

Antonio-Laguna:
Thanks for your feedback @romainmenke !

I've reviewed the language and added a new path. CSS path is a tad more simple in general. Would love to know your thoughts!

Fixes #48

### DIFF
--- a/.github/ISSUE_TEMPLATE/css-issue.yml
+++ b/.github/ISSUE_TEMPLATE/css-issue.yml
@@ -1,15 +1,9 @@
-name: '💥 Crash report'
-description: Create a report for one of the plugins or PostCSS Preset Env with guided questions
+name: '🐞 Unexpected CSS output report'
+description: CSS output isn't what you expected? This is your path!
 body:
   - type: markdown
     attributes:
       value:  Thanks for taking the time to fill out this bug report!
-  - type: input
-    id: reproduction
-    attributes:
-      label: Reproduction link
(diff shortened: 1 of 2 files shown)


romainmenke (approved):
LGTM!

PR #53, opened 2021-11-30
postcss-value-parser

romainmenke:


plugins/postcss-double-position-gradients/src/index.js
```diff
@@ -38,37 +56,38 @@ module.exports = function creator(opts) {
...
 
 				const nodes = func.nodes;
 
 				nodes.slice(0).forEach((node, index, nodes) => {
-					const node1back = Object(nodes[index - 1]);
-					const node2back = Object(nodes[index - 2]);
-					const node1next = Object(nodes[index + 1]);
-
-					const isDoublePositionLength = node2back.type && node1back.type === 'numeric' && node.type === 'numeric';
+					const oneValueBack = Object(nodes[index - 2]); // Skip one for space
```
romainmenke, line 67: I added a filter before starting the logic to strip out comments and spaces. This makes it easier to handle unexpected things. Maybe that could work here too?

Antonio-Laguna: Nice call, addressed

Antonio-Laguna (approved):

PR #57, opened 2021-12-02
Fix open issues

romainmenke:
- fixes : https://github.com/csstools/css-prefers-color-scheme/issues/5
- fixes : https://github.com/csstools/css-prefers-color-scheme/pull/7
- fixes : https://github.com/csstools/css-has-pseudo/pull/13
- fixes : https://github.com/csstools/css-has-pseudo/pull/15

`css-prefers-color-scheme` needed refactors to achieve good cross browser support and allow complex media queries.

The generated CSS remains compatible with older versioned polyfill scripts.
The new polyfill script remains compatible with old CSS bundles.

An option was added to make it possible for users to prefer only the new or old style to reduce output size.


Antonio-Laguna (approved):

Antonio-Laguna:
Great work! Nicely done keeping things consistent.

Issue #60, opened 2021-12-04
`postcss-dir-pseudo-class` has poor support for complex selectors


romainmenke:
### Bug description

Complex selectors in `postcss-dir-pseudo-class` have unexpected output.

related : https://github.com/csstools/postcss-dir-pseudo-class/issues/3

### Source CSS

```css
button > :dir(ltr) {
	order: 36;
}
```

### Expected CSS

```css
button > [dir="ltr"],
[dir="ltr"] button > :not([dir="rtl"]) {
	order: 36;
}
```

or 

```css
[dir="ltr"] button > * {
	order: 36;
}
```

Not a 100% sure that this expected CSS is fully correct.

### Actual CSS

```css
[dir="ltr"] button >  {
	order: 36;
}
```

### Does it happen with `npx <plugin-name> minimal-example.css`?

Yes

### Extra config

N/A

### What plugin are you experiencing this issue on?

PostCSS Dir Pseudo Class

### Plugin version

6.0.0

### What OS are you experiencing this on?

macOS

### Node Version

16

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [x] I'm willing to open a PR

PR #62, opened 2021-12-05
postcss-dir-pseudo-class : fix combinators and support for `:is(:dir(ltr))`

romainmenke:
- fixes : https://github.com/csstools/postcss-plugins/issues/60
- added support for `:dir` inside other pseudo class functions (e.g. `:is(:dir(ltr))`)
- ignore unknown values inside `:dir()` (`:dir(unknown)` is now ignored)
- added a warning when hierarchical `:dir` is detected.

### warning when hierarchical `:dir` is detected.

This will never be possible to correctly transform.
If there was a way, there would have been no need for `:dir()` in the first place.
Emitting a warning informs users that the output will not be what they expect.

Maybe we need to start building a knowledge base of cases like this so that we can link to it in the warning? Fully explaining the issue will be useful in the long term.

This is very specific and requires a long text so the general README is not a good place.
A specific directory in the mono repo with articles in markdown might be best?

### DIFF
--- a/plugins/postcss-dir-pseudo-class/.tape.js
+++ b/plugins/postcss-dir-pseudo-class/.tape.js
@@ -1,10 +1,12 @@
 module.exports = {
 	basic: {
 		message: "supports basic usage",
+		warnings: 2,
 	},
 	"basic:dir": {
 		message: 'supports { dir: "ltr" } usage',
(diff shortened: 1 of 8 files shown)


Antonio-Laguna (approved):
Nice work, glad you found those !

Issue #67, opened 2021-12-13
Add nvm/volta config in each directory.


romainmenke:
Currently only configured for root.
Should be set in each plugin/package/...

PR #68, opened 2021-12-13
add nvm and volta config to all packages

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/67


packages/base-cli/.gitignore
```diff
@@ -1,12 +1,13 @@
 node_modules
+dist
```
romainmenke, line 2: this still an old `.gitignore` and `.nvmrc` was excluded by it..

Antonio-Laguna:
Nice work!

PR #77, opened 2021-12-16
fix css escaping  in postcss-has-pseudo

romainmenke:
Used to mix `encodeURIComponent` with regex to decode partially after encoding and then encoding again differently with regex.

Now goes through the value once and encodes each piece correctly from the start.


Antonio-Laguna:
@romainmenke this looks good to me and I guess it'd be hard but is there any chance we can get a test?

romainmenke:
@Antonio-Laguna The new code now correctly handles escape sequences already present in selectors :

input :

```css
:has(.esc\:aped) {
	order: 7;
}
```

previous output :

This was handled as if `:` indicated a pseudo.

```css
[\:has\(.esc\%5C\:aped\)] {
	order: 7;
}
```

new output :

Here `\:` is handled as an escaped `:` and part of the `.esca:ped` class name.

```css
[\:has\(.esc\%5C\%3Aaped\)] {
	order: 7;
}
```

Antonio-Laguna:
Thanks! That's perfect!

Antonio-Laguna (approved):
LGTM

PR #92, opened 2021-12-26
fix browser script exports

romainmenke:
This was very broken :/

`"default": "./dist/browser.js"`

`browser.js` does not exist, only `browser.cjs`

This change should ensure that both `require` and `import` will work correctly.


Antonio-Laguna (approved):
Good catch!

PR #101, opened 2021-12-30
fix : css-has-pseudo

romainmenke:
This plugin was is a bit broken.
I don't know it the spec changed or if it was always a bit broken, or if I broke during the migration.

It does show that we really need to start testing these features in real browsers.
We are flying blind here :/

### DIFF
--- a/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
@@ -250,7 +250,7 @@
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
--- a/plugin-packs/postcss-preset-env/test/basic.autoprefixer.false.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.autoprefixer.false.expect.css
@@ -250,7 +250,7 @@
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
--- a/plugin-packs/postcss-preset-env/test/basic.ch38.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.ch38.expect.css
@@ -170,7 +170,7 @@
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
--- a/plugin-packs/postcss-preset-env/test/basic.ch88-ff78-saf10.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.ch88-ff78-saf10.expect.css
@@ -183,7 +183,7 @@ h1.test-custom-selectors,h2.test-custom-selectors,h3.test-custom-selectors,h4.te
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
--- a/plugin-packs/postcss-preset-env/test/basic.ch88-ff78.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.ch88-ff78.expect.css
@@ -165,7 +165,7 @@ h1.test-custom-selectors,h2.test-custom-selectors,h3.test-custom-selectors,h4.te
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
--- a/plugin-packs/postcss-preset-env/test/basic.ch88-ff78.no-is-pseudo.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.ch88-ff78.no-is-pseudo.expect.css
@@ -165,7 +165,7 @@ h1.test-custom-selectors,h2.test-custom-selectors,h3.test-custom-selectors,h4.te
 	background-color: yellow;
 }
 
-.test-has-pseudo-class[\:has\(.inner-class\)] {
+.test-has-pseudo-class[\:has\(\.inner-class\)] {
 	background-color: yellow;
 }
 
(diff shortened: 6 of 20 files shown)


romainmenke:
>It does show that we really need to start testing these features in real browsers.

Maybe we can have a simple test setup running against a semi random chrome version via github workflows?

This should tell give us some feedback when updating things.

A more comprehensive test can be run from https://github.com/mrhenry/web-tests

Browserstack (or something similar) is needed to test a wide range of devices but these services are very flaky and I don't want to slow down development by having a flaky CI.
A full CI run against all relevant browsers also easily takes 1h30m.

`web-tests` just does scheduled runs and spits out reports.
It requires manually checking the results (because I haven't made badges or an API yet).

experimental/css-has-pseudo/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changes to CSS Has Pseudo
+
+### Unreleased
+
+Tracking initial implementation of `:has()` pseudo-class in Safari Technology Preview.
+This is a breaking change and affects both the generated CSS and the client side polyfill.
```
Antonio-Laguna, line 6: Could we couple minor examples (but meaningful, not something like .foo but just an input with hover and that sort of thing to quickly identify how's changing? Just an idea, not really a request

experimental/css-has-pseudo/INSTALL-POSTCSS.md
```diff
@@ -0,0 +1,163 @@
...
+
+```js
+const postcssHasPseudo = require('css-has-pseudo');
+
+module.exports = {
+  plugins: [
+    postcssHasPseudo(/* pluginOptions */)
+  ]
+}
+```
```
Antonio-Laguna, line 45: Can we add here along these lines, how to integrate with postcss-preset-env? It's trivial to us and we know what's going on but I'm not sure it's clear enough outside of this.

Antonio-Laguna: Perhaps better on the README

experimental/css-has-pseudo/browser.js
```diff
@@ -0,0 +1,2 @@
...rHTML="<x "+e.attributeName+">",r.setAttributeNode(o.children[0].attributes[0].cloneNode()),t.documentElement.style.zoom=1,t.documentElement.style.zoom=null})),e.nodes.forEach((function(r){-1===n.indexOf(r)&&(r.removeAttribute(e.attributeName),t.documentElement.style.zoom=1,t.documentElement.style.zoom=null)})),e.nodes=n}))}))}function a(t){try{[].forEach.call(t.cssRules||[],(function(t){if(t.selectorText)try{var o=function(t){for(var r,n=[],o=0,c=!1,i=!1,a=0;a<t.length;a++){var s=t[a];switch(s){case"[":if(c)continue;0===o?r="":r+=s,o++;continue;case"]":if(c)continue;if(0==--o){var u=e(r);u.indexOf(":has(")>-1&&n.push(u)}else r+=s;continue;case"\\":if(c){a++;continue}r+=t[a],r+=t[a+1],a++;continue;case'"':case"'":if(c&&s===i){c=!1;continue}c=!0,i=s;continue;default:r+=s;continue}}for(var l=[],p=0;p<n.length;p++)-1===l.indexOf(n[p])&&l.push(n[p]);return l}(t.selectorText.toString());if(0===o.length)return;for(var c=0;c<o.length;c++){var i=o[c];n.push({rule:t,selector:i,attributeName:encodeURIComponent(i).replace(/%3A/g,":").replace(/%5B/g,"[").replace(/%5D/g,"]").replace(/%2C/g,","),nodes:[]})}}catch(e){r.debug&&console.error(e)}else a(t)}))}catch(e){r.debug&&console.error(e)}}}}();
```
Antonio-Laguna, line 1: Why is this being VCed?

experimental/css-has-pseudo/src/browser.js
```diff
@@ -0,0 +1,178 @@
...
+
+	// document.createAttribute() doesn't support `:` in the name. innerHTML does
+	const attributeElement = document.createElement('x');
+
+	// walk all stylesheets to collect observed css rules
+	[].forEach.call(document.styleSheets, walkStyleSheet);
+	transformObservedItemsThrottled();
+
+	// observe DOM modifications that affect selectors
+	if ('MutationObserver' in self) {
```
Antonio-Laguna, line 42: Given this is required by the requisites, I think we should test this earlier on, and if not just throw and not run. 

I don't see (but could be wrong) if we're checking support for `:has` on JS before allowing any of this code to run.

Antonio-Laguna (approved):
@romainmenke this is astonishing! Great achievement here and I think this is very thorough. I'm really glad that we got some tests for browser too.

Left some comments there

experimental/css-has-pseudo/INSTALL-POSTCSS.md
romainmenke: I'll add this both in the readme in the `/experimental/` folder and in the readme of this plugin!

experimental/css-has-pseudo/browser.js
romainmenke: will fix!

experimental/css-has-pseudo/src/browser.js
romainmenke: Mutation observer is actually not a requirement for this polyfill.
I was surprised by this too.

It is needed to re-calc styles if html changes through client side javascript.

It's better not to throw or do early returns as this polyfill should embrace CSS concepts -> it should degrade very gracefully.

Antonio-Laguna: Good to know and surprising indeed.

Thing is, as it is right now I think (if Safari was fully compliant) it would run in Safari and it wouldn't be needed?

romainmenke: The polyfill would still be needed if people did not set `preserve : true`.
Since we don't know if they set `preserve : true` we have to always run the polyfill :/

experimental/css-has-pseudo/browser.js
romainmenke: done

experimental/css-has-pseudo/INSTALL-POSTCSS.md
romainmenke: I added these

experimental/css-has-pseudo/src/browser.js
romainmenke: I added a new option `forcePolyfill`.
If this is not set we do feature detection and early return.

We can slim down the amount of options and config later in the experimental track.
Having more is ok now I think :)

experimental/css-has-pseudo/CHANGELOG.md
romainmenke: I added a few examples here but do not want to go into too much detail.
Given the complexity of `:has` it's difficult to explain something without explaining everything (or at least for me)

Antonio-Laguna (approved):

PR #103, opened 2022-01-01
add plugin id help for config's

romainmenke:
partial fix for : https://github.com/csstools/postcss-preset-env/issues/156

Also see : https://github.com/csstools/postcss-plugins/discussions/86#discussioncomment-1887319

Uses a levenshtein distance to possibly suggest the correct plugin.

```
options: {
	features: {
		"custom-media": true,
		"postcss-logical": true,
		"postcss-logica": true,
	}
},
```

```
Unknown feature: "custom-media" did you mean: "custom-media-queries"
Unknown feature: "postcss-logical" did you mean: "logical-properties-and-values"
Unknown feature: "postcss-logica" did you mean: "logical-properties-and-values"
```

----------

Tests failed because we still had some that were using `color-mod` and this then throw warnings with the new help stuff.

Maybe the entire `insertBefore` `insertAfter` stuff is outdated since this is no longer of PostCSS plugins run? But this is outside the scope of this change.

### DIFF
--- a/plugin-packs/postcss-preset-env/src/index.js
+++ b/plugin-packs/postcss-preset-env/src/index.js
@@ -1,16 +1,18 @@
 import autoprefixer from 'autoprefixer';
 import browserslist from 'browserslist';
 import cssdb from 'cssdb';
-import plugins from './lib/plugins-by-id';
+import { pluginsById as plugins } from './lib/plugins-by-id';
 import getTransformedInsertions from './lib/get-transformed-insertions';
 import getUnsupportedBrowsersByFeature from './lib/get-unsupported-browsers-by-feature';
 import idsByExecutionOrder from './lib/ids-by-execution-order';
 import writeToExports from './lib/write-to-exports';
 import getOptionsForBrowsersByFeature from './lib/get-options-for-browsers-by-feature';
+import { pluginIdHelp } from './lib/plugin-id-help';
 
 const plugin = opts => {
 	// initialize options
 	const features = Object(Object(opts).features);
+	const featureNamesInOptions = Object.keys(features);
 	const insertBefore = Object(Object(opts).insertBefore);
 	const insertAfter = Object(Object(opts).insertAfter);
 	const browsers = Object(opts).browsers;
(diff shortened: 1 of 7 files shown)


plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -113,14 +115,24 @@ const plugin = opts => {
 	const usedPlugins = supportedFeatures.map(feature => feature.plugin);
 	usedPlugins.push(stagedAutoprefixer);
 
+	const internalPlugin = () => {
```
romainmenke, line 118: This was broken.
It's not possible to have both `OnceExit` and `plugins` in the same PostCSS plugin.

These need to be chained.

I do not know the `exportsTo` plugin so I can't easily add a test for it to double check that everything is fine now.

plugin-packs/postcss-preset-env/test/insert.after.expect.css
```diff
@@ -1,5 +1,3 @@
-$font: system-ui;
-
 .test-variable {
-	font-family: $font;
+	font-family: system-ui, -apple-system, Segoe UI, Roboto, Ubuntu, Cantarell, Noto Sans, sans-serif;
 }
```
Antonio-Laguna, line 3: This change confuses me a bit. Why is this needed?

Antonio-Laguna (commented):
Thanks for this! This is great and I think it'll help in the future to ease the plugin pack usage. Just left a minor comment. I'm sure I'm missing something.

plugin-packs/postcss-preset-env/src/lib/plugin-id-help.js
```diff
@@ -11,6 +11,10 @@ export function pluginIdHelp(featureNamesInOptions, root, result) {
 		const byId = mostSimilar(featureName, featureNames);
 		const byPackage = mostSimilar(featureName, packageNames);
 
+		// TODO :
+		// 1. create markdown docs with the plugin id's
```
romainmenke, line 15: outside of scope of this change.
But I want to already indicate were in the code this can be mentioned to users.

plugin-packs/postcss-preset-env/test/insert.after.expect.css
romainmenke: It confused me too.

tape contents for this test :

```js
	'insert:after:array': {
		message: 'supports { stage: 1, after: { "lab-function": [ require("postcss-simple-vars") ] } } usage',
		options: {
			stage: 1,
			insertAfter: {
				'lab-function': [
					require('postcss-simple-vars')()
				]
			},
			features: {
				'lab-function': {
					unresolved: 'ignore'
				}
			}
		},
		expect: 'insert.after.expect.css'
	},
```

I replaced the `color-mod` stuff with `lab-function` to resolve the warnings.
I think that `insertAfter` looks up the index of the plugin (`lab-function`) and tries to add the other plugin after it.

If the feature is not found I think it was never inserted.

So when `color-mod` was removed `postcss-simple-vars` was no longer applied correctly in this test.

Now that I have changed it to `lab-function` it actually works again.
Although probably not as designed because of plugin (not) ordering in PostCSS 8

Antonio-Laguna: Hmm we should probably add another issue to test this meaningfully since I don't think this test proves anything right now

PR #119, opened 2022-01-02
preset-env : fix after/before and export tests

romainmenke:
## insertBefore / insertAfter

see : https://github.com/csstools/postcss-plugins/pull/103#discussion_r777131651

| insert | matches | should change | why |
| ---- | ---- | ---- | --- |
| `before` | `source` | yes | it runs before other plugins |
| `before` | `result` | yes | PostCSS 8 revisits changed nodes |
| `after` | `source` | no | source has been changed by other plugins |
| `after` | `result` | yes | it runs after other plugins |

I added tests that check all these options and this seems to work.
So there is nuance in PostCSS 8 plugin execution order.

Plugin execution has an order. But because nodes are revisited, plugins that come before can still do their work after other plugins.

I assume this will be different if one or more of the plugins have `async` functions.

-------

## exportTo

see : https://github.com/csstools/postcss-plugins/pull/103#discussion_r777125676

These tests had false positives before.

The test result files were tracked by git so always existed and the test just checked that they matched the contents before and after the test.

Because the `exportTo` functionality never ran after updating to PostCSS 8 the files were never changed. Tests passed simply because the files existed.

I updated the test to remove the files during the run.
If we have a regression and `exportTo` breaks again, tests will fail.

### DIFF
--- a/plugin-packs/postcss-preset-env/.tape.js
+++ b/plugin-packs/postcss-preset-env/.tape.js
@@ -200,10 +200,13 @@ module.exports = {
 			}
 		}
 	},
-	'insert:after:match-result': {
-		message: 'supports { stage: 2, features: { "lab-function": { unresolved: "ignore" } }, insertAfter: { "lab-function": require("postcss-simple-vars")() } } usage',
+	'insert:after:match-result:exec': {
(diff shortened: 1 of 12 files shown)


Antonio-Laguna (approved):
This looks awesome! Thanks for doing this so quickly!

PR #128, opened 2022-01-04
css-has-pseudo : specificity and hover

romainmenke:
- better `[a, b, c]` specificity matching
- use `mouseout` and `pointerleave` for better hover tracking


Antonio-Laguna (approved):
Great trick with `:not`!

PR #129, opened 2022-01-06
improve specificity calculation

romainmenke:
https://drafts.csswg.org/selectors/#specificity-rules

>The specificity of an :is(), :not(), or :has() pseudo-class is replaced by the specificity of the most specific complex selector in its selector list argument.

I misread this as being the sum, it is not.
It has to be the most specific.

_In practice it being wrong will rarely lead to a different result.
So not prio for release_

### DIFF
--- a/experimental/css-has-pseudo/src/index.js
+++ b/experimental/css-has-pseudo/src/index.js
@@ -131,10 +131,39 @@ function selectorSpecificity(node) {
 			case ':has':
 			case ':not':
 				{
-					const pseudoSpecificity = selectorSpecificity(node.nodes[0]);
-					a += pseudoSpecificity.a;
-					b += pseudoSpecificity.b;
-					c += pseudoSpecificity.c;
+					if (node.nodes && node.nodes.length > 0) {
+						let mostSpecificListItem = {
+							a: 0,
+							b: 0,
+							c: 0,
+						};
+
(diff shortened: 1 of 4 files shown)


experimental/css-has-pseudo/src/index.js
```diff
@@ -131,10 +131,39 @@ function selectorSpecificity(node) {
...
-					c += pseudoSpecificity.c;
+					if (node.nodes && node.nodes.length > 0) {
+						let mostSpecificListItem = {
+							a: 0,
+							b: 0,
+							c: 0,
+						};
+
+						node.nodes.forEach((child) => {
+							const itemSpecificity = selectorSpecificity(child);
```
Antonio-Laguna, line 142: This is why I was mentioning extracting since I've been seeing this

Antonio-Laguna (approved):
I'm OK merging and releasing this, let me know if you want this.

Great work!

experimental/css-has-pseudo/src/index.js
romainmenke: Yes, each time I copy/paste a bit I remind myself that I should fine a way to have these imported from a central util bundle :D

romainmenke:
>if you want this.

I want it :) 
thank you!

PR #130, opened 2022-01-06
`:is()` pseudo class function

romainmenke:
_Todo : summarize the change after initial reviews._

_I would like to review this step by step, starting with the external API before taking a deep dive into how it works._

-----

At the moment any feedback on plugin naming, plugin options and docs is very much appreciated!

- [ ] name is final
- [ ] option names and docs are final
- [ ] general docs are final

[readme link](https://github.com/csstools/postcss-plugins/blob/pseudo-is--reliable-doberman-pinscher-a5cede45d7/plugins/postcss-is-pseudo/README.md)

[tape test config](https://github.com/csstools/postcss-plugins/blob/pseudo-is--reliable-doberman-pinscher-a5cede45d7/plugins/postcss-is-pseudo/.tape.js)

_tape test config has all the options which might be useful_

------


Antonio-Laguna:
@romainmenke is this ready to be merged?

romainmenke:
Too soon to merge, but it is ready for review.
I mentioned the bits I think are useful to discuss and get right first.

I want to let it rest for a bit an come back to in a few days with fresh eyes.

plugins/postcss-is-pseudo/INSTALL.md
```diff
@@ -0,0 +1,164 @@
...
+
+| [Node](#node) | [PostCSS CLI](#postcss-cli) | [Webpack](#webpack) | [Create React App](#create-react-app) | [Gulp](#gulp) | [Grunt](#grunt) |
+| --- | --- | --- | --- | --- | --- |
+
+## Node
+
+Add [PostCSS Is Pseudo] to your project:
+
+```bash
+npm install postcss @csstools/postcss-is-pseudo --save-dev
```
Antonio-Laguna, line 14: Any strong opinions on adding the org vs not adding the org?

plugins/postcss-is-pseudo/README.md
```diff
@@ -0,0 +1,234 @@
...
+	order: 1;
+}
+:is(input, button):is(:hover, :focus) {
+	order: 1;
+}
+```
+
+### doesNotExistName
+
+The `doesNotExistName` option allows you to change to selector used for `:not()` selectors.
```
Antonio-Laguna, line 98: Let's add a note here that hints that this is needed and will be explained below (and potentially link to the shortcomings)

plugins/postcss-is-pseudo/README.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+## ⚠️ Known shortcomings
+
+### Specificity
+
+`:is` takes the specificity of the most specific list item.
+We can increase specificity with `:not` selectors, but we can't decrease it.
+
+Converted selectors are ensure to have the same specificity as `:is` for the most important bit.
```
Antonio-Laguna, line 153:
```suggestion
Converted selectors are ensured to have the same specificity as `:is` for the most important bit.
```

plugins/postcss-is-pseudo/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+```css
+/* without is */
+.alpha:focus > .beta ~ .beta {
+	order: 2;
+}
+```
+
+If you have a specific pattern you can open an issue to discuss it.
+We can detect and transform some case but can't generalize into a single solution.
```
Antonio-Laguna, line 223:
```suggestion
We can detect and transform some cases but can't generalize them into a single solution that tackles all of them. 
```

plugins/postcss-is-pseudo/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@csstools/postcss-is-pseudo",
+  "version": "0.1.0",
```
Antonio-Laguna, line 3: Is it intended to be experimental?

plugins/postcss-is-pseudo/src/split-selectors/split-selectors.ts
```diff
@@ -0,0 +1,175 @@
...
+function cartesianProduct(...args) {
+	const r = [], max = args.length - 1;
+	function helper(arr, i) {
+		for (let j = 0, l = args[i].length; j < l; j++) {
+			const a = arr.slice(0); // clone arr
+			a.push(args[i][j]);
+			if (i == max)
+				r.push(a);
+			else
+				helper(a, i + 1);
```
Antonio-Laguna, line 170: [nit] Can we be consistent with the brackets?

plugins/postcss-is-pseudo/src/split-selectors/specificity.ts
```diff
@@ -0,0 +1,128 @@
+import parser from 'postcss-selector-parser';
```
Antonio-Laguna, line 1: Wondering if we should extract this to be available elsewhere but you have more idea here.

Antonio-Laguna (commented):
Left some initial feedback as you suggested.

Is this intended to be removed from nesting on the same sweep?

plugins/postcss-is-pseudo/src/split-selectors/split-selectors.ts
romainmenke: good catch! I'll check if we can make the linter fix this.
Code was copy/paste

plugins/postcss-is-pseudo/src/split-selectors/specificity.ts
romainmenke: Going to make an issue for this.
Need to figure out how we handle reusable bits of code

plugins/postcss-is-pseudo/INSTALL.md
romainmenke: I like the org for multiple reasons :

- [avoiding substitution attacks](https://github.blog/2021-02-12-avoiding-npm-substitution-attacks/) (we would not be the victim but without scoping package names we can contribute to the issue)
- name recognition.
- we don't have to pick a less logical package name if the best option is already taken without scope.

plugins/postcss-is-pseudo/src/split-selectors/split-selectors.ts
romainmenke: I have added eslint rules for tabs and braces.
This should no longer pass CI.

plugins/postcss-is-pseudo/src/split-selectors/specificity.ts
romainmenke: I created an issue for this.

plugins/postcss-is-pseudo/package.json
romainmenke: I mostly do a `0.1.0` release first for other projects?
Not sure if there is an existing convention here?

It should not be placed in `/experimental/` because there is no existing plugin.

plugins/postcss-is-pseudo/README.md
romainmenke: Linked to the specificity docs and renamed to option to `specificityMatchingName`
It is still not fully clear from the name alone but better than `doesNotExistName`

plugins/postcss-is-pseudo/package.json
romainmenke, line 3:
```suggestion
  "version": "1.0.0",
```

PR #133, opened 2022-01-07
Fix Interop Issue with Import and Custom Properties

### LINKED ISSUE #132: Synchronous run of `postcss-custom-properties` leads to unexpected results
### Bug description

Introduced here: https://github.com/postcss/postcss-custom-properties/pull/181

Most of the time `canReturnSyncFunction` is `true` and this leads to vars in other places not being properly transformed, this is especially dangerous when combined with `postcss-import`.

### Source CSS

```css
/* util/an-import.css */
:root {
    --foo: red;
}
```

```css
/* Another file */
@import './util/an-import.css';

.foo {
    color: var(--foo);
}
```

### Expected CSS

```css
:root {
    --foo: red;
}

.foo {
    color: red;
    color: var(--foo);
}
```

### Actual CSS

```css
:root {
    --foo: red;
}

.foo {
    color: var(--foo);
}
```

_No response_

### Extra config

This needs to happen on external files

PostCSS Custom Properties

### Plugin version

12.0.2

macOS

### Node Version

14.17.3

### Validations

Antonio-Laguna:
Fixes #132

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,7 +72,7 @@
     },
     "experimental/css-has-pseudo": {
       "name": "@csstools/css-has-pseudo-experimental",
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "CC0-1.0",
       "dependencies": {
         "postcss-selector-parser": "^6.0.8"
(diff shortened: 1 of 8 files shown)


romainmenke (approved):
LGTM!

Thank you for taking a deep dive into this 🎉

PR #135, opened 2022-01-07
PostCSS Custom Properties – add tests for variables with `url()`

marcalexiei:
Follo up to [PR 262](https://github.com/postcss/postcss-custom-properties/pull/262).

### DIFF
--- a/plugins/postcss-custom-properties/test/basic.css
+++ b/plugins/postcss-custom-properties/test/basic.css
@@ -16,6 +16,7 @@ html {
 	--shadow: 0 6px 14px 0 color(var(--shadow-color) a(.15));
 	--font-family: "Open Sans", sans-serif;
 	color: var(--color);
+	--url: url("/my/path");
 }
 
 :root,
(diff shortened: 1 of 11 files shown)


plugins/postcss-custom-properties/test/basic.css
```diff
@@ -16,6 +16,7 @@ html {
 	--shadow: 0 6px 14px 0 color(var(--shadow-color) a(.15));
 	--font-family: "Open Sans", sans-serif;
 	color: var(--color);
+	--url: url("/my/path");
```
romainmenke, line 19:
```suggestion
	--url: url("/my/path");
	color: var(--color);
```

Maybe we also need variations with `'` quotes and without any quotes?

Good list of examples : https://developer.mozilla.org/en-US/docs/Web/CSS/url()

romainmenke:
@marcalexiei Thank you for this.

I added some more `url()` cases after merging this PR.
With the previous parser we have also seen issues with data urls and want to get a signal if there is a regression with the current value parser.

Issue #100, opened 2021-12-29
css-has-pseudo is not working when hovering on child


abdallahz3:
Hi,
I have css like this
```css
.wrapper:has(> input:hover) {
  /* style links that contain an image */
  background-color: green;
}
```
When hovering on the input nothing happens, but clicking on the input, the wrapper turns green, the same behaviour happens 
if we replaced input:hover with button:hover.
if we replaced input:hover with p:hover for example, nothing happens

I'm using css-has-pseudo version 3.0.1

here is the html just in case
```html
<div className="wrapper">
  <input />
</div>
```

romainmenke:
Note : 

```bash
pbpaste | npx -y css-has-pseudo
```

Input

```css
.wrapper[\:has\(\%3E\%20input\:hover\)] {
  /* style links that contain an image */
  background-color: green;
}
```

Output

```css
.wrapper:has(> input:hover) {
  /* style links that contain an image */
  background-color: green;
}
```

romainmenke:
@abdallahz3 Did you include the javascript polyfill on your page?
And could you create and share a demo html page?

romainmenke:
Noticed two things :

```html
<div className="wrapper">
  <input />
</div>
```

This html is incorrect and should be :

```html
<div class="wrapper">
  <input />
</div>
```

-------

The Javascript polyfill doesn't support `:hover`.
This was maybe done for performance as this requires `mouseenter` `mouseleave` event listeners.

abdallahz3:
> @abdallahz3 Did you include the javascript polyfill on your page? And could you create and share a demo html page?

Yes

abdallahz3:
> 

yeah I used react, sorry for that mistake

here is a stackblitz link using raw html https://stackblitz.com/edit/vite-ekfuvl?file=index.html

abdallahz3:
> Noticed two things :
> 
> ```
> <div className="wrapper">
>   <input />
> </div>
> ```
> 
> This html is incorrect and should be :
> 
> ```
> <div class="wrapper">
>   <input />
> </div>
> ```
> 
> The Javascript polyfill doesn't support `:hover`. This was maybe done for performance as this requires `mouseenter` `mouseleave` event listeners.

yeah I used react, sorry for that mistake

here is a stackblitz link using raw html https://stackblitz.com/edit/vite-ekfuvl?file=index.html

romainmenke:
> Is hover support planed or css has selector does not support hover

[...]

PR #136, opened 2022-01-08
`:has()` pseudo : more docs, tests and add typings

romainmenke:
- add docs for all plugin options
- explain how CORS comes into play with the browser plugin
- change tests to trigger CORS (so that we know our docs are correct)
- add typescript for the plugin
- get stryker coverage back up to 100% for everything in `src/encode/`
- investigated and added tests for plugin ordering.

-----------

see : https://github.com/csstools/postcss-plugins/issues/100#issuecomment-1007800947

After this change is merged and released we wait until Safari ships `:has`.
If we receive no further feedback, this version might become the next major update for `css-has-pseudo`.


experimental/css-has-pseudo/README-BROWSER.md
```diff
@@ -40,8 +40,72 @@ cssHasPseudo(document);
 <script>cssHasPseudo(document)</script>
 ```
 
+⚠️ Please use a versioned url, like this : `https://unpkg.com/@csstools/css-has-pseudo-experimental@0.1.0/dist/browser-global.js`
```
Antonio-Laguna, line 43:
```suggestion
⚠️ Please use a versioned url, like this : `https://unpkg.com/@csstools/css-has-pseudo-experimental@0.1.2/dist/browser-global.js`
```

Let's use latest

Antonio-Laguna (approved):
This is impressive! Loved the TS support and making this ready to swap with major.

Left a minor comment but happy to leave as is, your call :)

experimental/css-has-pseudo/README-BROWSER.md
romainmenke: This should also be generated :)
Made a note for this.

experimental/css-has-pseudo/README-BROWSER.md
```diff
@@ -40,8 +40,72 @@ cssHasPseudo(document);
 <script>cssHasPseudo(document)</script>
 ```
 
+⚠️ Please use a versioned url, like this : `https://unpkg.com/@csstools/css-has-pseudo-experimental@0.1.2/dist/browser-global.js`
```
romainmenke, line 43:
```suggestion
⚠️ Please use a versioned url, like this : `https://unpkg.com/@csstools/css-has-pseudo-experimental@0.2.0/dist/browser-global.js`
```

Next release will be 0.2.0

romainmenke:
Thank you for reviewing all this!

PR #139, opened 2022-01-09
`postcss-custom-properties` : fix types and some small bugs

romainmenke:
- converted to typescript
- fixed type definitions
- add support for `.mjs` in `importFrom`
- fix unicode support
- add `overrideImportFromWithRoot` option

#### typings

This plugin was a good example of how `.d.ts` files for typescript are often incorrect.
They are created manually without any tool analyses to check correctness.

Better to have true typescript as we then also get all the help from the right tools.

#### `.mjs`

The plugin actually already support `.mjs` in `importFrom` but this was not allowed.

- allow it
- update package.json so that it runs as a module when the plugin is called from another `.mjs` file.

#### unicode

This was a tiny fix.
One less regexp and it works.

#### `overrideImportFromWithRoot`

This really is a matter of perspective and both orders are equally valid.
They both allow different tactics and features.

https://github.com/postcss/postcss-custom-properties/pull/222

This was updated a while back as a minor version bump but this was actually a breaking change.

I fear not having an option that controls this will just cause someone to reverse it again in the future.

-------

fixes : https://github.com/postcss/postcss-custom-properties/issues/255
fixes : https://github.com/postcss/postcss-custom-properties/issues/237
fixes : https://github.com/postcss/postcss-custom-properties/issues/231


plugin-packs/postcss-preset-env/.tape.js
```diff
@@ -235,6 +235,27 @@ module.exports = {
 			stage: 0
 		}
 	},
+	'import:ch88': {
```
romainmenke, line 238: This is very much broken.

romainmenke: see : https://github.com/csstools/postcss-plugins/issues/140

Antonio-Laguna (approved):
Nice one, will keep an eye on #140 and potentially continue discussions there.

Issue #147, opened 2022-01-10
"r.assign is not a function" error coming from postcss-logical


alebelcor:
### Reproduction link

https://stackblitz.com/edit/nextjs-fx3c4g

### Bug description

It seems like the postcss-logical plugin is blowing up for me (for some reason) when I use the `transition` property in Next.js with CSS Modules.

### Actual Behavior

I see a `TypeError: r.assign is not a function` error on Next.js' dev overlay that is seemingly coming from `postcss-loader`.

The stacktrace in development is not very useful, but if I try to build I see this:

```
./src/components/Button/Button.module.css.webpack[javascript/auto]!=!./node_modules/next/dist/build/webpack/loaders/css-loader/src/index.js??ruleSet[1].rules[2].oneOf[
4].use[1]!./node_modules/next/dist/build/webpack/loaders/postcss-loader/src/index.js??ruleSet[1].rules[2].oneOf[4].use[2]!./src/components/Button/Button.module.css
TypeError: r.assign is not a function
    at /Users/alebelcor/projects/test/src/components/Button/Button.module.css:12:3
    at W (/Users/alebelcor/projects/test/node_modules/postcss-logical/dist/index.cjs:1:7015)
    at /Users/alebelcor/projects/test/node_modules/postcss-logical/dist/index.cjs:1:11461
    at LazyResult.visitTick (/Users/alebelcor/projects/test/node_modules/next/node_modules/postcss/lib/lazy-result.js:456:16)
    at LazyResult.runAsync (/Users/alebelcor/projects/test/node_modules/next/node_modules/postcss/lib/lazy-result.js:372:30)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at runNextTicks (internal/process/task_queues.js:64:3)
    at processImmediate (internal/timers.js:437:9)
    at async Span.traceAsyncFn (/Users/alebelcor/projects/test/node_modules/next/dist/trace/trace.js:74:20)
    at async /Users/alebelcor/projects/test/node_modules/next/dist/build/webpack/loaders/postcss-loader/src/index.js:46:22
    at async Span.traceAsyncFn (/Users/alebelcor/projects/test/node_modules/next/dist/trace/trace.js:74:20)
```

Which, to me, looks like it's coming from `postcss-logical`. Apologies if that's not the case.

If I comment out/remove the CSS line with the `transition` property the error goes away and everything works.

[...]

PR #148, opened 2022-01-10
fix postcss 8.2.x compat

romainmenke:
We do not support PostCSS v8.2.5 any more but as some have this version pinned by tools we can do a minimal effort to help them out.


Antonio-Laguna (approved):
This LGTM.

IMHO we should add links to #147 

Furthermore, I'm inclined to point to vercel/next.js/issues/33135 for a couple of reasons:

- So it's obvious we're fixing a Next.js issue.
- Easier navigation

Let me know your thoughts here.

romainmenke:
I was hesitant to point to one individual tool repackager as there are so many these days.
I bet more are doing it wrong or less than ideal.

Even if I have strong personal opinions about this trend it is however the reality of frontend development today. Keeping it vague might be more productive in the long run :)

Antonio-Laguna:
It's not a finger pointing thing IMHO but rather relating these two tools and one issue, Tomorrow could be Tailwind 😂

Won't enforce it though so will leave it up to you. 

Thoughts on linking the original issue though?

romainmenke:
Hehe, tailwind actually does this correctly.

https://github.com/tailwindlabs/tailwindcss/blob/62569092e7ba79fc45f77451c179b3f6cc0f80f8/package.json#L64

I have added the link to the original issue in the changelog.
Indeed a good addition to track this change back in the future!

PR #149, opened 2022-01-11
document autoprefixer grid edge case

romainmenke:
see : https://github.com/postcss/autoprefixer/issues/1442#issuecomment-1008939516

I ended up deciding it is best we document this here.
The `preserve` option is something we mainly do and we set a fixed order of plugins.
Both of which contribute to the issue.

### DIFF
--- a/plugin-packs/postcss-preset-env/README.md
+++ b/plugin-packs/postcss-preset-env/README.md
@@ -240,10 +240,10 @@ Becomes :
 }
 ```
 
-The prefixed `-ms-grid-columns` still has a custom property: `1fr var(--grid-gap) 1fr;` which won't work.
+The prefixed `-ms-grid-columns` still has a custom property: `1fr var(--grid-gap) 1fr;` which won't work.<br />
 This example shows issues with custom properties but other transforms might have similar issues.
 
-If you target IE or older Edge it is best to avoid using other modern features in grid property values.
+If you target IE or older Edge it is best to avoid using other modern features in grid property values.<br />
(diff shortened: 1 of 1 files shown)


plugin-packs/postcss-preset-env/README.md
```diff
@@ -214,6 +214,37 @@ postcssPresetEnv({
...
+.test-grid {
+	grid-gap: 15px;
+	grid-gap: var(--grid-gap);
+	-ms-grid-columns: 1fr var(--grid-gap) 1fr;
+	grid-template-columns: repeat(2, 1fr);
+}
+```
+
+The prefixed `-ms-grid-columns` still has a custom property: `1fr var(--grid-gap) 1fr;` which won't work.
+If you target IE or older Edge you can set [`preserve: false`](#preserve) or avoid certain features in grid related properties.
```
Antonio-Laguna, line 244: Is this the only issue you've detected? 

If so let's make this paragraph and above more clear that the issue is with custom properties and grids and offer both solutions.

I think preserve is great so discouraging that is a bit extreme (especially for custom properties) so I'd rather discourage using custom properties for grid prefixing or even doing your own auto prefixing if that's your thing and potentially adding `autoprefixer ignore next `

romainmenke: great point! I will update this.

romainmenke: I have clarified that it is not limited to custom properties and updated the proposed solution.

Antonio-Laguna (approved):
Nice!

PR #150, opened 2022-01-12
preset-env : always run plugins with side effects, ignoring browser lists

romainmenke:
It is not the right fix, but it is the one we can roll out now with the least disruption.

In the future we should remove any feature that is a side effect and not a polyfill.
The plugins in `postcss-preset-env` should never expose non-standard features.


plugin-packs/postcss-preset-env/test/import.ch88.expect.css
```diff
@@ -1,8 +1,9 @@
 .test-custom-properties {
+	order: 1;
 	order: var(--order);
 }
 
-@media (max-width: env(--sm)) {
+@media (max-width: 40rem) {
 	.test-custom-media-queries {
 		order: 2;
 	}
```
romainmenke, line 9: These are the plugins in question.
They didn't run before for modern chrome, but they do now.

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -99,18 +101,27 @@ const plugin = opts => {
...
-				),
-			),
-	);
+	// - features supported by the stage
+	// - features with `true` or with options
+	// - required for the browsers
+	// - having "importFrom" or "exportTo" options
+	const supportedFeatures = stagedFeatures.filter((feature) => {
+		if (feature.id in features) {
+			return features[feature.id];
```
romainmenke, line 110: I am pretty sure that this code is non ideal, but this bit (109-111) is equivalent to how it was before.

In practice it means that you lose the dynamic nature of postcss-preset-env if you set any explicity plugin options.

We should start a list of changes for a next major version to cleanup API that is hard to use and hard to implement.
Overloading enabling/disabling a plugin on top of providing plugin options is just not the best API design. :)

Antonio-Laguna (approved):
Thanks for doing this so quickly!

PR #152, opened 2022-01-12
fix cssdb badges

romainmenke:


Antonio-Laguna (approved):
Thanks for catching this!

PR #160, opened 2022-01-13
`postcss-preset-env` 7.3.0

romainmenke:


plugin-packs/postcss-preset-env/test/basic.ff49.expect.css
```diff
@@ -149,6 +149,10 @@
 	overflow-wrap: break-word;
 }
 
+.test-focus-visible-pseudo-class.focus-visible {
```
romainmenke, line 152: We need to check why this changed, might need the same work around as `:blank`

romainmenke: Added tests and this did not change because of a stage update.
Most likely better browser support stats than before coming from cssdb.

Antonio-Laguna:
From a quick search on the word, `unreleased` these are the plugins to be released:

* `postcss-is-pseudo`
* `css-prefers-color-scheme`
* `postcss-custom-properties`
* `postcss-image-set-function`
* `postcss-pseudo-class-any-link`

However there are more tags here so need to investigate this list further

romainmenke:
packages with code changes in this branch : 

```
cli/csstools-cli/src/cli.ts
plugin-packs/postcss-preset-env/src/index.js
plugins/postcss-image-set-function/src/lib/get-media.ts
plugins/postcss-is-pseudo-class/src/index.ts
plugins/postcss-pseudo-class-any-link/src/index.js
```

The rest of the labels where added because the Discord link was added ;)

PR #161, opened 2022-01-14
custom properties : plugin reset on exit

romainmenke:
When using the same plugin object for multiple processes or with a watcher there is an edge case we don't currently handle :

1. re-use the same instance
2. add a custom property declaration on `:root`
3. process runs
4. (accidentally) remove the declaration from `:root`
5. process runs again
6. the removed variable still exists in the `Map` on the plugin instance and can be used to polyfill

This can lead to hard to debug cases where something worked locally with a watcher running and then breaks after a build in CI.


Antonio-Laguna (approved):
This is great, one of the things you wonder why they weren't like this before!

Issue #131, opened 2022-01-07
Remove Can I Use Lite


Antonio-Laguna:
```
const browserslist = require( 'browserslist' );

module.exports = ( opts ) => {
    const targets = new Set( browserslist( opts.browsers ) );
    const requirements = browserslist( [
        'chrome < 88',
        'firefox < 89',
        'safari < 15',
        'edge < 88',
        'opera < 74',
        'ie <= 11',
    ] );

    const shouldPolyfill = requirements.find( ( x ) => {
        return targets.has( x );
    } );
```

https://github.com/mrhenry/core-web/blob/main/packages/core-web-generator/src/browsers-to-engines/browsers-to-engines.ts

Probably only need to iterate over the versions and prepend with < if caniuse has y or take the version from mdn and simply prepend <

Issue #115, opened 2020-03-08
Output the targets/plugins used


dmnsgn:
The `debug` option from babel preset env (https://babeljs.io/docs/en/babel-preset-env#debug) is particularly useful when using browsers and trying to understand which plugins are being used. 

I assume it would list which of these are used https://github.com/csstools/postcss-preset-env/blob/master/src/lib/plugins-by-id.js

romainmenke:
@dmnsgn `postcss-preset-env` now supports setting `{ debug: true }` in the options.
This will output a lot of information about which plugins are enabled/disabled and why.
@Antonio-Laguna did all the work here!

dmnsgn:
Thank you, very useful 👍

PR #163, opened 2022-01-15
Adding debug option

Antonio-Laguna:
This intentionally leaves out the browser matching debug since this will be handled via #131 (which is next on my list)

#115

### DIFF
--- a/plugin-packs/postcss-preset-env/README.md
+++ b/plugin-packs/postcss-preset-env/README.md
@@ -354,6 +354,10 @@ postcssPresetEnv({
 });
 ```
 
+### debug
+
+The `debug` option enables debugging messages to stdout which should be useful to help you debug which features have been enabled/disabled and why.
+
 [cli-img]: https://github.com/csstools/postcss-plugins/workflows/test/badge.svg
 [cli-url]: https://github.com/csstools/postcss-plugins/actions/workflows/test.yml?query=workflow/test
 [git-img]: https://img.shields.io/badge/support-chat-blue.svg
--- a/plugin-packs/postcss-preset-env/src/index.js
+++ b/plugin-packs/postcss-preset-env/src/index.js
@@ -10,6 +10,8 @@ import getOptionsForBrowsersByFeature from './lib/get-options-for-browsers-by-fe
 import { pluginIdHelp } from './lib/plugin-id-help';
 import { pluginHasSideEffects } from './lib/plugins-with-side-effects';
 
+const DEFAULT_STAGE = 2;
+const OUT_OF_RANGE_STAGE = 5;
 const plugin = opts => {
 	// initialize options
 	const options = Object(opts);
@@ -18,11 +20,15 @@ const plugin = opts => {
 	const insertBefore = Object(options.insertBefore);
 	const insertAfter = Object(options.insertAfter);
 	const browsers = options.browsers;
-	const stage = 'stage' in options
-		? opts.stage === false
-			? 5
-			: parseInt(opts.stage) || 0
-		: 2;
+	let stage = DEFAULT_STAGE;
+
+	if (typeof options.stage !== 'undefined') {
+		if (options.stage === false) {
+			stage = OUT_OF_RANGE_STAGE;
+		} else {
+			stage = parseInt(options.stage, 10) || 0;
+		}
+	}
 	const autoprefixerOptions = options.autoprefixer;
 	const sharedOpts = initializeSharedOpts(options);
 	const stagedAutoprefixer = autoprefixerOptions === false
@@ -51,7 +57,7 @@ const plugin = opts => {
 				browsers: unsupportedBrowsers,
 				plugin:   feature.plugin,
 				id:       `${feature.insertBefore ? 'before' : 'after'}-${feature.id}`,
-				stage:    6,
+				stage:    OUT_OF_RANGE_STAGE + 1, // So they always match
 			} : {
 				browsers: unsupportedBrowsers,
 				plugin:   plugins[feature.id],
--- a/plugin-packs/postcss-preset-env/src/index.js
+++ b/plugin-packs/postcss-preset-env/src/index.js
@@ -90,7 +90,7 @@ const plugin = opts => {
 			if (isAllowedFeature) {
 				log(`  ${feature.id} has been enabled by options`);
 			} else {
-				log(`  ${feature.id} has been disabled (Stage ${feature.stage})`);
+				log(`  ${feature.id} with stage ${feature.stage} has been disabled`);
 			}
 		}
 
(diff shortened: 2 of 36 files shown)


plugin-packs/postcss-preset-env/README.md
```diff
@@ -354,6 +354,10 @@ postcssPresetEnv({
 });
 ```
 
+### debug
+
+The `debug` option enables debugging messages to stdout which should be useful to help you debug which features have been enabled/disabled and why.
```
romainmenke, line 359: Should we use `result.warn` instead of `stdout`?
This can be done in `Once` or `OnceExit` and will use the plugin name as a prefix.

see : https://github.com/postcss/postcss/blob/b1bfc05bb70e414798af42ee9134d04c4d77285d/docs/guidelines/plugin.md#42-use-resultwarn-for-warnings

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -10,20 +10,38 @@ import getOptionsForBrowsersByFeature from './lib/get-options-for-browsers-by-fe
 import { pluginIdHelp } from './lib/plugin-id-help';
 import { pluginHasSideEffects } from './lib/plugins-with-side-effects';
 
+const DEFAULT_STAGE = 2;
+const OUT_OF_RANGE_STAGE = 5;
```
romainmenke, line 14: Love this!

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -10,20 +10,38 @@ import getOptionsForBrowsersByFeature from './lib/get-options-for-browsers-by-fe
...
+	const insertBefore = Object(options.insertBefore);
+	const insertAfter = Object(options.insertAfter);
+	const browsers = options.browsers;
+	let stage = DEFAULT_STAGE;
+
+	if (typeof options.stage !== 'undefined') {
+		if (options.stage === false) {
+			stage = OUT_OF_RANGE_STAGE;
+		} else {
+			stage = parseInt(options.stage, 10) || 0;
```
romainmenke, line 33: Should we add a lint rule to enforce adding the radix in `parseInt` ?

romainmenke (commented):
Thank you for this!

This will give users a lot more insight in which plugins are enabled and why.

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: Thought about that and there were two main reasons I didn't follow:

First is that this is not a warning but just debug data. 
Second is that it would mean queuing messages to be then spilled at once once it's done rather that as it goes which would be more useful if something odd happened?

Not a big lift though to change it. Let me know your thoughts!

plugin-packs/postcss-preset-env/src/index.js
Antonio-Laguna: Not sure. This was something automatic. Given the radix is 10 by default I can just remove it. Do you have any opinions on this?

romainmenke: radix should always be added as it is not actually 10 by default :)
So your code is more correct 🎉 

I was wondering if we should add a lint rule to make sure it is always `parseInt(foo, 10)`;

plugin-packs/postcss-preset-env/README.md
romainmenke: My main concern is that `stdout` should not be used for any output that isn't part of the main function of the program.

Piping output to a file or another program would become impossible if we add messages to `stdout`. (this would break our CLI for example)

Because of this I looked up the PostCSS API and guidelines and saw the notice to never use `console`. I am not sure which tools do not allow using `console`, and that also seems a bit weird to me.

I think the important bit is to not use `stdout` but maybe we need to ask PostCSS to clarify?

plugin-packs/postcss-preset-env/src/index.js
romainmenke: Added the lint rule, apparently only this file used to have `parseInt(foo)`.
So no other plugins needed fixes.

Antonio-Laguna: Thanks!

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: @romainmenke think you're right. I've added a "simple" and naive approach. Let me know if you see any downsides.

plugin-packs/postcss-preset-env/src/index.js
Antonio-Laguna: I've just added some clamping using this so if you put `144` as a stage you see the appropriate log `Stage has been disabled, features will be handled via the features option.`

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -9,23 +9,24 @@ import writeToExports from './lib/write-to-exports';
...
 	const features = Object(options.features);
-	const debug = options.debug ? logDebug : () => {};
 	const featureNamesInOptions = Object.keys(features);
 	const insertBefore = Object(options.insertBefore);
 	const insertAfter = Object(options.insertAfter);
 	const browsers = options.browsers;
 	let stage = DEFAULT_STAGE;
 
+	// Forcing logs to be emptied between runs
+	dumpLogs();
```
romainmenke, line 28: maybe a `resetLogger` function?
That would not require a code comment :)

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -35,9 +36,9 @@ const plugin = opts => {
 	}
 
 	if (stage === OUT_OF_RANGE_STAGE) {
-		debug('Stage has been disabled, features will be handled via the features option.');
+		log('Stage has been disabled, features will be handled via the features option.');
```
romainmenke, line 39:
```suggestion
		log('Stage has been disabled, features will be handled via the "features" option.');
```

plugin-packs/postcss-preset-env/src/index.js
```diff
@@ -85,12 +86,12 @@ const plugin = opts => {
...
 		if (isDisabled) {
-			debug(`  ${feature.id} has been disabled by options`);
+			log(`  ${feature.id} has been disabled by options`);
 		} else if (!isAllowedStage) {
 			if (isAllowedFeature) {
-				debug(`  ${feature.id} has been enabled by options`);
+				log(`  ${feature.id} has been enabled by options`);
 			} else {
-				debug(`  ${feature.id} has been disabled (Stage ${feature.stage})`);
+				log(`  ${feature.id} has been disabled (Stage ${feature.stage})`);
```
romainmenke, line 94:
```suggestion
				log(`  ${feature.id} with stage ${feature.stage} has been disabled`);
```

romainmenke (commented):
A few small things :)

plugin-packs/postcss-preset-env/src/index.js
Antonio-Laguna: You're more than right!

Antonio-Laguna:
> A few small things :)

Thanks! All done!

romainmenke (approved):
LGTM!

PR #164, opened 2022-01-16
cli : emit warnings

romainmenke:
Warnings were previously thrown away by the cli.
This was not intentional.

This change emits all warnings over `stderr`.

`stdout` is not poluted.

```
echo ".foo:dir(ltr) .baz:dir(rtl) { padding-inline: 20px 40px; }" | npx @csstools/csstools-cli postcss-dir-pseudo-class > out.css
```

In this case `out.css` will not contain the warning.
The warning will appear in the console.


Antonio-Laguna (approved):
Nice catch! Thanks for doing this

PR #166, opened 2022-01-16
postcss tape crafted for this repo

romainmenke:
- cli -> lib
- simplified API
- pretty diff
- github annotations
- test multiple PostCSS versions
- unify PostCSS versions used for tests (this removes `postcss` from plugin dev dependencies)
- verify that the result is still parsable with PostCSS
- verify source maps (no `<no-source>`)
- verify both `commonjs` and `es modules`
- verify [plugin standards](https://github.com/postcss/postcss/blob/main/docs/guidelines/plugin.md#54-include-postcss-plugin-keyword-in-packagejson)

-----

Making this a lib instead of a CLI simplifies a lot of things.
As this is a private package we know that we only use this in ways where a lib is possible.

-----

In general this change trades API complexity for test complexity.

1. The public interface is much smaller and simpler.
2. There are more tests and checks done with the same information.

-----

The tests for the package use `bash` and `git` to avoid adding real code that would itself also need tests.

- outputs on `stdout` and `stderr` for a test run are piped to one file
- exit code for a test run is piped to another file

These are compared with `git diff` to check if the new PostCSS Tape handles each case as expected.

-----

I have only updated a couple of plugins to use the new test suite.
I do not want to make review of this package harder by having too many changes at once.
A follow-up PR will migrate all other plugins.

If I maybe missed some part of the original PostCSS Tape that is used by a plugin, please let me know.

We could also open up select CI checks to forks.
This combined with the improved feedback will make it easier for contributors to iterate without needing to wait for maintainer feedback.


UPDATING_POSTCSS.md
```diff
@@ -0,0 +1,12 @@
+# Updating PostCSS
+
+see : `packages/postcss-tape/package.json`
+
+Set the `postcss` versions to match the peer dependency of all plugins.
+
+- Plugins use `^MAJOR.MINOR`.
+- PostCSS Tape uses `~MAJOR.MINOR`.
+
+Avoid raising these versions when possible.
```
romainmenke, line 10:
```suggestion
Avoid increasing these versions when possible.
```

packages/postcss-tape/src/index.ts
```diff
@@ -0,0 +1,344 @@
...
+
+			// Assert result with oldest supported PostCSS.
+			// Check against the actual result to avoid duplicate warnings.
+			//
+			// NOTE:
+			// The oldest version of PostCSS we support at this time.
+			// This does not need to be a different version than the latest version of PostCSS.
+			// There is no system behind setting this.
+			// It is here to allow testing a specific older version, if parts of the community can't update yet.
+			{
```
Antonio-Laguna, line 272: I completely and wholeheartedly love this feature. Thanks for thinking so much ahead of time. ❤️

That said, should we check versions before just in case we already ran this? I see no benefit in running this 3 times with the same version. We'd be "wasting" GH resources and eventually could be capped out.

I reckon this should be a first start: https://postcss.org/api/#processor-version ?

packages/postcss-tape/src/index.ts
```diff
@@ -0,0 +1,344 @@
...
+					}
+				}
+			}
+
+			// Assert that warnings have the expected amount.
+			{
+				try {
+					if (result.warnings().length || testCaseOptions.warnings) {
+						assert.strictEqual(result.warnings().length, testCaseOptions.warnings);
+					}
```
Antonio-Laguna, line 304: Should this be averaged between all of the runs so we make sure it's not warning anywhere else?

packages/postcss-tape/src/index.ts
```diff
@@ -0,0 +1,344 @@
...
+		'packages',
+		'plugins',
+		'plugin-packs',
+		'cli',
+		'experimental',
+	];
+
+	if (!workspaces.includes(parts[0])) {
+		throw new Error('PostCSS Tape was intended to be run from <workspace>/package');
+	}
```
Antonio-Laguna, line 341: Perhaps this is down the line 😂  but what about testing with the rest of the environments?

packages/postcss-tape/src/index.ts
romainmenke: I totally forgot we can do that!
Great idea!

Antonio-Laguna (commented):
🤯

This is fantastic! I wasn't sure what to expect but this definitely surpassed any expectations I might have had! Meta tests are great too :) 

I've left some minor suggestions around optimizing runtime and one concern with Node versions. This sorts of patch into sharing code/package within the monorepo without workspaces.

packages/postcss-tape/src/index.ts
romainmenke: Do you mean averaging between possible older en current postcss versions?

-------

_if not, you can ignore the bit here_

I have been trying to think of a case were this could differ and I can only think of things that would be intentional. (warning users that they should update postcss because feature X is required or gives a better result)

The test against older postcss is intended to catch exceptions that are thrown when using newer API's.

Antonio-Laguna: Yeah, actually thought the same, just wanted to ponder the question. I agree that's the only usecase that comes to mind, throwing warnings for deprecations, etc.

packages/postcss-tape/src/index.ts
romainmenke: My comment is perhaps confusing and should be updated to `<workspace>/<package>`

I might have been a bit lazy here, but there is currently no good way (afaik) to get a file path relative to an npm workspace root.

Since this can differ widely between local setups and also CI, I look for a pattern with a prefix of the known npm workspaces.

For github annotations we have to log file paths relative to the repo root which is the same as the workspace root.

From `/Users/<you>/some-projects-dir/postcss-plugins/plugins/postcss-is-pseudo-class/test/basic.expect.css` I want to get just `plugins/postcss-is-pseudo-class/test/basic.expect.css`.

This heuristic limits us in a way.
We can't do this without updating the code here :

- `plugins/plugins-for-color/css-color-4`
- `plugins/plugins-for-color/css-color-5`

Currently we don't have a need for this hierarchy, so I am hoping it's fine for now :)

romainmenke: _my comments above can be ignored_

----

Discussed : current setup didn't work in node 12/14
I had forgotten about this.

Are we ok with doing this for now?

```js
import postcssTape from '../../packages/postcss-tape/dist/index.mjs';
```

It works in all node versions and doesn't require publishing the test suite.

I would prefer this as I want to have the freedom to break the API when needed without needing to consider other users.

Antonio-Laguna: I'm completely ok with this approach and also helps with my point above (sharing code)

One other option is https://www.npmjs.com/package/module-alias

packages/postcss-tape/src/index.ts
romainmenke: This has been implemented :)

packages/postcss-tape/src/index.ts
romainmenke: I added a note in the README about how to clean this up after node 20 is released.
That should make it easy to move fully to workspaces then.

Antonio-Laguna (approved):

PR #182, opened 2022-01-20
apply new test suite to more plugins and sourcemap fixes

romainmenke:
- `postcss-preset-env` and a few more plugins now use the new test suite.
- fixed sourcemaps in `image-set-function`
- some plugins can throw, extended `postcss-tape` to handle this

I also created an issue to remove these exceptions as it goes against plugin best practices.

### DIFF
--- a/packages/postcss-tape/src/index.ts
+++ b/packages/postcss-tape/src/index.ts
@@ -6,7 +6,7 @@ import path from 'path';
 import fs, { promises as fsp } from 'fs';
 import { strict as assert } from 'assert';
 
-import type { PluginCreator, Plugin } from 'postcss';
+import type { PluginCreator, Plugin, Result } from 'postcss';
 import { formatGitHubActionAnnotation } from './github-annotations';
 import { dashesSeparator, formatCSSAssertError, formatWarningsAssertError } from './format-asserts';
 import noopPlugin from './noop-plugin';
@@ -20,6 +20,9 @@ type TestCaseOptions = {
 	plugins?: Array<Plugin>,
 	// The expected number of warnings.
 	warnings?: number,
+	// Expected exception
+	// NOTE: plugins should not throw exceptions, this goes against best practices. Use `errors` instead.
+	exception?: RegExp,
 
 	// Override the file name of the "expect" file.
 	expect?: string,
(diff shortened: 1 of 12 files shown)


plugins/postcss-nesting/.tape.mjs
```diff
@@ -6,7 +6,7 @@ const mixinPluginRule = () => {
 		postcssPlugin: 'mixin',
 		AtRule: {
 			mixin(node, { postcss }) {
-				node.replaceWith(postcss().process('& .in{ &.deep { color: blue; }}', {from : 'mixin.css'}).root);
+				node.replaceWith(postcss.parse('& .in{ &.deep { color: blue; }}', {from : 'mixin.css'}));
```
romainmenke, line 9: This was triggering PostCSS to warn about now having any plugins.
Using `.parse` resolves this.

plugins/postcss-image-set-function/src/lib/get-media.ts
```diff
@@ -15,6 +15,7 @@ export function getMedia(dpi: number | false, postcss) {
 	const media = postcss.atRule({
 		name: 'media',
 		params: `(-webkit-min-device-pixel-ratio: ${dpr}), (min-resolution: ${dpi}dpi)`,
+		source: decl.source,
```
romainmenke, line 18: This fixes the sourcemaps issue

packages/postcss-tape/src/index.ts
```diff
@@ -87,7 +90,13 @@ export default function runner(currentPlugin: PluginCreator<unknown>) {
 
 		// https://github.com/postcss/postcss/blob/main/docs/guidelines/plugin.md#11-clear-name-with-postcss--prefix
 		// Clear name with postcss- prefix
-		if (!packageInfo.name.startsWith('postcss-') && !packageInfo.name.startsWith('@csstools/postcss-')) {
+		const isOlderPackageName = [
+			'css-has-pseudo',
+			'css-blank-pseudo',
+			'css-prefers-color-scheme',
+		].includes(packageInfo.name);
```
Antonio-Laguna, line 97: Should we deprecate and republish?

romainmenke: I think very few people use these as a direct dependency.
Looking at the download numbers on npm seems to confirm this.

I think we can deprecate and republish without causing much disruption.

`:has` already has a breaking change planned, and `prefers-color-scheme` could use one to clean up old code.

I'll create an issue for this

Antonio-Laguna (approved):
Looks great!

PR #183, opened 2022-01-20
:any-link fixes

romainmenke:
- fixes failures in IE/Edge with `<area href>` see : [web-tests](https://mrhenry.github.io/web-tests/#75fe312b-386d-4648-b6cf-1a588a5cee3b)
- fixes `:not(:any-link)`
- ~~uses `:is` pseudo to reduce plugin complexity~~

------

- [x] ~~document that `:is` will be used and link to that plugin in `README.md`~~
- [x] changelog
- [x] update preset-env to toggle the settings
- [x] update preset-env tests

------

Decided against using `:is` as we have seen in the past that making another plugin required leads to frustrations with users. (see `logical` and `:dir`)

Making the plugin just work even if we have some code duplication is preferable imho.

A small benefit of the current implementation is that `:any-link` doesn't transform into duplicate rules like it would with `:is`.

`:is` is a forgiving selector list and each list item needs to be in its own rule to achieve this. The only browser that supports `:link` but not `:visited` and would benefit from this is IE3. So CSS becomes bigger with an `:is` based implementation without any real world benefit.

-----

`postcss-tape` now emits a summary if there were failures :

I found that the new tape is really great to inspect unexpected changes.
But when I knowingly change something that affects `postcss-preset-env`, I just want to know which cases I need to update.

```
----------------------------------------

unexpected failures:
  - basic:stage0
  - basic:autoprefixer
  - basic:autoprefixer:false
  - basic:export
```


Antonio-Laguna (approved):
Nice work!

PR #70, opened 2021-12-13
css-color-4 & css-color-5

romainmenke:

PR #184, opened 2022-01-20
postcss-hwb-function

romainmenke:
see : https://github.com/csstools/postcss-plugins/pull/70

_splitting that one in smaller bits for easier review_

----------------

prior work : https://github.com/postcss/postcss-color-hwb

This plugin differs in that it does not accept comma separated values:
`hwb(90, 0%, 0%, 0.5);`

That notation is not covered by the latest spec.

----------------

This change adds only `hwb()`.

Largely follows the methods in `lab` and `functional-color-notation`.


Antonio-Laguna (approved):
Impressive as usual!

Do you plan to enable this too for 7.3?

romainmenke:
Would be great if we can include this in `7.3`!

PR #185, opened 2022-01-20
Removing Can I Use Lite

### LINKED ISSUE #131: Remove Can I Use Lite

Antonio-Laguna:
Fixes #131


Antonio-Laguna:
@romainmenke this needs 6.0.2 but if you could use pulls from here to override locally and check it makes sense that'd be great:

https://github.com/csstools/cssdb/pull/77

I _think_ this is mostly what we've been talking about and nothing else new/strange but want an extra pair of eyes before moving forward with this.

Antonio-Laguna:
Tests are expected to fail, they'll come back green with CSSDB 6.0.2

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
```diff
@@ -1,26 +1,41 @@
...
-
 	// if feature support can be determined
-	if (caniuseFeature) {
-		const stats = caniuse.feature(caniuseFeature).stats;
+	if (feature.browser_support) {
+		const query = [];
+
+		browsers.forEach(browser => {
+			const browserSupport = feature.browser_support[browser];
+			// If the version is something like TP we can't do < TP
```
romainmenke, line 28: Are pre-release versions counted as vendor implementations in cssdb?

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
```diff
@@ -1,26 +1,41 @@
...
-				).map(
-					version => `${browser} ${version}`,
-				),
-			),
-			[],
-		);
+			if (isValid) {
+				query.push(`${browser} < ${feature.browser_support[browser]}`);
+			} else {
+				query.push(`${browser} >= 1`);
```
romainmenke, line 34: What is the current behaviour if you pass browser to postcss-preset-env that doesn't have stats for a certain feature?

Is it enabled or disabled?

romainmenke (commented):
Some small questions, will take a better look tomorrow!

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
Antonio-Laguna: Yes they are, which is not right. Will fix that, thanks for flagging!

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
Antonio-Laguna: This happens with `custom-properties`, `ie` does not appear so that means every IE version is added to the "unsupported list". If the list matches any `ie` it will be polyfilled.

Browser support for custom-properties:

```json
"browser_support": {
  "edge": "16",
  "firefox": "31",
  "chrome": "49",
  "safari": "10",
  "opera": "36",
  "ios_saf": "10",
  "android": "97",
  "op_mob": "64",
  "and_chr": "97",
  "and_ff": "95",
  "and_uc": "12.12",
  "samsung": "5",
  "kaios": "2.5"
}
```

Test:

```
'custom-properties': {
	message: 'supports { browsers: "ie >= 10" } usage',
	options: {
		browsers: 'ie >= 10'
	}
}
```

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
Antonio-Laguna: This has been fixed, only affected `:has`.

romainmenke: Nice!

plugin-packs/postcss-preset-env/src/lib/get-unsupported-browsers-by-feature.js
romainmenke: Thank you for checking this!

PR #193, opened 2022-01-22
add a shortcut to rewrite expect files and log warnings

romainmenke:
_adding it towards 7.3 because `postcss-tape` was touched there_

Running this will override all `expect` files to match `result` files.
The tests will still fail the first time as the original `expect` contents have been read into memory by the time this shortcut runs.

`npm run test:rewrite-expects`

------------

Warnings are now logged when the amounts do not match.


romainmenke:
@Antonio-Laguna If approved you can merge this so that you can leverage these in your work.

Antonio-Laguna (approved):
Thanks for this, very very helpful!

PR #198, opened 2022-01-23
Adding PostCSS clamp

### LINKED ISSUE #114: Add `postcss-clamp`
https://github.com/polemius/postcss-clamp

> PostCSS plugin to transform clamp() to combination of min/max

Antonio-Laguna:
Fixes #114 

---

One thing to note is that if the feature is unsupported completely (not even min/max) it still applies the transform. Not really sure about this though. Not harmful though


romainmenke (approved):
LGTM!

PR #199, opened 2022-01-23
postcss-is-pseudo-class : better warnings and skip

romainmenke:
- remove `skip` flag and make it's behaviour the default
- always add `warning` flag in `postcss-preset-env`
- add guards for infinite recursion
- add another complex selector transform

This change makes the plugin even more cautious than before.
(applying what we learned from `nesting`)

If we introduce a transform that isn't spec compliant it is technically a bug, but this can be perceived as a feature.

It is better not to transform so that we don't create cases were users depend on the bug. 

----

The main feature for now in this plugin is resolving forgiving lists.

```css
:is(input, .input) {}

:is(:focus, :focus-visible) {}
```

With only very limited support for complex and compound selectors.


package-lock.json
```diff
@@ -6405,34 +6405,6 @@
...
-      "dependencies": {
-        "postcss-value-parser": "^4.2.0"
-      },
-      "engines": {
-        "node": "^12 || ^14 || >=16"
-      },
-      "peerDependencies": {
-        "postcss": "^8.3"
-      }
-    },
```
romainmenke, line 6435: This was causing issues as it made `npm install` and `npm ci` to add the published version in `postcss-preset-env`.

I did a reset of the lock file to resolve this.

Antonio-Laguna (approved):
This looks good to me. Thanks for leading this!

PR #205, opened 2022-01-25
Integrate font & display plugins

### LINKED ISSUE #112: Add `postcss-normalize-display-values`
https://github.com/cssnano/cssnano/tree/master/packages/postcss-normalize-display-values

> Normalize display property values with PostCSS.
> ### Input
> 
> ```css
> div {
>     display: inline flow-root
> }
> ```
> 
> ### Output
> 
> ```css
> div {
>     display: inline-block
> }
> ``` 

Spec: https://developer.mozilla.org/en-US/docs/Web/CSS/display

### LINKED ISSUE #116: Add postcss-font-format-keywords
Seems like [postcss-font-format-keywords](https://github.com/valtlai/postcss-font-format-keywords) provides a syntax sugar from the CSSWG spec. Should we add it to polyfills pack?

cc @valtlai

Antonio-Laguna:
Fixes #116 
Fixes #112

### DIFF
--- a/cli/csstools-cli/CHANGELOG.md
+++ b/cli/csstools-cli/CHANGELOG.md
@@ -3,12 +3,16 @@
 ### Unreleased (minor)
 
 - Emit warnings from plugins
-- Added `@csstools/postcss-is-pseudo-class`
+- Added `@csstools/postcss-font-format-keywords`
 - Added `@csstools/postcss-hwb-function`
+- Added `@csstools/postcss-is-pseudo-class`
+- Added `@csstools/postcss-normalize-display-values`
 
 ```sh
-npx -y @csstools/cli postcss-is-pseudo-class
+npx -y @csstools/cli postcss-font-format-keywords
 npx -y @csstools/cli postcss-hwb-function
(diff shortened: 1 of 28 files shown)


cli/csstools-cli/CHANGELOG.md
```diff
@@ -3,12 +3,16 @@
 ### Unreleased (minor)
 
 - Emit warnings from plugins
-- Added `@csstools/postcss-is-pseudo-class`
+- Added `@csstools/postcss-font-format-keywords`
```
romainmenke, line 6: This also needs a change in `cli/csstools-cli/src/cli.ts`

1. add the plugin in the `switch` statement
2. add the key from the `switch` statement in the general help output

The general help also exists in the README, but this doesn't always have to be up to date as it's an example

romainmenke (commented):
A few small things for the CLI

cli/csstools-cli/CHANGELOG.md
Antonio-Laguna: Thanks for reviewing that! I've added those!

romainmenke (approved):
LGTM!

PR #206, opened 2022-01-25
Update basic.css

romainmenke:
@Antonio-Laguna Can you check these diffs (annotations)? It seems the regexp based find/replace in this plugin doesn't work well enough :/


Antonio-Laguna:
@romainmenke thanks a lot for finding this. It actually took me a while to figure out what was going on there!

PR #207, opened 2022-01-26
Adding more to the README

### LINKED ISSUE #202: [postcss-preset-env] needs documentation clarification
Hello,
I'm checking this project for the first time, and I didn't find the answer in your documentation.
Can you tell me what is the difference between the plugin postcss-preset-env and autoprefixer ?

Or can you correct me if my analysis is wrong ?
cssdb is a comprehensive list of CSS features and their positions in the process of becoming implemented web standards.

"Can I use" provides up-to-date browser support tables for support of front-end web technologies on desktop and mobile web browsers.

"Can I use"  is for browser implementation while CSSDB is for “on which state is this CSS feature related to the standardization, no matter some browser already support it or not". Said differently, cssdb is a comprehensive list of CSS features and their positions in the process of becoming implemented web standards. Before being standardized(stage 4) a css feature go through 3 stages(Aspirational, Experimental, Allowable) or Rejected https://github.com/csstools/cssdb/blob/main/STAGES.md

So if I understand well, PostCSS Preset Env allows to convert some css proposals before being validated to css compatible with targetted browsers using polyfills provided by cssdb.

am I right ?

Antonio-Laguna:
Fixes #202

Take this as a re-iteration of the current README for Preset-Env. I was even thinking about bringing features by id as a first list that links to all the plugins but wanted to spark discussion first :)


plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,17 @@
 can understand, determining the polyfills you need based on your targeted
 browsers or runtime environments.
 
-```bash
-npm install postcss-preset-env
-```
+## How does it work?
+
+[PostCSS Preset Env] is a Plugin Pack for [PostCSS]. It leverages the list of the features we keep an eye from [CSSDB][cssdb] and applies plugins, so you can use those new features without having to worry about browser support.
```
romainmenke, line 14: Should we add a small intro with keywords and what they mean in the context of this project?

Some things that could be confusing or vague :

- browser list option vs. browser support stats
- css feature vs. csstools plugin
- plugin

This will help people who are not yet familiar with polyfills and fallbacks in general while also serving as a guide for us to be consistent.

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,17 @@
...
+
+CSSDB exposes the browser support that each feature has which can come from [Can I Use](https://caniuse.com/css-all) or from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/all) (through [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)).
+
+However, if your site doesn't need to support older browsers you wouldn't want for certain features to be expanded, effectively increasing your CSS bundle without any real benefit.
```
romainmenke, line 18: Maybe we can omit this here and save it for a more dedicated doc or entry focused on performance? I don't disagree with the statement but feel it needs more context.

The one clear value proposition that any project like this has is making it cheaper and simpler for developers to target a wider range of browsers and in doing so making the web more open and accessible.

The few kb saved (after gzip) are often negligible even when comparing something like IE11 and latest Chrome.

I also want to avoid people setting targets that are too modern for performance gains that are maybe too small to measure.

romainmenke (commented):
Thank you for this!
Great intro for the project and this will help a lot of users understand it all beter :)

----

Some first thoughts, still need to do a proper check tomorrow!

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: I love that idea!

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: The reasoning for this sort of section is to explain why does it need a browser list or why would you want to be accurate with it. I don't think right now was clear what was going on. I'm keen on rewording it though! Let's connect on this.

romainmenke: So true!
Maybe we can focus on how it is related to `stage` without mentioning bundle size?

(`stage` enables features and `browsers` disables them.)

So maybe something like this :

By providing a list of browser targets for your project, plugins that aren't needed will be skipped. Over time your targets will change and by updating the settings your CSS bundle will only ever contain the needed fallbacks.

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,26 @@
...
-```bash
-npm install postcss-preset-env
-```
+## How does it work?
+
+[PostCSS Preset Env] is a Plugin Pack for [PostCSS]. It leverages the list of the features we keep an eye from [CSSDB][cssdb] and applies plugins, so you can use those new features without having to worry about browser support.
+
+CSSDB exposes the browser support that each feature has which can come from [Can I Use](https://caniuse.com/css-all) or from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/all) (through [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)).
+
+By providing a list of browser targets for your project, plugins that aren't needed will be skipped. Over time your targets will change and by updating the settings your CSS bundle will only ever contain the needed fallbacks.
```
romainmenke, line 18:
```suggestion
By providing a list of browser targets for your project, plugins that aren't needed will be skipped. Over time your targets might change and by updating the settings your CSS bundle will only ever contain the needed fallbacks.
```

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,26 @@
...
+
+What [PostCSS Preset Env] does is to take that support that comes from MDN and Can I Use and guess from a [browserlist](https://github.com/browserslist/browserslist) whether those transformations are needed. It does also pack [Autoprefixer](https://github.com/postcss/autoprefixer) within and shares the list with it so certain prefixes aren't applied unless you're going to need them given your browser support list.
```
romainmenke, line 20:
```suggestion
What [PostCSS Preset Env] does is to take the support data that comes from MDN and Can I Use and determine from a [browserlist](https://github.com/browserslist/browserslist) whether those transformations are needed. It also packs [Autoprefixer](https://github.com/postcss/autoprefixer) within and shares the list with it, so prefixes are only applied when you're going to need them given your browser support list.
```

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,26 @@
...
+
+There are some terms that might be confusing here, let's go over them:
```
romainmenke, line 22:
```suggestion
Glossary:
```

I know you don't mean it like that at all, but might have sounded a bit patronising :D

`Glossary` is fully neutral

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,26 @@
...
+
+* **Browser list option**: [Browserlist](https://github.com/browserslist/browserslist) is a package that lets you create a query. For example, `chrome < 42` will give you a list of every Chrome version that has been released up to, but not including, 42.
```
romainmenke, line 24:
```suggestion
* **Browser list option**: [Browserlist](https://github.com/browserslist/browserslist) is a package that gives you a list of browsers for a given query. For example, `chrome < 42` will give you a list of every Chrome version that has been released up to, but not including, 42.
```

plugin-packs/postcss-preset-env/README.md
```diff
@@ -9,9 +9,26 @@
...
+* **Browser support stats**: Features get introduced on browsers at certain versions. They're often visible on [MDN](https://developer.mozilla.org/en-US/) and [Can I Use](https://caniuse.com/) and let you know if can safely use a given feature or not, given the _support_ you need on your project / page.
```
romainmenke, line 25:
```suggestion
* **Browser support stats**: Features get introduced on browsers at certain versions. They're often visible on [MDN](https://developer.mozilla.org/en-US/) and [Can I Use](https://caniuse.com/). Comparing these stats with the needed _support_ for your project tells you if it's safe to use a feature or not.
```

romainmenke (commented):
Some small re-wordings :)
Very nice!

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: Great suggestion!

plugin-packs/postcss-preset-env/README.md
Antonio-Laguna: Thanks for this!

romainmenke (approved):
LGTM! 🎉

PR #211, opened 2022-01-29
postcss-preset-env: filter features by implementation status

romainmenke:
**Note on the size of this change :**

I did not dare adding more logic to preset-env without having more granular unit tests.
This required splitting everything up into separate functions and files.

My aim was to make the entire code base more clear and easier to update with [this](https://github.com/csstools/postcss-plugins/discussions/138) in mind.

Some newer JS features like `Symbol` and `Map` are used to make everything a bit more predictable.

- `Symbol` makes it safer to combine data from multiple source (plugin options, cssdb) while still knowing which keys were set by `postcss-preset-env`.
- `Map` prevents unexpected behaviour when users set options that overlap with build-ins like `toString`.

This does not yet have full test coverage with unit tests alone but the combo of the new unit tests and `postcss-tape` should still be a step in the right direction.

------------

After this initial refactor it became trivial to add the `minimumVendorImplementations` option.

There is no overloading of this option or hidden logic around it.

`minimumVendorImplementations` is always converted to a number between `0` and `3`.
`0` is equivalent to not setting the option.

------

There is still one bit of undefined logic that I haven't tracked down and added tests for.

When using `insertAfter` or `insertBefore` it is unclear if those plugins become dynamic based on the browser lists of the feature they are attached to.

I need to check this on `main` first.

**Update :**

- added tests in main and merged them into this branch
- behaviour is to always run the inserted plugins, these are not dynamic

---------

**Notable fix :**

see : https://github.com/csstools/postcss-plugins/blob/e83f3558d0c14c85d2f6f120157a5d84f01128b4/plugin-packs/postcss-preset-env/test/basic.nesting.true.expect.css#L39-L41

We added `getOptionsForBrowsersByFeature` some time ago and this had a bug.

```js
if (features[feature.id] === true) {
	// if the plugin is enabled
	options = sharedOpts ? Object.assign({}, options, sharedOpts) : undefined;
}
```

When manually enabling a feature like `postcss-nesting` but without adding any `sharedOpts` the flags set by `getOptionsForBrowsersByFeature` would be erased. As `options` was set to `undefined` in this case.

This has been fixed.


Antonio-Laguna (approved):
Wow! This is really great!

Thanks for cleaning this up and making this much more manageable and sustainable in the future.

Only question I have is if you want to clarify more around the `vendorImplementations` as in explaining what stability usually means?

Antonio-Laguna (approved):
❤️

PR #212, opened 2022-01-30
postcss-is-pseudo-class : browser tests and transform in 2 stages

romainmenke:
The previous version wasn't able to transform nested `:is` easily without causing infinite loops.

Finally had an idea to work around this that is more solid :

- transform lists firsts and replace `:is` with something else `:-csstools-matches`. This clearly separates touched vs untouched, making recursion much easier.
- handle removing `:is` or `:-csstools-matches` in a second step.

This also made the code easier to split in smaller, more understandable parts.

For debugging I also found it handy to disable the second stage and see the output from only the first.

This doesn't break any existing behaviour and adds a few more cases it can handle :

- `:is()` -> `:not(*)`
- `:is(<complex selector>)` -> `<complex selector>` (no parts before or after `:is()`)

### DIFF
--- a/plugins/postcss-is-pseudo-class/.tape.mjs
+++ b/plugins/postcss-is-pseudo-class/.tape.mjs
@@ -19,13 +19,19 @@ postcssTape(plugin)({
 	},
 	'basic:oncomplex:warning': {
 		message: "supports basic usage with { onComplexSelector: 'warning' }",
-		warnings: 8,
+		warnings: 11,
 		options: {
 			onComplexSelector: 'warning'
 		}
 	},
 	'browser': {
 		message: "prepare CSS for chrome test",
+		options: {
+			preserve: false
+		}
+	},
+	'complex': {
+		message: "supports complex selectors",
 		options: {
 			preserve: false
 		}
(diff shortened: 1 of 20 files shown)


plugins/postcss-is-pseudo-class/.tape.mjs
```diff
@@ -19,11 +19,23 @@ postcssTape(plugin)({
 	},
 	'basic:oncomplex:warning': {
 		message: "supports basic usage with { onComplexSelector: 'warning' }",
-		warnings: 8,
+		warnings: 11,
```
romainmenke, line 22: Recursion is handled better, so the result has more selectors with `:is` remaining, so more warnings.

The previous version halted transforms earlier.
This is a good thing.

plugins/postcss-is-pseudo-class/test/browser.expect.css
```diff
@@ -0,0 +1,178 @@
+/* https://github.com/web-platform-tests/wpt/blob/master/css/selectors/is-nested.html */
```
romainmenke, line 1: There are still instances of `:is` in this file and that is ok for now.
These are taken from WPT and currently pass in latest chrome.

This tells us that what we transform doesn't break things.
The remaining untransformed bits work because latest chrome natively supports `:is`.

This will be a good guide going forward to make sure that future transforms for complex selectors do not introduce issues.

Antonio-Laguna (approved):
Thanks for being so thorough again, specially with all of the tests!

PR #213, opened 2022-01-30
postcss-preset-env: pin some features to stage: 1

romainmenke:
makes : https://github.com/csstools/cssdb/pull/78 non-breaking


Antonio-Laguna (approved):
LGTM, if this becomes larger before major we might consider an array to just check if it's included

PR #217, opened 2022-01-31
postcss-preset-env : make logger a class instance

romainmenke:
This adds support for debug logs when doing multiple parallel runs.


Antonio-Laguna (approved):
Great! Thanks for taking care of this!

PR #219, opened 2022-01-31
Update ci to fail in case we add more packages that don't support node 12/14

romainmenke:
failing run : https://github.com/csstools/postcss-plugins/actions/runs/1772295768

This was before `postcss-opacity-percentage` was updated.

### DIFF
--- a/packages/postcss-tape/package.json
+++ b/packages/postcss-tape/package.json
@@ -28,8 +28,8 @@
     "test": "bash ./test-self/test.sh"
   },
   "dependencies": {
-    "postcss-8.2": "npm:postcss@~8.2",
-    "postcss": "~8.4"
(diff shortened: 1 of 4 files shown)


packages/postcss-tape/package.json
```diff
@@ -28,8 +28,8 @@
     "test": "bash ./test-self/test.sh"
   },
   "dependencies": {
-    "postcss-8.2": "npm:postcss@~8.2",
-    "postcss": "~8.4"
+    "postcss": "~8.4",
+    "postcss-8.2": "npm:postcss@~8.2"
```
romainmenke, line 32: this was ordered by npm

Antonio-Laguna (approved):
Thanks for this!

PR #222, opened 2022-02-02
ci : add e2e tests for bundling `postcss-preset-env` with webpack

romainmenke:
This is not a test for using `postcss-preset-env` in a Webpack build.
This tests bundling `postcss-preset-env` with Webpack.

------

- all plugins are now written as `es modules` (they use `import` and `export`)
- all plugins have an `exports` field in `package.json`


.github/workflows/test.yml
```diff
@@ -50,8 +50,8 @@ jobs:
       - run: npm install --ignore-scripts
       - run: npm run test:cli --workspace="@csstools/postcss-base-plugin"
 
-  # Run Deno tests one by one
-  test_deno:
+  # Run e2e tests
```
romainmenke, line 53: Deno tests are also a form of e2e.
So made sense to me to group these.

One less full build of all plugins if we have then in a single job.

e2e/webpack/bundle-through/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "build": "webpack --config webpack.config.js",
+    "test": "node ./dist/bundle-from-mjs.cjs && node ./dist/bundle-from-cjs.cjs "
+  },
+  "dependencies": {
+    "postcss": "^8.4.6",
+    "webpack": "^5.68.0",
+    "webpack-cli": "^4.9.2",
+    "postcss-preset-env": "file:../../../plugin-packs/postcss-preset-env"
+  }
+}
```
romainmenke, line 17: This is not part of the npm workspaces but it can be.

Instinctively I did not want to "pollute" the workspaces with dependencies for e2e tests.

I don't have a technical reason for this and it might make more sense to make it workspace.

e2e/webpack/bundle-through/webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+		target: 'node',
+		optimization: {
+			minimize: false,
+		},
+		output: {
+			path: path.resolve(__dirname, 'dist'),
+			filename: 'bundle-from-cjs.cjs',
+		},
+	},
+];
```
romainmenke, line 28: I am not a Webpack wizard but I managed to make it do two things :
- take `esm` and bundle
- take `cjs` and bundle

The output is always a commonjs bundle.
I have no idea how to make Webpack output `esm` and reading their docs sucks the life out of me :D

plugin-packs/postcss-preset-env/src/lib/format-feature.mjs
```diff
@@ -51,8 +51,10 @@ export function formatStagedFeature(cssdbList, browsers, features, feature, shar
...
 	}
 
-	if (feature.plugin.postcss) {
+	if (feature.plugin.postcss && typeof feature.plugin === 'function') {
 		plugin = feature.plugin(options);
+	} else if (feature.plugin && feature.plugin.default && typeof feature.plugin.default === 'function' && feature.plugin.default.postcss) {
+		plugin = feature.plugin.default(options);
 	} else {
 		plugin = feature.plugin;
 	}
```
romainmenke, line 60: Account for the edge case where the imported/required package is an object with a key `default`.

e2e/webpack/bundle-through/webpack.config.js
Antonio-Laguna: 🧙🏼‍♂️

Antonio-Laguna (approved):
Thanks a lot for doing this!

I think keeping it out of the workspace makes sense and moving Deno there also makes sense.

Merging!

PR #226, opened 2022-02-05
postcss-color-function

romainmenke:
see : https://github.com/csstools/postcss-plugins/pull/70

This adds transforms for `color()` following https://www.w3.org/TR/2021/WD-css-color-4-20211215/

- [x] updated labeler
- [x] updated issue templates
- [x] plugin README.md and INSTALL.md
- [x] checked all links (trying to have less broken links on first release :) )
- [x] exists in cssdb : https://github.com/csstools/cssdb/pull/81


plugins/postcss-color-function/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+function modifiedValues(originalValue: string, decl: Declaration, result: Result): {rgb: string} | undefined {
+	let valueASTSRgb: ParsedValue | undefined;
+
+	try {
+		valueASTSRgb = valueParser(originalValue);
+	} catch (error) {
+		decl.warn(
+			result,
+			`Failed to parse value '${originalValue}' as a lab or hcl function. Leaving the original value intact.`,
```
romainmenke, line 73:
```suggestion
			`Failed to parse value '${originalValue}' as a color function. Leaving the original value intact.`,
```

plugins/postcss-color-function/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+	const preserve = 'preserve' in Object(opts) ? Boolean(opts.preserve) : false;
+
+	return {
+		postcssPlugin: 'postcss-color-function',
+		Declaration: (decl: Declaration, { result }: { result: Result }) => {
+			if (hasFallback(decl)) {
+				return;
+			}
+
+			if (hasSupportsAtRuleAncestor(decl)) {
```
romainmenke, line 20:
```suggestion
			if (preserve && hasSupportsAtRuleAncestor(decl)) {
```

plugins/postcss-color-function/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+			});
+
+			if (atSupportsRules.length === 0) {
+				return;
+			}
+
+			atSupportsRules.reverse().forEach((atSupports) => {
+				rule.after(atSupports);
+			});
+		},
```
romainmenke, line 80: I am unsure about this bit.

```css
:root {
  --one-var: red;
  --one-var: color(srgb 0.64331 var(--point-5) 0.16771 / var(--opacity-50));
}
```

As setting the same custom property twice doesn't make sense the second is wrapped in `@supports`.

It is assumed that one is a fallback created by a plugin with `preserve: true`;

It is however not limited to `color`.

Should this be a separate plugin?
Or is it ok if we have this logic in multiple plugins?

I don't see how it can cause harm, I am just not a 100% happy with it.

Antonio-Laguna: This is an interesting thought and I feel like you. How does it sound about doing something such as what you did to avoid infinite loops? Like an interim step that another plugin can then take over (only within preset env)

plugins/postcss-color-function/test/basic.css
```diff
@@ -0,0 +1,178 @@
...
+	/* out */
+	prop-14: color(display-p3 0.99687 0.35134 0.71095);
+
+	/* https://romainmenke.github.io/lab-lch-display-p3/lab-interactive-b.html#%5B%5B60%2C-28%2C15%5D%2C%5B65%2C81%2C-11%5D%5D */
+	/* in */
+	prop-15: color(display-p3 0.43165 0.61289 0.47061);
+	/* out */
+	prop-16: color(display-p3 0.99937 0.35096 0.69833);
+
+	/* https://romainmenke.github.io/lab-lch-display-p3/lab-interactive-b.html#%5B%5B78%2C-64%2C-44%5D%2C%5B76%2C81%2C-11%5D%5D */
```
Antonio-Laguna, line 123: ❤️ these

Antonio-Laguna (approved):
👏 Hats off to you! This is impressive and enables this beautifully. 

Let me know what you think about my comment, could be dealt with afterward too.

plugins/postcss-color-function/src/index.ts
romainmenke: From a technical standpoint an interim step would not be needed.

Maybe one of these :
- separate plugin and list it as a peer dependency in package.json?
- separate plugin and make each that needs it a tiny plugin pack with the option to disable

romainmenke: A separate plugin would remove the duplication (each value transform plugin needs this logic).

And it would also resolve all the bug reports we have seen in preset-env and plugins about broken vars.

I think it is ok to have "utility" plugins in preset-env that cleanup and fix css.
As long as these are limited to such technical issues caused by plugin transforms and do not aim to fix stylesheet author issues.

Antonio-Laguna (approved):

PR #229, opened 2022-02-06
postcss-progressive-custom-properties

romainmenke:
see : https://github.com/csstools/postcss-plugins/pull/226#discussion_r800079410

-----

A utility plugin to fix fallbacks and enhancements in Custom Properties for all plugins that transform values.

This is only the plugin itself.

A next step will add this to plugins that need it and `postcss-preset-env`.


plugins/postcss-progressive-custom-properties/README.md
```diff
@@ -0,0 +1,110 @@
...
+[![NPM Version][npm-img]][npm-url]
+[![Build Status][cli-img]][cli-url]
+[<img alt="Discord" src="https://shields.io/badge/Discord-5865F2?logo=discord&logoColor=white">][discord]
+
+[PostCSS Progressive Custom Properties] is a utility plugin to correctly declare Custom Property fallbacks and enhancements.
+
+⚠️ It is not intended to be used directly by stylesheet authors.
+Meant to be included in other PostCSS plugins that provide CSS value transforms as fallbacks.
+
+[Custom Properties are not discarded like regular declarations when invalid.](https://drafts.csswg.org/css-variables/#invalid-variables)
```
Antonio-Laguna, line 12:
```suggestion
[Custom Properties are not discarded like regular declarations when invalid.](https://www.w3.org/TR/css-variables-1/#invalid-variables)
```

Let's use the link from the Candidate Recommendation

plugins/postcss-progressive-custom-properties/README.md
```diff
@@ -0,0 +1,110 @@
...
+This makes it tricky to provide fallback values for older browsers.
+
+The solution is to wrap Custom Property declarations in an `@supports` rule.
```
Antonio-Laguna, line 15: Should we make any notes about `@support` support? See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports#browser_compatibility)

Not for us since I think features we'll be using this for are ahead of `@support` and they just fail silently, but rather for any plugin author that might be interested in this?

plugins/postcss-progressive-custom-properties/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+				if (decl.type !== 'decl') {
+					return;
+				}
+
+				if (!decl.variable) {
+					return;
+				}
+
+				if (decl.value.trim() === 'initial') {
+					// https://drafts.csswg.org/css-variables/#guaranteed-invalid
```
Antonio-Laguna, line 20:
```suggestion
					// https://www.w3.org/TR/css-variables-1/#guaranteed-invalid
```

plugins/postcss-progressive-custom-properties/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+					return;
+				}
+
+				if (decl.value.trim() === '') { // empty string value
+					// https://drafts.csswg.org/css-variables/#guaranteed-invalid
```
Antonio-Laguna, line 25:
```suggestion
					// https://www.w3.org/TR/css-variables-1/#guaranteed-invalid
```

Antonio-Laguna (commented):
Thanks for this!

Just some minor things commented. Do you have any particular old issue we should bring over for this?

plugins/postcss-progressive-custom-properties/README.md
romainmenke: > Not for us since I think features we'll be using this for are ahead of @support and they just fail silently, but rather for any plugin author that might be interested in this?

exactly :)
I wouldn't specifically mention this.

plugins/postcss-progressive-custom-properties/README.md
romainmenke: good catch!

Antonio-Laguna (approved):

romainmenke:
> Just some minor things commented. Do you have any particular old issue we should bring over for this?

The ones reported have been fixed, but this plugin would fix everything, even the issues not (yet) reported.

PR #241, opened 2022-02-10
postcss-color-function : none color component

romainmenke:
add support for the `none` color component

https://drafts.csswg.org/css-color/#missing

-----

Also add some tests and a fix for percentage values.
These are not supported in all color spaces

https://drafts.csswg.org/css-color/#color-function


romainmenke:
still todo : alpha can also be `none`

PR #243, opened 2022-02-11
postcss-color-function : fix percentage handling

romainmenke:
`xyz` color spaces do not support channel components with `%` units.

Also cleaned up a few bits :
- code comments
- plugin option typings
- some tests for future features (for interop)


Antonio-Laguna (approved):
LGTM! Nice one

Issue #105, opened 2021-09-29
CSS Cascade Layers (@layer) support


paulshryock:
Would it be possible to add support for [CSS Cascade Layers](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) (`@support`)? This plugin would potentially handle ordering/scoping during compilation to provide the expected behavior of Cascade Layers in compiled CSS files if browsers indicated in `browserslist` do not yet support Cascade Layers.

I haven't given extensive thought to this idea, so if it just _wouldn't work_ the way I described, then fair enough--feel free to close this.

### Resources
- [Bram.us article about Cascade Layers](https://www.bram.us/2021/09/15/the-future-of-css-cascade-layers-css-at-layer/)
- [CSS Tricks article about Cascade Layers](https://css-tricks.com/cascade-layers/)

Antonio-Laguna:
@paulshryock this is a great feature but I'm afraid is not as straightforward as just reordering given that specificity changes drastically its meaning and layer order matters most.

As an example:

```html
<input type="text" id="input" class="my-input" />
```

With this CSS:

```css
@layer theme {
  #input {
    color: red;
  }
}

@layer overrides {
  .my-input {
    color: blue;
  }
}
```

Would make the input be colored `blue` since order matters more than specificity. Moving things around wouldn't solve this issue. I'd say we'd need JS here to apply things. It'd be surely fun to prototype but not something we can quickly achieve.

Thanks for articulating it and linking to resources though!

paulshryock:
Fair enough, that makes sense.

romainmenke:
More explainer : https://www.smashingmagazine.com/2022/01/introduction-css-cascade-layers/

--------

I think we can implement this with `:not()` selectors.

1. calculate the highest a (id's) specificity in any selector (`X`)
2. calculate the number of layers (`Y`)
3. take an obscure short id selector (`#obscure`)
4. assign each unlayered selector `X` + `Y` `:not(#obscure)` suffixes
5. for each layered selector do the same but use `X` + index in layer order

This can be further minimised by using `:not(a, b, c)` lists and by more carefully calculating the minimum number of needed selectors.

```pcss
@layer utilities {
  .blue {
    color: blue;
  }
}

/* wins */
p {
  color: red;
}
```

becomes

```pcss
@layer utilities {
  .blue {
    color: blue;
  }
}

/* wins */
p:not(#layer-1) {
  color: red;
}
```

[...]

PR #244, opened 2022-02-11
css cascade layers polyfill

oluoluoxenfree:
https://github.com/csstools/postcss-plugins/issues/105

Writing a polyfill for cascade layers using `:is()` and `:not()`.


romainmenke:
Hi @oluoluoxenfree,

Was wondering if you could give a short status update.
_But do not feel rushed by this :)_

Let us know if there is anything we can do to help.
If you no longer have bandwidth for this, that is also fine!

oluoluoxenfree:
Hi @romainmenke,

I'm still at the beginning stages; trying to wrap my head around the AST stuff... I was trying to find some test case stylesheets too, and set up some unit tests.

Will work on it this evening and hopefully have more of an update!

romainmenke:
For more complex feature I find it handy to have real browser tests and use wpt as a reference :

https://github.com/web-platform-tests/wpt/blob/fd7a9e9a139e01cd5017b57b8d3123d3a9a1a7bb/css/css-cascade/layer-basic.html

If you want I can open a PR to add the browser tests setup?
Then you can focus on the PostCSS stuff.

oluoluoxenfree:
yeah I was trying to use wpt directly in a different repo! if you could do
the tests that would be great, thanks so much for your help

romainmenke:
@oluoluoxenfree I have added some basic tests and opened a PR against your repo : 
https://github.com/oddbird/postcss-plugins/pull/1#pullrequestreview-890721574

I hope this helps :)

oluoluoxenfree:
@romainmenke thanks so much!

Is there any other tool besides the AST explorer for checking what i'm targeting? It is apparently behind on its post css version

romainmenke:
I never used any tools for it, so not sure if which one is up to date and helpful.

I always browse the docs : https://postcss.org/api/
And then place `console.log` statements to verify if I read those correctly.

This could give you a head start :

```js
postcssPlugin: 'cascade-layers',
Once(root: Container) {
	root.walkAtRules((atRule) => {
		if (atRule.name !== 'layer') {
			return;
		}

		if (atRule.nodes && atRule.nodes.length) {
			console.log(atRule.name, atRule.params);
			// parse .params as layer name
			/*
				pattern:
				@layer {
					.foo {
						color: red;
					}
				}
			*/
		} else {
			console.log(atRule.name, atRule.params);
			// parse .params as list of layer names
			/*
				pattern:
				@layer foo foo.baz;
			*/
		}
	});
},
```

oluoluoxenfree:
@romainmenke thank you!

romainmenke:
@oluoluoxenfree I just noticed that I opened that PR against the wrong branch.
I didn't spot that there were both `cascase-layers` and `feature/cascade-layers`.

Hopefully not too much hassle for you?

oluoluoxenfree:
@romainmenke completely fine; do you want me to change the branch name on github?

i wanted to sense check my approach with you as it seems different to the way other people are doing it;

- replace layer.name with :is()
- add layer.param after 
- duplicate node contents 

is it better to replace line by line with a walk function than to do it this way?

romainmenke:
Hi @oluoluoxenfree!

@oluoluoxenfree said :

> would you be able to explain more about what walk() is doing

@romainmenke said :

> walk the entire AST and build a data model.
> walk the entire AST again, remove @layer rules and adding :not(<N times "#foo">)

`walk` in this context refers to `root.walkAtRules((atRule) => { ...`

This function and callback will go over each `@foo` statement in the source CSS file.
Allowing you to make edits if they are `@layer` statements.

In my mind it would be clearest (not most efficient) to have something like this :

```js
const state = ....;

root.walkAtRules((atRule) => {
  // analyse the source CSS and store in "state"
  //
  // you want to have a structure or mechanic that enables this :
  // give it a layer name : "foo" or "foo.bar"
  // get a number back that indicates how much specificity needs to be added
});

root.walkRules((rule)) => {
 // modify unlayered styles.
 //
 // these need to get the highest possible specificity compared to layered styles
 // rule.selector = ...
});

root.walkAtRules((atRule) => {
  // modify the layered CSS based on what is stored in "state"
  //
  // move all CSS out of the "atRule" and remove the now empty "atRule"
  //   https://postcss.org/api/#container-insertbefore
  //
  // modify the selectors so that they have the correct specificity
});
```

_At the moment there is no other plugin doing something like this._

--------

@oluoluoxenfree said :

> I'm not sure what you mean by a data model here either

Very simplistic it could be something like this :

```js
const state = {
  "foo": 1,
  "bar": 2
}
```

Corresponding to this :

```css
@layer foo {
  a {
    order: 1;
  }
}

@layer bar {
  a {
    order: 2;
  }
}

a {
  order: 3;
}
```

```css
a:not(#foo) { /* 1 -> #foo */
  order: 1;
}

a:not(#foo) { /* 2 -> #foo#foo */
  order: 2;
}

a:not(#foo#foo#foo) { /* unlayered -> 3 -> #foo#foo#foo */
  order: 3;
}
```

Building this state separately should make it easier to handle nested `@layer` and other tricky bits.

I hope this makes it a bit clearer :)
Feel free to ask any questions!

sanajaved7:
Hi @romainmenke and @oluoluoxenfree - I have an approach to handle when layers have both nested layers **and** unlayered styles and I'd love to get your feedback on it. 

Here is a [draft PR](https://github.com/oddbird/postcss-plugins/pull/4) I just opened and also a [forked AST](https://astexplorer.net/#/gist/53c13d303430cc637c35ba0ccc8c3bc9/b194a73bed8d0cbae1472ebcd2012795d7141deb) explorer to help visualize my thinking. 

In a scenario like this: 
```
@layer A {
	target {
		color: red;
	}

	@layer Z {
		target {
			color: yellow;
		}
	}
} 
```

The rule inside layer A but **outside** of layer Z (i.e. `target { color: red } } `) should have higher priority since it is outside of the nested layer. Here's a tiny [codepen](https://github.com/oddbird/postcss-plugins/pull/4) to see this live in case that is helpful.  

When we do our first walkthrough and build out the data model, I wanted to ensure that our state reflects this priority correctly when there is layer nesting. 

Here is how I am thinking to handle it: 
- In the first walkthrough (`root.walkAtRules((atRule)`), look within the atRule for any nested layers **and** rules outside of those nested layers 
- When that condition is met, create a new nested layer. Append this new layer at the end as the final node. 
- Clone any outside rules and place them into the new final layer. Remove original nodes from the parent atRule.

This means that after the first walkthrough our example above would now look like: 
```
@layer A {
	@layer Z {
		target {
			color: yellow;
		}
	}
	@layer A.last {
		target {
			color: red;
		}
	}
}

```

And since `layer A.last` is after `layer Z`, it will have the higher priority. And when we build out the data model, it should resemble something like this: 

```
const state = {
  "A": 1,
  "Z": 2,
  "A.last": 3,
}
```

❓The issue I am having with this approach is that after the first walk, when I am doing the second walk to build out the data model, I don't see the new layer. The AST explorer and postcss-tape tests show the insertion is working correctly but I don't think the root is showing the new insertions. Any advice on how to tackle this? 🤔 

Also, please feel free to share any thoughts on this approach or feedback. I am happy to answer any questions, too 🙏🏼

romainmenke:
@sanajaved7 Can you try with this code? :


```js
if (hasNestedLayers && hasUnlayeredStyles) {
	const implicitLayer = atRule.clone({ params: `${atRule.params}.implicit` });
	implicitLayer.each((node) => {
		node.remove();
	});

	//create final layer
	atRule.append(implicitLayer);

	// go through the unlayered rules, clone, and delete from top level atRule
	atRule.each((node) => {
		if (node.type == 'rule') {
			implicitLayer.append(node.clone());
			node.remove();
		}
	});
}
```

sanajaved7:
@romainmenke Awesome, that worked! Thanks for such a quick reply! I'll keep an eye out for your longer reply as well. Thanks again!

oluoluoxenfree:
@romainmenke I am correct that all the unlayered styles should have the same spcififictity in the data model, right? Not sure if I'm missing something, this step seemed a bit simple: https://github.com/oddbird/postcss-plugins/pull/6/files

romainmenke:
@oluoluoxenfree I added a comment on your PR with some suggestions.
I tested these changes and it seems to work 🎉

romainmenke:
@oluoluoxenfree Can you pull in the main branch from here into your fork?
There are some bits for docs and validation that have been added since starting work on this plugin and I'd like to open a PR to add these to the plugin.

oluoluoxenfree:
will do!

oluoluoxenfree:
@romainmenke is there something like `atRule.decl` and if not, how do you select the body of the rule?

romainmenke:
> is there something like atRule.decl and if not, how do you select the body of the rule?

There are multiple ways, but it really depends on what you are trying to achieve.


`atRule.nodes` is an array of children of the `atRule`.
This includes declaration, but also comments, other at rules, ...

oluoluoxenfree:
I'm trying to do the third walk, where i have to copy and delete the body of the atrule, but can't work out what to pass to `insertbefore()`?

@romainmenke

romainmenke:
@oluoluoxenfree It also depends on the type of the nodes in the atrule :)
Another `@layer` node would need to be handled differently than a selector node.

`insertBefore` and `before` are similar and just take nodes as arguments :

- `insertBefore()` : `atRule.parent.insertBefore(atRule, /* a child node of "atRule" */)`
- `before()` : `atRule.before(/* a child node of "atRule" */)`

Which nodes and if these nodes need certain mutations really depends on the feature you are building.

oluoluoxenfree:
on this PR, is there a reason not to use `replacewith`, and if so, how do add a selector when you use `insertbefore`? 

https://github.com/oddbird/postcss-plugins/pull/8

@romainmenke

romainmenke:
Hi @oluoluoxenfree,

This was really tricky to answer without taking a deep dive into the code :)

In general there are often multiple ways to get from A to B with PostCSS and it can depend on subtle aspects or personal preference.

In the end I settled on `replaceWith`.
But I mostly got there through a bit of trial and error.

```js
// transform layered styles:
//  - give selectors the specificity they need based on layerPriority state
root.walkRules((rule) => {
	const layerForCurrentRule = layerAtRuleAncestor(rule);
	if (!layerForCurrentRule) {
		return;
	}

	// some code was here

	const fullLayerName = getLayerParams(layerForCurrentRule).join('.');
	rule.selectors = rule.selectors.map((selector) => {
		return adjustSelectorSpecificity(selector, layerOrder.get(fullLayerName) + highestASpecificity);
	});
});

// Remove all @layer rules
// Contained styles are inserted before
while (root.some((node) => node.type === 'atrule' && node.name === 'layer')) {
	root.walkAtRules('layer', (atRule) => {
		atRule.replaceWith(atRule.nodes);
	});
}
```

These should all be equivalent, but non of these also modify the selector in a one-liner :

```js
atRule.replaceWith(atRule.nodes);
```

```js
atRule.before(atRule.nodes);
atRule.remove();
```

```js
parentOfAtRule.insertBefore(atRule, atRule.nodes);
atRule.remove();
```

For this case `replaceWith` is maybe the one that is the most expressive for the intended transform?

--------------

While trying to answer your question I did work through several more steps of this plugin as I needed to get a better sense of what was needed.

Shall I open a pull request with these additions to your fork?

Wanted to ask first as I didn't want to swoop in and finish a large chunk.
Building it is the fun part.

oluoluoxenfree:
@romainmenke feel free to open a branch!

sanajaved7:
@romainmenke @oluoluoxenfree I started on handling `revert-layer` this week but after chatting with Miriam, I think it might make more sense for `revert-layer` handling to be grouped with the browser transformations for layers that need to happen (like layers in media queries). What Miriam pointed out and that I didn't realize is that as we go through the other layers looking for a selector match, there could be other rules that impact that selector that we won't know or have access to at compile time. Let me know if you have any questions but for now, I'll hold off on adding any code here for `revert-layer`.

romainmenke:
> What Miriam pointed out and that I didn't realize is that as we go through the other layers looking for a selector match, there could be other rules that impact that selector that we won't know or have access to at compile time.

@sanajaved7 Do you maybe have an example?

If possible we prefer not to have a client side component to this polyfill as this will cause a flash of incorrectly styled content and make presentation dependant on javascript.

At the moment the only part of the specification that we can not handle is re-ordering of layers in `@media` / `@supports`. I think this is an ok trade off.

I am however curious if we can not handle `revert-layer` in PostCSS.
I haven't had time yet to read up on how it is supposed to work, so I have no idea if it is feasible.

romainmenke:
Have read up a bit on `revert` and `revert-layer` and I am unsure if we can provide a good enough fallback/polyfill for this :/

My best idea at the moment : 

- assign each property/value pair to a custom property as `--<property>-<layer>: <the-value>`
- keep track of the created custom properties for each layer
- anywhere `revert-layer` is used we insert `var(--<property>-<previous-layer>)`

This has many issues :)

- shorthands won't work well when mixed with longhand
- logical properties and values won't work well when mixed with non-logical
- needs more complexity to handle multiple layers

<details><summary><strong>Basic example (click to expand)</strong></summary>
<p>


```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<meta http-equiv="X-UA-Compatible" content="IE=edge">
	<meta name="viewport" content="width=device-width, initial-scale=1.0">
	<title>Document</title>

	<style>
		@layer base, special;

		@layer special {
			.item {
				--color-special: red;
				color: red;
			}
			.feature {
				color: var(--color-base);
				color: revert-layer;
			}
		}

		@layer base {
			.item {
				--color-base: blue;
				color: blue;
			}
			.feature {
				--color-base: green;
				color: green;
			}
		}
	</style>
</head>
<body>
	<p>This example contains a list.</p>

	<ul>
		<li class="item feature">Item one</li>
		<li class="item">Item two</li>
		<li class="item">Item three</li>
	</ul>
</body>
</html>
```

</p>
</details>

------------

Is this something that can be handled better client side with Javascript?

Although I understand the usefulness of `revert-layer` in the specification, I do wonder if it is blocking for this plugin?

Does this plugin work well enough without any support for `revert-layer`?

romainmenke:
Hi @mirisuzanne!

> We might just want to throw a warning if people do it anyway.

I think this is the right way to go at the moment.
This way we help people write CSS that will work the same with or without the plugin.

We can adjust later if there is a good client-side polyfill.

--------------

I did still have some questions about `!important` handling as WPT tests seemed inclusive and I don't fully understand how `!important` behaves in layered styles.

Also opened an issue about this on WPT : https://github.com/web-platform-tests/wpt/issues/33767

Do you have any guidance on this aspect?

_if this is resolved I think we can do a last review on this plugin and finalise documentation and examples_

mirisuzanne:
> Do you have any guidance on this aspect?

@romainmenke I can answer the question about proper behavior, but not the question about what other tests to look at. I replied on the WPT thread with the proper behavior (layer ordering is reversed in important layers). I also pinged some of the Chrome devs who wrote those tests, so hopefully they can speak to the other tests that exist.

romainmenke:
> I was looking at layer importing and realized there is a comment from earlier that @import shouldn't be handled by the [plugin](https://github.com/postcss/postcss-import) but does that mean that we should use the preexisting postcss import plugin for when layers are imported? Or do we not need to additionally handle layer importing at all?

You would need to use the postcss import plugin if you want to use `@layer` with `@import`.
Maybe this also needs a warning or a mention in the README?

> are we good to merge everything in without the wpt tests or are they blocking?

I do not think we need to wait on the wpt tests.
As long as the current handling of `!important` (in the feature branch) seems good to everyone.

Would be nice to have a few final touches :
- add contributors from oddbird in `package.json` in the `contributors` field
- add a reference to oddbird in the README

sanajaved7:
@romainmenke I've got a draft [PR](https://github.com/oddbird/postcss-plugins/pull/12) where I'm adding a warning and info the README for using the `@import` at-rule  - could you take a look and let me know if I'm on the right track? I also had two questions: 
1) Do we need to check if the postcss-import plugin is being used before throwing the warning? Is that possible? 
2) When the tests run, the import warning is placed after the at-rule in the CSS - is that expected/correct? I see the others have the warning above but I'm not sure why this warning comes after. Any advice? 
3) Does the language in the [README](https://github.com/oddbird/postcss-plugins/pull/12/files#diff-a3a321acee6223c7ac506f5a4e83bc30adf735eff4a073251f189a03788a0d88R117) make sense?

romainmenke:
> 1. Do we need to check if the postcss-import plugin is being used before throwing the warning? Is that possible?

This is not needed. In practice people who use `postcss-import` wont see this notice because any `@import` will have been removed by that plugin before this plugin runs.

If any `@import` is visible to this plugin it means that the user isn't using `postcss-import`.

> 2. When the tests run, the import warning is placed after the at-rule in the CSS - is that expected/correct? I see the others have the warning above but I'm not sure why this warning comes after. Any advice?

This is expected but maybe we need to make the sorting of root nodes a bit more strict 🤔.
`@charset` and `@import` are shifted up if needed, which in practice should be harmless, but people are very creative with PostCSS.

We can improve this later, after an initial release.

> 3. Does the language in the [README](https://github.com/oddbird/postcss-plugins/pull/12/files#diff-a3a321acee6223c7ac506f5a4e83bc30adf735eff4a073251f189a03788a0d88R117) make sense?

This looks great!

sanajaved7:
@romainmenke thanks so much! I've updated that test and think the PR should be [ready](https://github.com/oddbird/postcss-plugins/pull/12) to review whenever you have a chance. Thanks in advance!

romainmenke:
@oluoluoxenfree @sanajaved7 I think the whole feature is now ready to be merged?

sanajaved7:
@romainmenke yes, I think so! Should we just flip it to ready for review? Are there any other things needed as well?

romainmenke:
@sanajaved7 There seems to be a merge conflict, most likely `package-lock.json`.
Can you pull in the latest changes from main?

That should be the last step 🎉

romainmenke:
@sanajaved7 @oluoluoxenfree and everyone at oddbird thank you for all this!
Has been awesome to work with you all on this 🎉 

We will merge this in and perform a few last housekeeping steps before releasing this.
Keeping you posted on the release status!

Antonio-Laguna:
Even if I haven't been personally involved I've been keeping an eye and all I can say is great work you all! It's been awesome to see it take shape and can't wait to get it out.

Kudos!

sanajaved7:
Thank you @romainmenke and @Antonio-Laguna, this has been super fun to work on with you and we appreciate all of your help along the way! 🚀 🎉

romainmenke:
@sanajaved7 @oluoluoxenfree This has just been released : https://www.npmjs.com/package/@csstools/postcss-cascade-layers by @Antonio-Laguna

We always wait a little bit before adding it to postcss-preset-env but plan to do so next week.

Did you have anything in mind for an announcement for this?
We usually do this on twitter with some examples and docs.
But would be nice to have your info so that we can give you credit while doing so :)

PR #246, opened 2022-02-12
postcss-progressive-custom-properties : use ast matching

romainmenke:
`@supports (--foo: oklch(0% 0 0))` does not work, so implementing AST matching a bit sooner that I had planned.

This still has the same benefits as the original intention : a single plugin that fixes value and unit fallbacks in custom properties.

But it now generates much smaller and simpler `@supports` statements and they actually work.

Currently covers what we need plus 2 extra cases as examples :
- `css-color-4` (check the diffs in tests for those plugins)
- `double-position-gradients` (this is a more complex case that needs manual handling)
- `ic` unit (unit detection)

This will grow over time as other values and units are needed in plugins.

### DIFF
--- a/plugins/postcss-progressive-custom-properties/generate/color.mjs
+++ b/plugins/postcss-progressive-custom-properties/generate/color.mjs
@@ -1,26 +1,27 @@
 import { matcherForValue } from './matcher-for-value.mjs';
 
 export const colorMatchers = [
-	{
-		'supports': 'color(display-p3 0 0 0)',
-		'property': 'color',
-		'sniff': 'color',
-		'matchers': [
-			matcherForValue('color(srgb $1 $2 $3)'),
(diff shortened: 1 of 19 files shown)


plugins/postcss-progressive-custom-properties/generate/color.mjs
```diff
@@ -0,0 +1,143 @@
+import { matcherForValue } from './matcher-for-value.mjs';
+
+export const colorMatchers = [
```
Antonio-Laguna, line 3: ❤️ the matchers!

Antonio-Laguna (approved):
Nice work!

jokull:
@romainmenke testing this out it looks like the test suite does not convert this:

```css
--firebrick-var: oklch(40% var(--point-five) 0.1324 / var(--opacity-50));
```

Not sure if that's a bug or not. It supports variables for the opacity but not the LCH values.

PR #247, opened 2022-02-13
postcss-preset-env : website - a11y

romainmenke:
This contains a lot of fixes for a11y but also breaks a few things at the moment.

- header nav is no longer fixed, this gives more screen real estate to users who view pages zoomed in.
- contrast fixes
- labels
- `<h1>`, `<h2>`, ... hierarchy

At the moment this passes a11y tests by Firefox and Lighthouse.

Side effects :
- [x] side nav has been patched after making the header nav non-fixed
- [x] large mobile and tablet no longer breaks on some views.

before :

Screenshot 2022-02-13 at 12 47 59

after :

Screenshot 2022-02-13 at 14 25 35

-----

Notes :

1. The desktop first styling is a bit confusing to me, should we clean this up? And now or later?
2. Should we remove Google fonts as a font provider?
3. Should we make fonts local assets and apply font subsetting? (most of the page size is fonts and we don't use that many chars)


Antonio-Laguna (approved):
❤️ this! thanks

PR #248, opened 2022-02-13
postcss-preset-env : website - system ui font

romainmenke:
Use `system-ui` and `monospace` as fonts.
Difference is not that big, I did adjust a few weights and sizes to compensate.

side effect : removes dependency on Google Fonts (perf & GDPR)


Antonio-Laguna:
Looks great! Thanks 🎉

Issue #61, opened 2021-12-04
`postcss-double-position-gradients` does not support `preserve: true` with CSS variables


romainmenke:
### Bug description

When assign a gradient value to a variable with `preserve: true` the actual value will be invalid.

### Source CSS

```css
:root {
	--some-var: linear-gradient(90deg, black 25% 50%, blue 50% 75%);
}
```

### Expected CSS

```css
:root {
	--some-var: linear-gradient(90deg, black 25%, black 50%, blue 50%, blue 75%);
}

@supports (linear-gradient(90deg, black 25% 50%, blue 50% 75%)) {
	:root {
		--some-var: linear-gradient(90deg, black 25% 50%, blue 50% 75%);
	}
}
```

### Actual CSS

```css
:root {
	--some-var: linear-gradient(90deg, black 25%, black 50%, blue 50%, blue 75%);
	--some-var: linear-gradient(90deg, black 25% 50%, blue 50% 75%);
}
```

### Does it happen with `npx <plugin-name> minimal-example.css`?

Yes

### Extra config

`preserve` set to `true`

### What plugin are you experiencing this issue on?

PostCSS Double Position Gradients

### Plugin version

3.0.1

### What OS are you experiencing this on?

macOS

### Node Version

16

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #256, opened 2022-02-15
postcss-progressive-custom-properties : fix double position gradients

romainmenke:
- various fixes for double position gradients and custom props.
- copy/paste fix voor `postcss-lab-function`
- a bit more docs for the progressive custom props

fixes : https://github.com/csstools/postcss-plugins/issues/61

### DIFF
--- a/plugins/postcss-double-position-gradients/src/has-supports-at-rule-ancestor.js
+++ b/plugins/postcss-double-position-gradients/src/has-supports-at-rule-ancestor.js
@@ -7,27 +7,14 @@ export function hasSupportsAtRuleAncestor(node) {
 		}
 
 		if (parent.name === 'supports') {
-			if (parent.params.indexOf('conic-gradient(') !== -1) {
-				return true;
-			}
-
-			if (parent.params.indexOf('linear-gradient(') !== -1) {
-				return true;
-			}
-
-			if (parent.params.indexOf('radial-gradient(') !== -1) {
-				return true;
-			}
-
-			if (parent.params.indexOf('repeating-conic-gradient(') !== -1) {
-				return true;
-			}
-
-			if (parent.params.indexOf('repeating-linear-gradient(') !== -1) {
-				return true;
-			}
-
-			if (parent.params.indexOf('repeating-radial-gradient(') !== -1) {
+			if (
+				parent.params.includes('conic-gradient(') ||
+				parent.params.includes('linear-gradient(') ||
+				parent.params.includes('radial-gradient(') ||
+				parent.params.includes('repeating-conic-gradient(') ||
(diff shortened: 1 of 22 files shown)


plugins/postcss-double-position-gradients/src/has-supports-at-rule-ancestor.js
```diff
@@ -0,0 +1,39 @@
...
+
+			if (parent.params.indexOf('repeating-conic-gradient(') !== -1) {
+				return true;
+			}
+
+			if (parent.params.indexOf('repeating-linear-gradient(') !== -1) {
+				return true;
+			}
+
+			if (parent.params.indexOf('repeating-radial-gradient(') !== -1) {
```
Antonio-Laguna, line 30: A bit [nit], should we use `includes` for clarity?
Furthermore, should we abstract this a bit given that this is what is being tested on `gradientRegExp` and `gradientPartsRegExp`? Fine if not!

plugins/postcss-progressive-custom-properties/README.md
```diff
@@ -94,6 +94,14 @@ postcss([
 ]).process(YOUR_CSS /*, processOptions */);
 ```
 
+## @supports
```
Antonio-Laguna, line 97: ❤️  this notice

plugins/postcss-double-position-gradients/src/has-supports-at-rule-ancestor.js
romainmenke: I avoided the `regexp` on purpose as I completely missed that this plugin also does more than `linear-gradient` and `conic-gradient`.

There were no tests for the other 4 functions, no mention in the README.md and the regexp does have these, but it's not immediately obvious.

plugins/postcss-progressive-custom-properties/src/custom/double-position-gradients.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	// custom matchers :
+	if (node.type === 'function' && (
+		node.value === 'conic-gradient' ||
+		node.value === 'linear-gradient' ||
+		node.value === 'radial-gradient' ||
+		node.value === 'repeating-conic-gradient' ||
+		node.value === 'repeating-linear-gradient' ||
+		node.value === 'repeating-radial-gradient'
+	)) {
```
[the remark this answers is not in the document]
romainmenke: started an issue on this as it indeed needs more thought and a more careful design : https://github.com/csstools/postcss-plugins/issues/257

Maybe the result of this will be that we keep the existing things as-is but want to open it up and see if we come up with something better.

PR #259, opened 2022-02-15
Introducing `color()` `oklab/oklch`, `ic` to PostCSS Preset Env

Antonio-Laguna:
* Integrates both in PostCSS Preset Env and CLI the following:
  * `@csstools/color-function`
  * `@csstools/oklab-function`
  * `@csstools/ic-unit`
* Also fixes README and add missing options to CLI plugins as well as some logger texts


plugin-packs/postcss-preset-env/test/insert.after.match-result.expect.css
```diff
@@ -2,5 +2,5 @@
 	background-color: rgba(255, 255, 255, 0);
 	background-color: color(display-p3 1.00000 1.00000 1.00000 / 0%);
 	after: changed-this-declaration;
-	color: color(display-p3 0.00000 0.00000 0.00000 / 0%);
+	after: changed-this-declaration;
```
Antonio-Laguna, line 5: Something odd going on here, it gets retransformed somehow. These three tests on this commit seem wrong

plugin-packs/postcss-preset-env/.tape.mjs
```diff
@@ -349,6 +329,38 @@ postcssTape(plugin)({
...
+			],
+			insertAfter: {
+				'lab-function': [
+					orderDetectionPlugin('after', (decl) => {
+						return decl.prop === 'color' && decl.value !== 'changed-this-declaration';
+					})
+				]
+			}
+		}
+	},
```
romainmenke, line 363: This two test cases had copy/paste issues.
Was intended to test `insertBefore` and `insertAfter` when the attached feature is skipped.

- messages was not acurate
- both had the same config

I also placed them lower in the list.

plugin-packs/postcss-preset-env/.tape.mjs
```diff
@@ -271,44 +280,13 @@ postcssTape(plugin)({
...
-		}
-	},
 	'insert:after:match-source': {
 		message: 'supports { insertAfter } usage when looking for source',
 		options: {
 			stage: 0,
 			features: {
-				'lab-function': true
+				'lab-function': true,
+				'color-function': false,
```
romainmenke, line 289: `color-function` transforms `color(display-p3 ...)`.
These tests look for the results of `lab-function` transforms, so displaying `color-function` for now in these cases.

Maybe these need to be rewritten to use a more basic plugin?

Antonio-Laguna: I thought it was related to this and also thought about using something else such as `rebeccapurple` or something that's less prone to change.

Antonio-Laguna: Thanks for taking care of this!

romainmenke (approved):
LGTM!

Thank you for carefully adding all the new things to both `preset-env` and the `cli`! 🎉

Issue #111, opened 2020-06-25
feature request: unset


MathGeniusJodie:
https://developer.mozilla.org/en-US/docs/Web/CSS/unset

romainmenke:
@MathGeniusJodie This was just released as https://www.npmjs.com/package/@csstools/postcss-unset-value

It will be added to `postcss-preset-env` in the next minor version.
Thank you for requesting this!

PR #264, opened 2022-02-17
postcss-unset-value

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/111

Uses [@webref/css](https://www.npmjs.com/package/@webref/css) data to determine if it should become `inherit` or `initial`.

reading :
- [inherit](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit) has pretty good support
- [initial](https://developer.mozilla.org/en-US/docs/Web/CSS/initial) has a [plugin](https://github.com/maximkoretskiy/postcss-initial) (mainly IE)

done :
- [x] more tests
- [x] figure out why some props are both inherited and non-inherited in webref
- [x] write all the README and INSTALL
- [x] labeler
- [x] issue templates
- [x] check all the links more than twice
- [x] cssdb


romainmenke:
Note to self : changed the name from `postcss-unset` to `postcss-unset-value` to match caniuse.

Do an extra pass to make sure all names and links were updated correctly.

Antonio-Laguna (approved):
This looks great! Thanks for tackling this!

PR #265, opened 2022-02-19
postcss-preset-env : website - design tokens

romainmenke:
A first pass that uses CSS custom properties as design tokens.
This makes it easier to have a consistent result.

This then also fixes some issues with inconsistency :)
A few tweaks to the landing made it usable up too IE9.


Antonio-Laguna (approved):
Thanks for this! This looks much cleaner!

PR #266, opened 2022-02-20
css-color-4 : percentage & number unit color components

romainmenke:
see : 
- https://github.com/w3c/csswg-drafts/commit/af5097250a1aac55f1a889803c4fb19aafebe1e1
- https://drafts.csswg.org/css-color/#funcdef-lch

todo :
- [x] `oklab` and `oklch`
- [x] `lab` and `lch`
- [x] `rgb` and `hsl` (already supported in functional notation)
- [x] `color(...)`

Currently no plan to add a dedicated plugin to transform to something that Safari natively supports. It will be rare that people build specifically for Safari.

Would be better to treat the current Safari implementation as a partial implementation. (needs cssdb, mdn, ... updates)


romainmenke:
@Antonio-Laguna The spec hasn't had any changes relevant to this update in a month and I haven't spotted any issues related to this.

In my opinion we can move forward with this.

Antonio-Laguna (approved):
Nice work!

PR #267, opened 2022-02-20
format package.json and generate INSTALL.md

romainmenke:
Currently `package.json` files can have widely different ordering which creates a small bit of friction when looking for info between multiple plugins.

Having consistent ordering is low effort and makes the whole a bit cleaner.

- `npm run lint` fixes ordering
- in CI process exits with `1` if ordering is incorrect

-------

Started adding bits of metadata to `package.json`.

_It is much easier to parse, test and use values stored from `package.json` to build markdown docs than the other way around._

```json
  "csstools": {
    "exportName": "postcssColorFunction",
    "humanReadableName": "PostCSS Color Function"
  },
```

These make it possible to generate `INSTALL.md`.

(`README.md` is more complex and is still a fully manual process.)


Antonio-Laguna (approved):
This looks great! Thanks for doing this! Love that things are normalized!

PR #272, opened 2022-02-21
Removing Astro Workaround

Antonio-Laguna:
Astro's team acknowledged the failure and it seemed to be a cross-platform issue which was solved on Astro 0.23.0


romainmenke (approved):
LGTM!

PR #276, opened 2022-02-23
css-has-pseudo : base36 encoding

romainmenke:
Uses base36 encoding with `-` as a separator. This results in a smaller file size than base36 with fixed length chunks.

There is no limit to what can be encoded this way aside from browser support for the characters being encoded.

`csstools-has-` prefix is added because we need something we can detect.

### DIFF
--- a/experimental/css-has-pseudo/src/browser.js
+++ b/experimental/css-has-pseudo/src/browser.js
@@ -2,6 +2,7 @@
 
 import '@mrhenry/core-web/modules/~element-qsa-has.js';
 import extractEncodedSelectors from './encode/extract.mjs';
+import encodeCSS from './encode/encode.mjs';
 
 export default function cssHasPseudo(document, options) {
 	// OPTIONS
@@ -236,10 +237,14 @@ export default function cssHasPseudo(document, options) {
 
 						for (let i = 0; i < hasSelectors.length; i++) {
 							const hasSelector = hasSelectors[i];
+
+							console.log(hasSelector);
+							console.log(encodeCSS(hasSelector));
+
 							observedItems.push({
 								rule: rule,
 								selector: hasSelector,
-								attributeName: encodeURIComponent(hasSelector).replace(/%3A/g, ':').replace(/%5B/g, '[').replace(/%5D/g, ']').replace(/%2C/g, ','), // TODO : needs unit tests.
+								attributeName: encodeCSS(hasSelector),
 								nodes: [],
 							});
 						}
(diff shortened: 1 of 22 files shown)


experimental/css-has-pseudo/src/encode/test.mjs
```diff
@@ -130,201 +130,26 @@ testExtract(
...
+	'[' + encodeCSS('.x:has(> .🧑🏾‍🎤)') +']',
 	['.x:has(> .🧑🏾‍🎤)'],
 );
 
 testExtract(
-	'[\\.x\\:has\\(\\%3E\\%20\\.a\\:has\\(\\%20\\%2B\\%20\\.b\\)\\)][\\.x\\:has\\(\\%3E\\%20\\.a\\:has\\(\\%20\\%2B\\%20\\.b\\)\\)][\\.x\\:has\\(\\%3E\\%20\\.a\\:has\\(\\%20\\%2B\\%20\\.b\\)\\)]',
+	'[' + encodeCSS('.x:has(> .a:has( + .b))') + ']',
 	['.x:has(> .a:has( + .b))'],
 );
-
```
romainmenke, line 156: The previous encoder/decoder had to cover escaped `[`/`]`.
This required a lot of tests.

The new encoder/decoder doesn't have this complexity so a lot of tests became unneeded.

Antonio-Laguna (commented):
This looks great! Thanks for quickly fixing this.

I bet you don't want to leave `console.log` within the `browser` file?

experimental/css-has-pseudo/src/browser.js
```diff
@@ -236,10 +237,14 @@ export default function cssHasPseudo(document, options) {
 
 						for (let i = 0; i < hasSelectors.length; i++) {
 							const hasSelector = hasSelectors[i];
+
+							console.log(hasSelector);
+							console.log(encodeCSS(hasSelector));
+
```
romainmenke, line 243:
```suggestion
```

experimental/css-has-pseudo/src/index.ts
```diff
@@ -53,7 +53,7 @@ const creator: PluginCreator<{ preserve?: boolean, specificityMatchingName?: str
 						// We can't leave `:has` untouched as that might cause broken selector lists.
 						// Replacing with the specificity matching name as this should never match anything without `:not()`.
 						node.replaceWith(parser.className({
-							value: '.' + options.specificityMatchingName,
+							value: options.specificityMatchingName,
```
romainmenke, line 56: Noticed this in the logs, `parser.className` already adds leading `.`

Antonio-Laguna (approved):

PR #278, opened 2022-02-25
fix selector order in pseudo :is

romainmenke:
Pseudo elements ordering appears important : https://github.com/w3c/csswg-drafts/issues/7085#issue-1149016131

With re-ordering in compound selectors we broke this in `:is()`

-----

Before this change : 

```css
:is(a, button):is(:hover::before, :focus) {
	order: 25;
}

/* becomes */

button::before:hover {
	order: 26;
}
```

After this change :

```css
:is(a, button):is(:hover::before, :focus) {
	order: 25;
}

/* becomes */

button:hover::before {
	order: 26;
}
```

Also added a test in nesting to prevent introducing the same bug in the future and clarified a few comments.


Antonio-Laguna (approved):
Nice one! Thanks for fixing this!

Issue #23, opened 2021-11-21
Generate README files from templates


romainmenke:
Docs are often out of date :
- incorrect imports
- incorrect urls
- incorrect examples

They also always follow the same patterns.

Could we generate these?

----------

Examples could be pulled from `test` dir :

- example is known to be correct, otherwise CI would fail

```css
/* test/example.css */
.foo {
  color: red;
}
```

```css
/* test/example.expect.css */
.foo {
  color: blue;
}
```

----------

Package name, urls, ... 
This can be pulled from `package.json` and the directory of the package.

----------

`package.json` can include extra fields with links to spec, caniuse, mdn, ...
This would also make it possible for others to extract this kind of meta data.

for example :

```json
{
  "plugin_meta": {
    "mdn_url": "...",
    "spec_url": "...",
  }
}
```

-----

CLI usage can be generated by running the CLI without any args.
This will cause it to output the help text on `stderr`.

-------

What would then still need to be provided manually :
- why the plugin exists and what it does
- what the plugin options are

> [PostCSS Environment Variables] lets you use `env()` variables in CSS, following the [CSS Environment Variables] specification.

romainmenke:
Note :

Separate edge cases and warnings into a separate doc.

romainmenke:
Note :

things like versions in README docs (see browser polyfill cdn urls) should also be filled in based on `package.json`

romainmenke:
Progress :

`INSTALL.md` is now generated for all packages in `plugins/*`

PR #280, opened 2022-02-26
readme templating

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/23

This is different from `INSTALL.md`.

- `INSTALL.md` should be the same for each plugin and only specific links and names need inserting.
- `README.md` has plugin specific docs for options, edge cases, ...

This remains opt-in, just like `INSTALL.md`. So any plugin can still have their own way of doing docs.

-------

Each plugin that wants to use this needs :
- a change in `package.json` `scripts.docs`
- a file in `docs/README.md` with the plugin specific bits.
- examples as `test/examples/*.css`

The file in `docs/README.md` is the template and a plugin author can do anything here.
At the top of the template is a list of available content tags.

e.g. : `<header>` will insert a readme header with the correct values and links.

-------

This approach is not intended to be less work than doing everything manually.
It aims to increase accuracy and correctness and reduce the cost of batch updates.

When used correctly each plugin will have correct links and examples.

- links need manual checking but it's much harder to miss one or have a typo
- examples are tested with `postcss-tape` and are assured to be accurate.

The extra work lies mainly in needing to understand the format and providing the examples.
But it allows you to focus on the important bits and ignore the chores.

---------

Side note :

`*.md` now has tabs in editor config. tabs are better for a11y and should always be used when allowed by the file format.

_________

_drafted because I first want to discuss_

--------

What a template looks like :

````
<header>

[<humanReadableName>] lets you use the `color` function in
CSS, following the [CSS Color] specification.


```pcss
<example.css>

/* becomes */

<example.expect.css>
```

<usage>

<env-support>

## Options

### preserve

The `preserve` option determines whether the original notation
is preserved. By default, it is not preserved.

```js
<exportName>({ preserve: true })
```

```pcss
<example.css>

/* becomes */

<example.preserve-true.expect.css>
```

### enableProgressiveCustomProperties

The `enableProgressiveCustomProperties` option determines whether the original notation
is wrapped with `@supports` when used in Custom Properties. By default, it is enabled.

⚠️ We only recommend disabling this when you set `preserve` to `false` or if you bring your own fix for Custom Properties. See what the plugin does in its [README](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-progressive-custom-properties#readme).

```js
<exportName>({ enableProgressiveCustomProperties: false })
```

```pcss
<example.css>

/* becomes */

<example.preserve-true.progressive-false.expect.css>
```

_Custom properties do not fallback to the previous declaration_

## Supported Color Spaces

```css
.color-spaces {
	color: color(a98-rgb 0.803 0.484 0.944);
	color: color(display-p3 0.8434 0.509 0.934);
	color: color(prophoto-rgb 0.759 0.493 0.898);
	color: color(rec2020 0.772 0.491 0.920);
	color: color(srgb 0.897 0.488 0.959);
	color: color(srgb-linear 0.783 0.203 0.910);
	color: color(xyz 0.560 0.377 0.904);
	color: color(xyz-d50 0.550 0.375 0.680);
	color: color(xyz-d65 0.560 0.377 0.904);
}
```

## Out of gamut colors

Depending on the browser implementation out of gamut colors may be clipped, resulting in a different color.<br>
Fallback values generated by [<humanReadableName>] are always mapped to a close alternative in sRGB.

When setting `preserve` to `true` the original values will be used by some browsers and these may be clipped.<br>
Certain browsers will have an incorrect color if this occurs.

If the plugin detects out of gamut colors it will emit a warning :

> "color(srgb-linear -0.01656 0.23079 0.25298)" is out of gamut for "srgb-linear". When "preserve: true" is set this will lead to unexpected results in some browsers.

To resolve this warning pick a larger color space when declaring the original value.

## Copyright : color conversions

This software or document includes material copied from or derived from https://github.com/w3c/csswg-drafts/tree/main/css-color-4. Copyright © 2022 W3C® (MIT, ERCIM, Keio, Beihang).

<link-list>
````


romainmenke:
@Antonio-Laguna With the change to tabs in `*md` this PR became very noisy.
So maybe best to first merge this batch.

I also updated the base plugin and tested `npm run new-plugin`.

In the root `package.json` there is a new helper command `prepare-repo-for-releases`.

This runs : 

- `npm run clean`
- `npm ci`
- `npm run docs`
- `npm run lint`
- `npm run build`
- `npm run test`
- `bash ./.github/bin/fail-on-changes.sh`

This is intended to be run once before starting one or more plugin releases.
It will emit a warning and `exit 1` if the git state isn't clean.

This can happen if `docs` became outdated for example.

If it doesn't emit a warning and has `exit 0` you are sure of these things :
- `node_modules` matches `package-lock.json`
- there are no zombie untracked files in any `/dist` folders
- all plugins were freshly build and tested
- all docs were freshly generated and matched what was already on `main` brach

Antonio-Laguna:
@romainmenke thanks a lot for this! It was a bit noisy indeed for reviewing but ignoring those files on my end and reviewed the rest. All looks great! Thanks for doing this!

PR #284, opened 2022-03-02
Patching PostCSS Clamp

### LINKED ISSUE #274: Infinite loop with postcss-clamp
@Antonio-Laguna @filipjakov 
I'm using this line of code for instance which now breaks with the newest postcss-preset-env and postcss-clamp (using PostCSS on a SvelteKit install - took me quite a while to figure where the issue came from)

```scss
padding: clamp(48px, 6.5vw, 104px) 0 calc(var(--footer-height) + clamp(48px, 6.5vw, 104px));
```

This used to compile properly but now it just keeps compiling until node crashes being out of memory 😅 
Is there a way to fix this or disable postcss-clamp as the support is pretty good now? I was using stage 2, maybe switching to stage 3?

_Originally posted by @flayks in https://github.com/csstools/postcss-plugins/issues/260#issuecomment-1047633511_

Antonio-Laguna:
This is a patch to fix #274 which copies the plugin submitted to the PR here https://github.com/polemius/postcss-clamp/pull/19

While we want to keep using Polemius' work, we also want to get this fixed ASAP so once it's integrated we can just get back to using the dependency

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11588 +1,11590 @@
 {
-  "name": "@csstools/postcss-plugins",
-  "version": "1.0.0",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
-      "name": "@csstools/postcss-plugins",
-      "version": "1.0.0",
-      "license": "CC0-1.0",
-      "workspaces": [
-        "packages/*",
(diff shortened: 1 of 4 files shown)


romainmenke (approved):
LGTM!

Issue #233, opened 2022-02-06
Add `in color-space` syntax to gradients


ai:
### What would you want to propose?

Safari TP already [got](https://twitter.com/argyleink/status/1490376117064065025) color space support for gradients:

```css
background: linear-gradient(in lch, blue, red);
```

### Suggested solution

This feature (but without `in X` syntax) was already implemented in [postcss-easing-gradients](https://github.com/larsenwork/postcss-easing-gradients) by @larsenwork.

Maybe we can start from that code.

### Additional context

This feature could be part of amazing colors refactoring going by @romainmenke.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
https://drafts.csswg.org/css-images-4/#linear-gradients

romainmenke:
At @ai The pull request linked above will close this issue as fixed.
But that plugin on its own will be useless.

Hopefully I will have time for `color-mix` soon, then we can all have fancy new gradients.

ai:
Looks amazing. I will promote it later, after some sort end of war. I can’t force myself to post any non-anti-war content 😓.

Issue #85, opened 2021-12-22
Support for `at` keyword in css gradient functions


romainmenke:
### Bug description

Currently gradient functions with extra keywords lead to incorrect output as they are mistaken for color values.

-----

see : https://github.com/csstools/postcss-preset-env/issues/120

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient()

https://drafts.csswg.org/css-images/#radial-gradients

-----

Propose rewriting gradients plugin from an order based find/replace to something that first builds a data model and then stringifies this to a syntax that most browsers understand.

### Source CSS

```css
.at-keyword {
	background-image: radial-gradient(farthest-corner at 40px 40px, #f35 0%, #43e 100%);
}
```

### Expected CSS

_need to check if there is a valid alternative_

### Actual CSS

```css
.at-keyword {
	background-image: radial-gradient(farthest-corner at 40px,at 40px, #f35 0%, #43e 100%);
}
```

### Does it happen with `npx <plugin-name> minimal-example.css`?

Yes

### What plugin are you experiencing this issue on?

PostCSS Double Position Gradients

### Plugin version

7.1.0

### What OS are you experiencing this on?

macOS

### Node Version

16

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

Issue #177, opened 2022-01-18
css-color-5


depoulo:
### What would you want to propose?

While I know that [css-color-4 is just being implemented](https://github.com/csstools/postcss-plugins/pull/70#issuecomment-1015407740), I already wish postcss-preset-env supported https://drafts.csswg.org/css-color-5/.

- [x] color-mix
- [ ] ~~color-contrast~~ (css-color-6 now)
- [x] relative colors

### Suggested solution

🤷🏽‍♂️ 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
🎉 shipped :
- color-mix-function: https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-color-mix-function
- gradients-interpolation-method: https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-gradients-interpolation-method

t0byman:
Do you have an ETA when the `relative colors` polyfill (with full support for dynamic CSS Custom Properties) ships?
Would make me very happy because then I can change:
```
:root {
    --c-dark: #333;

    @supports (color: oklch(0 0 0)) {
        --c-dark: oklch(0.3211 0 0);
    }
}

@supports not (background-color: oklch(from #000 l c h / 0%)) {
    background-color: color-mod(var(--c-dark) a(80%));
}

@supports (background-color: oklch(from #000 l c h / 0%)) {
    background-color: oklch(from var(--c-dark) l c h / 80%);
}
```
to just:
```
:root {
    --c-dark: oklch(0.3211 0 0);
}

background-color: oklch(from var(--c-title) l c h / 80%);
```

romainmenke:
@t0byman We will never ship support for variables because it is impossible to statically determine there value.

We are still waiting on the specification to finalize.
Currently we have no idea what we need to create.

romainmenke:
🎉 shipped :

- https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-relative-color-syntax#readme

We now cover `css-color-5` as much as possible.
We will open a new issue for `css-color-6` and `color-contrast` when the specification progresses enough.

Thank you everyone for your interest in these features 🚀

[...]

PR #287, opened 2022-03-04
postcss-gradients-interpolation-method

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/85
fixes : https://github.com/csstools/postcss-plugins/issues/233

related to : https://github.com/csstools/postcss-plugins/issues/177

-----

It uses `calc` to avoid having to calculate the positions of color stops.
This makes it possible to support `calc` and `var` values in positions.

just works : `linear-gradient(in oklch 90deg, black 20% var(--perc-50),blue 70% 75%);`

_Technically allowed to have `var` in colors but the future `color-mix` plugin won't be able to handle those._

Color stops with only a color and Interpolation hints are not supported.

-----

Note : this plugin in isolation is useless.
A separate plugin is required to downgrade `color-mix()`.

This approach makes it possible to have all `color-mix` logic in a single plugin.

I think we can merge this as is just to get the ball rolling.


plugins/postcss-double-position-gradients/src/index.ts
```diff
@@ -80,11 +106,18 @@ const basePlugin = (opts) => {
...
 
-						insertBefore(func.nodes, node, [comma, color]);
+						func.nodes.splice(
+							// 1 before the current node
+							func.nodes.indexOf(node) - 1,
+							// remove none
+							0,
+							// insert these :
+							comma, color,
+						);
```
Antonio-Laguna, line 116: ❤️

Antonio-Laguna (approved):
Great work! This looks awesome and paves the way for color-mix

PR #299, opened 2022-03-07
fix gamut mapping and powerless color components

romainmenke:
for powerless color components see : https://drafts.csswg.org/css-color-4/#binsearch

> let origin_OKLCH be origin converted from origin color space to the OKLCH color space
if the Lightness of origin_OKLCH is greater than or equal to 100%, return { 1 1 1 origin.alpha } in destination
if the Lightness of origin_OKLCH is less than than or equal to 0%, return { 0 0 0 origin.alpha } in destination


plugins/postcss-color-function/src/convert-prophoto-rgb-to-srgb.ts
```diff
@@ -15,6 +15,13 @@ export function prophotoRgbToSRgb(prophoto: color): color {
 	let oklch = conversion.slice() as color;
 	oklch = XYZ_to_OKLab(oklch);
 	oklch = OKLab_to_OKLCH(oklch);
+	if (oklch[0] < 0.000001) {
+		oklch = [0, 0, 0] as color;
+	}
+
+	if (oklch[0] > 0.999999) {
+		oklch = [1, 0, 0] as color;
+	}
```
romainmenke, line 24: This is closer to 0 or 1 than to the smallest/largest possible value that can be expressed in a 16bit color system.

Antonio-Laguna:
@romainmenke thanks! Looks great and learning about powerless was awesome!

Issue #151, opened 2022-01-12
Remove side effects from all plugins


romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/140

romainmenke:
This is done. Plugin releases and rollout in `postcss-preset-env` is tracked here : https://github.com/csstools/postcss-plugins/issues/464

PR #304, opened 2022-03-14
design tokens

romainmenke:
see : https://github.com/csstools/postcss-plugins/discussions/192
see : https://github.com/csstools/postcss-plugins/issues/151

-------

_Notice :_

_A lot has changed since the initial pull request was opened._
_I have hidden all comments as they likely no longer apply._

-------

[The README in this pull request illustrates the current design direction](https://github.com/csstools/postcss-plugins/blob/design-tokens--affectionate-sponge-94364bcfba/plugins/postcss-design-tokens/README.md)

```json
{
	"color": {
		"background": {
			"primary": { "value": "#fff" }
		}
	}
}
```

```pcss
@design-tokens url('./tokens.json') format('style-dictionary3');

.foo {
	color: design-token('color.background.primary');
}

/* becomes */

.foo {
	color: #fff;
}
```

Current approach aims to support multiple token file formats side by side.
Will also heavily lean on IDE support for ergonomics.


Antonio-Laguna:
@romainmenke this looks great!

I have a few questions and some thoughts.

I assume we'd have sensible defaults to not have to specify that something isn't `deprecated` or that we `allowedProperties` is empty, right?

My only concern here comes with doing things that standards will eventually cover. That's mostly custom media and custom selectors that I'm worried about since these would be clashing. They wouldn't clash as in code issues since this is all new and clean (nice work!) but in terms of ownership and confusion. 

What were your thoughts on this?

romainmenke:
> I assume we'd have sensible defaults to not have to specify that something isn't deprecated or that we allowedProperties is empty, right?

- `deprecated` defaults to `false` when omitted.
- `allowedProperties` and `blockedProperties` default to empty lists in which case they have no effect.

------

> My only concern here comes with doing things that standards will eventually cover. That's mostly custom media and custom selectors that I'm worried about since these would be clashing. They wouldn't clash as in code issues since this is all new and clean (nice work!) but in terms of ownership and confusion.

This is a valid concern!

Having no mechanic in CSS to define values for design tokens helps I think.
It is much more clear that this is build tool plugin and not a runtime feature.

I don't think we can avoid certain issues completely.

1. Any user of this plugin would be locked in. They would not be able to switch to another build tool without refactors.
2. People will abuse this as it is a more powerful find/replace string.

Maybe we need a few more options to help creators of design token packs communicate?
For example this warning :

```
[postcss-design-tokens]: design-token: "color" is not configured.
```

Assuming that a stylesheet author is working in a big company with multiple teams they might not know where/how the tokens are defined and how to resolve this.

Same for this warning :

```
[postcss-design-tokens]: design-token: "color" is deprecated.
```

Maybe better like this :

```
[postcss-design-tokens]: design-token: "color" is not configured. (fancy-design-system https://fancy-design-system.example.com/design-tokens)
[postcss-design-tokens]: design-token: "color" is deprecated. Use "<new token name>" instead. (<design token pack name> <design token pack docs url>)
[postcss-design-tokens]: design-token: "color" is deprecated and will be removed without replacement. (<design token pack name> <design token pack docs url>)
```

romainmenke:
@equinusocio If you have any thoughts or feedback on this proposal?

romainmenke:
> @romainmenke thoughts on adding some option to the props so they can be exported to :root?

This is possible with the current design :

```css
:root {
  --a: design-token(a);
  --b: design-token(b);
  ...
}
```

I am hesitant to make this something automatic.

At the moment anyone can create a pack of design tokens following this API and publish these on npm. Adding new tokens will not have unexpected side effects.

But if we make the plugin add custom properties to `:root`, it can also break things.
If you already have `--z` in your CSS it can have unexpected effects if your preferred design token pack suddenly injects a different value for `--z`;

This would also blur the line between `design-token()` and `var()`.

romainmenke:
see : https://design-tokens.github.io/community-group/format/

equinusocio:
My opinion is that the plugin should not force any syntax for design tokens, which may be really different in each case. For example, I would never use this syntax:

```
values: [
	{
		name: 'my-color',
		value: '#f00',
		deprecated: false,
		allowedProperties: [], // only allowed on certain properties
		blockedProperties: [], // not allowed on certain properties
	}
]
```

Design tokens are platform-agnostic used to generate platform-specific format, so specifying CSS-related stuff, or using the postcss-preset-env JSON structure, is basically wrong and very limiting. Also, I think design tokens are never manually generated. At Wonderflow we use [style dictionary](https://amzn.github.io/style-dictionary/#/) to generate them, and the format output is totally different.

That's said, we dropped the `importFrom` option some week ago, in favor of [this](https://www.npmjs.com/package/postcss-replace) which just replaces `token(--[TOKEN])` with our design tokens. 

## Conclusion

I don't think preset-env should handle design tokens in a strict and lock-in way like the proposed one. It should just accept a dataset (json, css?, exported js object) and provide a syntax to use that token as a value. Everything else would be lock-in.

Should preset-env use `importFrom` just to import `var()` dropping custom `env()`, and providing an extra option like `tokens: {}`?

romainmenke:
>My opinion is that the plugin should not force any syntax for design tokens, which may be really different in each case. For example, I would never use this syntax:

> Design tokens are platform-agnostic used to generate platform-specific format, so specifying CSS-related stuff, or using the postcss-preset-env JSON structure, is basically wrong and very limiting

That is what the initial discussion was about.
We have to enforce a schema.
We can not reasonably support everyones preferred way to express a token schema.

Since then we became aware that there is an attempt to create a spec for a schema and token types to share design tokens between different tools. Choosing to adopt that format is a no brainer in my opinion.

https://design-tokens.github.io/community-group/format/

> Also, I think design tokens are never manually generated. At Wonderflow we use [style dictionary](https://amzn.github.io/style-dictionary/#/) to generate them, and the format output is totally different.

Their schema looks compatible with the current proposal and style dictionary is even listed as one of the translation tools in the proposal.

https://design-tokens.github.io/community-group/format/#translation-tool

> I don't think preset-env should handle design tokens in a strict and lock-in way like the proposed one. It should just accept a dataset (json, css?, exported js object) and provide a syntax to use that token as a value. Everything else would be lock-in.

The proposed PostCSS plugin here will never be included in `postcss-preset-env`.
`postcss-preset-env` is a tool to downgrade and polyfill CSS. Inserting design tokens is outside the scope of `postcss-preset-env`.

The plugin would only accept one format/schema, preferably the one from the design tokens specification. At the moment the specification only covers values, so the plugin will start with only this.

> Should preset-env use importFrom just to import var() dropping custom env(), and providing an extra option like tokens: {}?

`var()` and `env()` are ill suited because that would mean overloading standard CSS.

I am currently favouring this for the CSS side :

```css
.foo {
	color: design-token("color.background.primary");
}
```

plugins/postcss-design-tokens/README.md
```diff
@@ -0,0 +1,259 @@
...
+	"size": {
+		"spacing": {
+			"small": { "value": "16px" }
+		}
+	}
+}
+```
+
+```pcss
+@design-tokens url('./tokens.json') format('style-dictionary3');
```
Antonio-Laguna, line 23: I wonder if we should default to `style-dictionary3` to avoid the need to add this for convenience? Specially given there's no extra systems right now

plugins/postcss-design-tokens/README.md
```diff
@@ -0,0 +1,259 @@
...
+
+.foo {
+	color: design-token('color.background.primary');
+	padding-left: design-token('size.spacing.small' to px);
```
Antonio-Laguna, line 27: Is it possible to use `size.spacing.small` without `to`? If so, should we illustrate?

plugins/postcss-design-tokens/src/data-formats/parse-import.ts
```diff
@@ -0,0 +1,67 @@
...
+import path from 'path';
+import { promises as fsp } from 'fs';
+
+function parseImport(statement: string): { filePath: string, format: string, conditions: Array<string> } {
+	const importAST = valueParser(statement);
+
+	const result = {
+		filePath: '',
+		format: 'standard',
+		conditions: ['6b4e71e7-4787-42f7-a092-8684961895db'], // a random, but shared default condition
```
Antonio-Laguna, line 13: Not entirely sure what this is from reading the code, given it's shared below, should it be made a constant instead of a literal?

plugins/postcss-design-tokens/src/data-formats/parse-import.ts
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	const resolvedPath = path.resolve(path.dirname(sourceFilePath), filePath);
+	if (alreadyImported.has(resolvedPath)) {
+		return false;
+	}
+
+	alreadyImported.add(resolvedPath);
+
+	const fileContents = await fsp.readFile(resolvedPath, 'utf8');
```
Antonio-Laguna, line 52: This could throw, should some sort of try/catch be added here that handles failure?

plugins/postcss-design-tokens/src/data-formats/style-dictionary/v3/value.ts
```diff
@@ -0,0 +1,106 @@
...
+		if (!dimension || dimension.unit === transformOptions.toUnit) {
+			return `${value}`;
+		}
+
+		if (dimension.unit === 'rem' && transformOptions.toUnit === 'px') {
+			return remToPx(parseFloat(dimension.number), transformOptions.pluginOptions?.rootFontSize ?? 16);
+		}
+
+		if (dimension.unit === 'px' && transformOptions.toUnit === 'rem') {
+			return pxToRem(parseFloat(dimension.number), transformOptions.pluginOptions?.rootFontSize ?? 16);
```
Antonio-Laguna, line 73: Should we validate & parse `pluginOptions` this would limit the number of checks we need to do down the road and we can offer meaningful errors. 

It's OK to handle later just thinking what happens if someone puts `'foo'` and starts spitting `NaN`, do we want to warn?

plugins/postcss-design-tokens/README.md
Antonio-Laguna: Perhaps this could be a plugin option with that as default?

plugins/postcss-design-tokens/test/basic.css
```diff
@@ -0,0 +1,54 @@
+@import url('./imported.css');
```
Antonio-Laguna, line 1: I love this test!

Could we have this sort of test too? Is it even supported?

```pcss
/* styles.css */
@import url('./vars.css');
@import url('./base.css');
```

```pcss
/* vars.css */
@design-tokens url('./tokens/basic.json') format('style-dictionary3');

:root {
  --ff-primary: design-token('font.family.base');
}
```

```pcss
/* base.css */
.is-color-base {
  color: design-token('color.font.base');
}
```

I think this is the natural path on mid to large CSS codebases, you have a single file that sets everything up and everything else is kept separated. This can also alleviate if you need `styles-dark.css` as a different entrypoint that just uses a different `var.css` file.

Note that `base.css` does not need to import design-tokens anymore which I don't think is showcased in these tests (could be wrong since I'm reading top to bottom and doing a thorough read 😂

equinusocio:
`design-token` is still too long imo. Can't we go for something shorter like `token()`?

plugins/postcss-design-tokens/README.md
equinusocio: `design-token` is still too long imo. Can't we go for something shorter like `token()`?

romainmenke:
> design-token is still too long imo. Can't we go for something shorter like token()?

I absolutely want to avoid having a future conflict with native CSS. The only way to get this guarantee would be to have `token` reserved for preprocessors. 
I am however unsure if this is something that will ever be accepted in the CSS spec as `token` could be the best name for a completely different feature. 

`design-token` however is unambiguous and getting everyone to reserve this for tooling might be achievable.

----------

Editor support will help here.

Type `spa` and get suggestions for things with `spacing` :

Screenshot 2022-05-22 at 13 47 02

Type `16px` and get suggestions for things with value `16px` :

Screenshot 2022-05-22 at 13 46 43

`design-token('...')` gets inserted, you won't ever need to type it manually :

Screenshot 2022-05-22 at 10 59 43

Hover `design-token('...')` and get more info :

Screenshot 2022-05-22 at 10 59 48

_Without editor support it will be a pain to use and `token` vs `design-token` will only be a small part if it._

plugins/postcss-design-tokens/README.md
romainmenke: It is easy to mistake default for "best practice" and I'd rather not promote a single format as they are all problematic at the moment :/

If there ever is a format that is truly good for code authors I want to make that one the default, but `style-dictionary3` wont be it.

Making this a plugin option will make code less portable.
If you copy/paste between two projects you will need to make sure they both have the same default configured.

plugins/postcss-design-tokens/README.md
romainmenke: Yes and done!

plugins/postcss-design-tokens/src/data-formats/parse-import.ts
romainmenke: The caller has try catch as it also has access to PostCSS objects and can emit PostCSS warnings :

```js
// plugins/postcss-design-tokens/src/index.ts
let importResult: { filePath: string, tokens: Map<string, Token> }|false;
try {
	importResult = await tokensFromImport(buildIs, atRule.filePath, atRule.params, importedFiles);
	if (!importResult) {
		continue;
	}
} catch (e) {
	atRule.node.warn(result, `Failed to import design tokens from "${atRule.params}" with error:\n\t` + (e as Error).message);
	continue;
}
```

plugins/postcss-design-tokens/src/data-formats/style-dictionary/v3/value.ts
romainmenke: I added a bunch of type checks around plugin options and value parsing.
This should make the whole more predictable.

> we can offer meaningful errors.

We'll handle this later after we gather some feedback.
For most other plugins we also don't give feedback on incorrect plugin options and we don't receive too many questions about this :)

plugins/postcss-design-tokens/test/basic.css
romainmenke: > This can also alleviate if you need styles-dark.css as a different entrypoint that just uses a different var.css file.

This should not be needed and I think it would even best be avoided.
By using `when` other tools (editor) can also be aware of conditional tokens.

```css
@design-tokens url('./tokens/basic.json') format('style-dictionary3');
@design-tokens url('./tokens/basic--theme-a.json') format('style-dictionary3') when('theme-a');

.foo {
  color: design-token('something.from.basic');
}
```

An editor plugin should understand `when` and give you feedback that `color` can have one of two possible values. This will give you greater insights into the design system.

------

> I think this is the natural path on mid to large CSS codebases, you have a single file that sets everything up and everything else is kept separated.

> Note that base.css does not need to import design-tokens anymore which I don't think is showcased in these tests (could be wrong since I'm reading top to bottom and doing a thorough read

True, this just works when using `postcss-import`.
But it also places extra burden on editor plugins.

They might not be able to reliably determine how you are building your CSS and which imports apply.

```
- entrypoint
  - a (has @design-tokens rules)
  - b
    - c (active file)
```

Not every tool/editor can figure out that everything comes together in `entrypoint` when you have `c` open as `c` doesn't reference `entrypoint` at all.

Adding a test is not really needed as it is not a feature we provide, but part of `postcss-import`.

plugins/postcss-design-tokens/src/data-formats/parse-import.ts
romainmenke: Made it a constant

plugins/postcss-design-tokens/test/basic.css
romainmenke: >An editor plugin should understand when and give you feedback that color can have one of two possible values. This will give you greater insights into the design system.

Example implementation of this :

- hover on `design-token` shows a table with all possible values and their conditions
- you can click through on the values to go to the tokens file

Screenshot 2022-05-22 at 18 12 30

plugins/postcss-design-tokens/README.md
Antonio-Laguna: That makes sense,  given that most usages will rely on `postcss-import` it's not something you're going to write often.

plugins/postcss-design-tokens/src/data-formats/parse-import.ts
Antonio-Laguna: Didn't see that! Thanks for catching this!

plugins/postcss-design-tokens/test/basic.css
Antonio-Laguna: All of this makes sense!

plugins/postcss-design-tokens/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+<format-name> = [ 'style-dictionary3' ]
+```
+
+All `@design-tokens` rules in a document are evaluated in order of appearance.
+If a token with the same path and name already exists it will be overridden.
+
+All `@design-tokens` rules are evaluated before any `design-token()` functions.
+
+`@design-tokens` rules can be conditional through `when` conditions. Multiple values can be specified in `when`.
```
Antonio-Laguna, line 232: Should we add a comment that specifies that when multiple it means `AND`, not `OR`? Or something as in, all conditions must be met?

rollup/presets/package-typescript.js
```diff
@@ -21,7 +21,7 @@ export function packageTypescript() {
 					extensions: ['.js', '.ts'],
 					presets: packageBabelPreset,
 				}),
-				terser(),
+				// terser(),
```
Antonio-Laguna, line 24: Should we revert this?

romainmenke: done, good catch!

plugins/postcss-design-tokens/README.md
romainmenke: Added this bit : 
> Multiple conditions always have an `AND` relationship.

Antonio-Laguna (approved):
LGTM!

PR #305, opened 2022-03-15
add deprecation notices

romainmenke:
see : https://github.com/csstools/postcss-plugins/discussions/192

- `postcss-env-function` is fully deprecated. It might come back in a different form in the future, but at the moment there is no spec for author defined values here.
- `postcss-custom-properties` only has deprecation warnings when using `importFrom` or `exportTo`.

_custom media and custom selectors are not housed in this repo._

### DIFF
--- a/plugins/postcss-custom-properties/src/index.ts
+++ b/plugins/postcss-custom-properties/src/index.ts
@@ -91,7 +91,7 @@ const creator: PluginCreator<PluginOptions> = (opts?: PluginOptions) => {
 					},
 					OnceExit: (root, { result }) => {
 						if (!disableDeprecationNotice && (importFrom.length > 0 || exportTo.length > 0)) {
-							root.warn(result, '"importFrom" and "exportTo" will be removed in a future version of postcss-custom-properties.\nCheck the discussion on github for more details. https://github.com/csstools/postcss-plugins/discussions/192');
+							root.warn(result, '"importFrom" and "exportTo" will be removed in a future version of postcss-custom-properties.\nWe are looking for insights and anecdotes on how these features are used so that we can design the best alternative.\nPlease let us know if our proposal will work for you.\nVisit the discussion on github for more details. https://github.com/csstools/postcss-plugins/discussions/192');
 						}
 
 						customProperties.clear();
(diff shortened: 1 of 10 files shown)


plugins/postcss-env-function/README.md
```diff
@@ -129,6 +129,13 @@ See example imports written in [JS](test/import-variables.js) and [JSON](test/im
 Currently only valid [custom property names] (beginning with `--`) are accepted.
 Not all valid [declaration value names] are accepted.
 
+### disableDeprecationNotice
+
+Silence the deprecation notice that is printed to the console when using `importFrom``.
+
+> "importFrom" will be removed in a future version of postcss-env-function.
```
romainmenke, line 136:
```suggestion
> postcss-env-function is deprecated and will be removed.
```

plugins/postcss-custom-properties/src/index.ts
```diff
@@ -86,7 +89,11 @@ const creator: PluginCreator<PluginOptions> = (opts?: PluginOptions) => {
 					Declaration: (decl) => {
 						transformProperties(decl, customProperties, { preserve });
 					},
-					OnceExit: () => {
+					OnceExit: (root, { result }) => {
+						if (!disableDeprecationNotice && (importFrom.length > 0 || exportTo.length > 0)) {
+							root.warn(result, '"importFrom" and "exportTo" will be removed in a future version of postcss-custom-properties.\nCheck the discussion on github for more details. https://github.com/csstools/postcss-plugins/discussions/192');
```
Antonio-Laguna, line 94: Should we also encourage to weigh in on the discussion?

Antonio-Laguna (commented):
This looks great.

I wonder if we should make the option less generic but I doubt we'll deprecate two things at once.

My only point is to try and make the messages more welcoming for user's feedback.

plugins/postcss-custom-properties/src/index.ts
romainmenke: Good point!

I initially tried to keep these brief.
But then I added the flag to silence them.

Something like this :

> "importFrom" and "exportTo" will be removed in a future version of postcss-custom-properties.
> We are looking for insights and anecdotes on how these features are used so that we can design the best alternative.
> Please let us know if our proposal will not work for you.
> Visit the discussion on GitHub for more details. https://github.com/csstools/postcss-plugins/discussions/192

Antonio-Laguna (approved):

PR #308, opened 2022-03-16
postcss-preset-env : readme - fix list of ids link

cncolder:
keep doc link up-to-date


romainmenke:
This looks good!

🤔 but maybe time we fixed this in a better way?
We will keep breaking this link and it's not that helpful anyway.

Users shouldn't have to browser source code to find basic config.

Antonio-Laguna:
Thanks @cncolder for spotting this! Merging this and we'll find some way to make sure this doesn't happen again if we move things around again!

PR #316, opened 2022-03-23
Preset env home revamp

### LINKED ISSUE #254: More info to main page
[Main page](https://preset-env.cssdb.org/) looks really awesome. But it is not really informational.

What questions visitor will have:

1. What is this site about?
2. What is `postcss-preset-env`?
3. Why should I use it?
4. Is it small “indie” tool or everyone is using it.

If we will answer this questions on the main page, it will improve conversion.

Antonio-Laguna:
This is _yet another_ site rewriting. Astro is too experimental which is great but we're betting on stability and long term maintenance. This PR changes Astro by Eleventy.

This also tackles a long-forgotten issue from @ai which revamps the Homepage. Thanks to @severin-d too which did the original work and I was able to get most of that back :) Removed gulp / grunt since they're no longer relevant and replaced with PostCSS disclaimer.

Sadly this removes the playground for now as it doesn't seem trivial to get updated, mostly due to node bindings with `importFrom` which are being discussed here: https://github.com/csstools/postcss-plugins/discussions/192

Fixes #254


romainmenke (approved):
This looks great!

Thank you for doing another port and updating the landing page :)

I suggest we merge this as-is and iterate on it where needed 🎉

PR #318, opened 2022-03-25
improve pseudo element handling

romainmenke:
The previous change to all this logic left the order of simple selectors of the same type in tact. This worked for this case :

```css
:is(:hover, :focus)::before {
  color: red;
}
```

But didn't work for some other cases.
With all the tools available to stylesheet authors to compose selectors (nesting, `:is`, custom selectors, ...) it is better to be more forgiving and have an explicit order for pseudo elements vs pseudo classes.


Antonio-Laguna (approved):
Thanks for tackling this! Looks more solid

PR #322, opened 2022-03-31
better selector specificity calculation

romainmenke:
Contains a few tiny fixes and overal smaller footprint.
Will also work better with future pseudo elements as it's no longer a hard coded list.

_No immediate need to ship this as it doesn't fix any real world issues (yet).
Can be bundled with a next release._


Antonio-Laguna (approved):
Looks good to me!

PR #327, opened 2022-04-04
update has pseudo polyfill

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/6952#issuecomment-1042654695

The `querySelector` polyfill has been updated to match Safari

### DIFF
--- a/experimental/css-has-pseudo/test/_browser.html
+++ b/experimental/css-has-pseudo/test/_browser.html
@@ -10,7 +10,7 @@
 	<!-- This is the real test stylesheet and has correct CORS attributes and http headers -->
 	<link rel="stylesheet" href="http://localhost:8081/test/browser.expect.css" crossorigin="anonymous">
 	<script src="http://localhost:8082/dist/browser-global.js"></script>
-	<script>cssHasPseudo(document, { observedAttributes: ['attrname'], debug: true, hover: true, forcePolyfill: true });</script>
+	<script>cssHasPseudo(document, { observedAttributes: ['attrname'], debug: false, hover: true, forcePolyfill: true });</script>
 </head>
 <body>
 	<the-fixture id="fixture"></the-fixture>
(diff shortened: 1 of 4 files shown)


Antonio-Laguna:
@romainmenke is this correct? I'm not seeing any tests nor files changed except for core-web which doesn't seem to include anything related to this on 0.7.0 but rather the drop on Intl we briefly chatted about.

romainmenke:
PR comment :

Screenshot 2022-04-04 at 23 42 31

Release notes :

Screenshot 2022-04-04 at 23 42 23

TIL : it's not because issues and PR's have fancy links that everything else will :D

-----

I'll also add a specific test case for this.
That will make it easier to detect and illustrate it here.

experimental/css-has-pseudo/test/_browser.html
```diff
@@ -10,7 +10,7 @@
 	<!-- This is the real test stylesheet and has correct CORS attributes and http headers -->
 	<link rel="stylesheet" href="http://localhost:8081/test/browser.expect.css" crossorigin="anonymous">
 	<script src="http://localhost:8082/dist/browser-global.js"></script>
-	<script>cssHasPseudo(document, { observedAttributes: ['attrname'], debug: true, hover: true, forcePolyfill: true });</script>
+	<script>cssHasPseudo(document, { observedAttributes: ['attrname'], debug: false, hover: true, forcePolyfill: true });</script>
```
romainmenke, line 13: `true` is very noisy because of how style invalidations work.

Best to set this to false for our own tests unless we are actually debugging the tests.

experimental/css-has-pseudo/test/_browser.html
```diff
@@ -733,8 +733,8 @@
 
 			function testColor(el, color) {
 				var actual = getComputedStyle(el).color;
-				if (actual !== color) {
-					throw new Error('nested :has is allowed : div#' + el.id + '.color; expected ' + color + ' but got ' + actual);
+				if (actual === color) {
+					throw new Error('nested :has should not be allowed : div#' + el.id + '.color; expected ' + color + ' but got ' + actual);
```
romainmenke, line 737: I added the test to main with the inverse check and that passes.
After the `core-web` update it fails 🎉

Antonio-Laguna (approved):
Thanks for adding these tests!

PR #332, opened 2022-04-08
Fix race condition issue of postcss-custom-properties

### LINKED ISSUE #331: A race condition issue in `postcss-custom-properties`
### Bug description

Recently I found a race condition issue of `postcss-custom-properties` while using it with other async PostCSS plugins.

I cost a lot of time to figure out the root issue and finally I can provide a easy way to reproduce this bug.

Full example repo: https://github.com/malash/postcss-plugins-issue-331

To reproduce this issue you need:

1. Use `postcss-custom-properties` with another async plugin ( which means the listeners of plugin should return a promise ).

I implemented this fake plugin:

```tsx
// this fake async plugin do nothing but await for next tick
const fakeAsyncPlugin = (opts = {}) => {
  // Plugin creator to check options or prepare caches
  return {
    postcssPlugin: 'A fake async plugin',
    Declaration: async () => {
      await new Promise(resolve => setTimeout(resolve, 0));
    },
  };
};
fakeAsyncPlugin.postcss = true;
```

2. Create a PostCSS instance and make sure the async runs before `postcss-custom-properties`.

```tsx
const postcssInstance = postcss([
  fakeAsyncPlugin(),
  postcssCustomProperties({
    preserve: false,
  }),
]);
```

3. Call `process` method in parallel. Make sure one CSS contains custom properties and others doesn't.

```tsx
const WITH_CUSTOM_PROPERTIES = `
:root {
  --color: red;
}

h1 {
  color: var(--color);
}
`;

const WITHOUT_CUSTOM_PROPERTIES = `
div {
  height: 10px;
}
`;

await Promise.all([
  postcssInstance.process(WITH_CUSTOM_PROPERTIES);
  postcssInstance.process(WITHOUT_CUSTOM_PROPERTIES);
  postcssInstance.process(WITHOUT_CUSTOM_PROPERTIES);
  postcssInstance.process(WITHOUT_CUSTOM_PROPERTIES);
  ...
]);
```

4. Check the output of `WITH_CUSTOM_PROPERTIES`, the `var(--color)` is not replaced as expected.

> In my case I use PostCSS with Webpack, the `postcss-loader` usually run in parallel, that's why I fall into this bug.

### Root issue and solution

I believe the root cause is this line:

https://github.com/csstools/postcss-plugins/blob/e9893a0d0d1fc8ed59efa843c5eff2750cc081f3/plugins/postcss-custom-properties/src/index.ts#L51

The plugin function is initialed only once in the `postcss([pluginFn()])` call. But the `prepare` listener runs for each CSS input file. In this plugin, the `customProperties` should be placed `prepare function.

```tsx
// let customProperties: Map<string, valuesParser.ParsedValue> = new Map();

return {
  postcssPlugin: 'postcss-custom-properties',
  prepare() {
    // move to here
    let customProperties: Map<string, valuesParser.ParsedValue> = new Map();
    return {
      Once:() => {},
      Declaration: () => {},
    };
  },
};
```

### Source CSS

```css
:root {
  --color: red;
}

h1 {
  color: var(--color);
}
```

### Expected CSS

```css
h1 {
  color: red;
}
```

### Actual CSS

```css
:root {
  --color: red;
}

h1 {
  color: var(--color);
}
```

No

### Debug output

_No response_

### Extra config

_No response_

_No response_

### Plugin version

12.1.6

macOS

### Node Version

16.13.0

### Validations

malash:
This PR fixed #331 . For more background please see that issue.

As I understand the `prepare` function is designed to store context variables for each CSS file, while the plugin function itself should contain some common code across CSS files, like some constant variables only related to `opts`.

In this case the `customProperties` is mutated and read between listeners. So it should be placed in `prepare` rather tha plugin function.

Ref: https://github.com/postcss/postcss/blob/main/docs/writing-a-plugin.md#step-3-find-nodes

### DIFF
--- a/plugins/postcss-custom-properties/src/index.ts
+++ b/plugins/postcss-custom-properties/src/index.ts
@@ -48,14 +48,14 @@ const creator: PluginCreator<PluginOptions> = (opts?: PluginOptions) => {
 	// promise any custom selectors are imported
 	const customPropertiesPromise = getCustomPropertiesFromImports(importFrom);
 
-	let customProperties: Map<string, valuesParser.ParsedValue> = new Map();
-
 	// whether to return synchronous function if no asynchronous operations are requested
 	const canReturnSyncFunction = importFrom.length === 0 && exportTo.length === 0;
 
 	return {
 		postcssPlugin: 'postcss-custom-properties',
 		prepare () {
+			let customProperties: Map<string, valuesParser.ParsedValue> = new Map();
+
 			if (canReturnSyncFunction) {
 				return {
 					Once: (root) => {


Antonio-Laguna:
@malash could you merge `main` into this PR? We've hopefully enabled tests on PRs with the last commit, would love to see them here :)

malash:
@Antonio-Laguna PR rebased and all CI passed.

Antonio-Laguna (approved):
LGTM! 

Thanks @malash !

PR #333, opened 2022-04-10
Ignore has inside supports

### LINKED ISSUE #312: Ignore `:has()` selectors nested inside `@supports` rules
### Bug description

`@supports` can now also check for selector features.

`:has()` should not be transformed if the stylesheet author made the selector conditional.

### Source CSS

```css
@supports (selector(:has(works))) {
  :has(foo) {}
}
```

### Expected CSS

```css
@supports (selector(:has(works))) {
  :has(foo) {}
}
```

### Actual CSS

```css
@supports (selector(:has(works))) {
  [\:has\(foo\)] {}
  :has(foo) {}
}
```

Yes

### Debug output

```css
@supports (selector(:has(works))) {
  [\:has\(foo\)] {}
  :has(foo) {}
}
```

### Extra config

_No response_

CSS Has Pseudo

### Plugin version

any

_No response_

### Node Version

any

### Validations

Antonio-Laguna:
This ignores `:has()` rules defined within `@support selector` check (only if it's checking `:has()` support.

Note that the `selector` syntax differs from the issue. The issue was suggesting `@supports (selector(:has(foo)))` whereas the syntax seems to be `@supports selector(:has(foo))`.

Fixes #312

### DIFF
--- a/experimental/css-has-pseudo/CHANGELOG.md
+++ b/experimental/css-has-pseudo/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changes to CSS Has Pseudo
 
+### Unreleased (minor)
+
+- Rules within `@supports` selector check for `:has` won't be transformed. This respects stylesheet author of making the selector conditional:
+
+```pcss
+@supports selector(:has(:focus)) {
+	/* Won't be transformed */
+	:has(:focus) {}
+}
+```
+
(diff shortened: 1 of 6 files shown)


romainmenke (commented):
One small thing :)

romainmenke (approved):
LGTM!

PR #353, opened 2022-04-26
use the new selector specificity package

romainmenke:


rollup/configs/externals.js
```diff
@@ -104,7 +105,6 @@ export const externalsForPlugin = [
 	'postcss-selector-not',
 	'postcss-selector-parser',
 	'postcss-value-parser',
-	'postcss-values-parser',
```
Antonio-Laguna, line 107: Nice catch

Antonio-Laguna (approved):
Looks great! 

Glad we can reuse the code here!

PR #357, opened 2022-05-01
package json lint error as github annotation

romainmenke:
Noticed that there wasn't a github annotation for lint issues in `package.json`.
Then also noticed that some of the annotations from `postcss-tape` didn't resolve to the correct path.

This fixes both issues.

### DIFF
--- a/.github/bin/format-package-json.mjs
+++ b/.github/bin/format-package-json.mjs
@@ -199,7 +199,7 @@ const formatted = {};
 Object.assign(formatted, packageJSONInfo);
 
 if (process.env.GITHUB_ACTIONS && JSON.stringify(formatted, null, '\t') !== packageJSONInfoCopy) {
-	console.error('package.json has an incorrect field order. Run "npm run lint" to resolve.')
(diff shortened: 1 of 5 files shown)


.github/bin/format-package-json.mjs
```diff
@@ -199,7 +200,7 @@ const formatted = {};
 Object.assign(formatted, packageJSONInfo);
 
 if (process.env.GITHUB_ACTIONS && JSON.stringify(formatted, null, '\t') !== packageJSONInfoCopy) {
-	console.error('package.json has an incorrect field order. Run "npm run lint" to resolve.')
+	console.error(`::error file=${path.relative(process.env.GITHUB_WORKSPACE, path.resolve('./package.json'))},line=1,col=1::package.json has an incorrect field order. Run "npm run lint" to resolve.`);
```
romainmenke, line 203: Didn't want to include the entire formatting logic here as it is a pretty static message.

Antonio-Laguna (approved):
LGTM!

Thanks! Learning a lot from these annotations!

PR #358, opened 2022-05-01
Color Helpers

Antonio-Laguna:
Extracting and abstracting color functions to its own package that gets distributed across different plugins instead of being copied over


Antonio-Laguna:
This should have all of the functions we've been using.

I think the summary is:

- Create folders and files as needed
- Add documentation that was lost given most of it needs JSDoc for licensing
- Ensuring names are consistent

From this, I feel like the next step would be to create vanity files that export all the things and that get bundled and then start to cleaning up  and make sure checks aren't blowing but I can be obviously wrong.

romainmenke:
I like all of this!

I would not have thought of giving each it's own file but I prefer this over the big fat files we had before.

_This was very daunting in the github UI, but became much more clear once I pulled it locally were I could easily browse the package as a whole :) _

Antonio-Laguna:
@romainmenke as we discussed, I've added a little script that hashes the files we want to track on external sources, you can verify it works by just changing the order in the array

packages/color-helpers/package.json
```diff
@@ -37,23 +37,21 @@
 		"README.md",
 		"dist"
 	],
-	"eslintConfig": {
-		"rules": {
-			"@typescript-eslint/no-loss-of-precision": "off"
-		}
+	"devDependencies": {
+		"node-fetch": "^3.2.4"
```
romainmenke, line 41: Should we switch to node 18 as the main version?
This has `fetch` natively.

Antonio-Laguna: Doing that works great but logs this: 

```
 node scripts/check-changes.mjs
(node:14076) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
No changes detected
```

Should we [live on the edge](https://www.youtube.com/watch?v=7nqcL0mjMjw) here? 

FWIW I'm OK with the warning :)

romainmenke: Maybe not :D

We can add an issue to switch to native node fetch whenever it no longer is experimental. Don't want to block development if there ever is an API change :)

rollup/presets/package-typescript.js
```diff
@@ -1,16 +1,21 @@
...
 import { terser } from 'rollup-plugin-terser';
+import path from 'path';
 import { packageBabelPreset } from '../configs/babel-presets';
 import { externalsForPlugin } from '../configs/externals';
 
-export function packageTypescript() {
-	return [
-		{
-			input: 'src/index.ts',
+export function packageTypescript(pkg) {
```
Antonio-Laguna, line 8: @romainmenke let me know your thoughts about this change. This moves the configuration decisions to `exports` which feel more sensible. I haven't tested out of this nor consolidated the JS counterpart but wanted to check if this worked and how did you feel about it.

rollup/presets/package-typescript.js
```diff
@@ -1,16 +1,21 @@
...
+	return Object.keys(pkg.exports).map((exportKey) => {
+		const cjsFileName = pkg.exports[exportKey].require;
+		const mjsFileName = pkg.exports[exportKey].import;
+		const input = path.resolve('/src/', exportKey, './index.ts').slice(1);
```
Antonio-Laguna, line 12: If `src/` is used instead, paths become relative to the running script which is undesirable. That's why it's being sliced after the resolving happens.

rollup/presets/package-typescript.js
Antonio-Laguna: Tests already suggest I need to look into this further

Antonio-Laguna:
@romainmenke I've finished what we discussed, I'm not sure about the import and tree-shakeability especially there. I remember that during our internal discussions you did some testing for this, can you check how we're faring now? I think we're sacrificing a bit of DX here but I'm OK as long as we're not creating a massive bundle.

romainmenke (approved):
> I've finished what we discussed, I'm not sure about the import and tree-shakeability especially there. I remember that during our internal discussions you did some testing for this, can you check how we're faring now? I think we're sacrificing a bit of DX here but I'm OK as long as we're not creating a massive bundle.

I think the current approach works really well for us.
It was really easy for me to use the code from this package and to migrate all the plugins!

Tree shaking is not something that I am currently concerned about because we aren't bundling this package.

I think we will see some gains because we are sharing the same code between multiple packages which is really nice on its own 🎉 

-----------------

This is absolutely awesome!
It took a while to finally review and integrate all this because we got side tracked with the major effort to ship version 8. But really happy to finally get around to this one :)

For me this is good to go 🚀 
There was one test value that changed between this and the old code, but CSSWG also did some code changes, so nothing unexpected.

packages/color-helpers/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Changes to Color Helpers
 
-### Unreleased
+### Unreleased (major)
```
Antonio-Laguna, line 3: I don't know if this will be picked up by the release plan since it needs the --public flag the first time

romainmenke: Good catch! I've modified the publishing script to handle this.

Antonio-Laguna:
@romainmenke I've done a final review here and I think this is good to go too :)

romainmenke:
Nice! you do the honors :)

PR #362, opened 2022-05-02
Stepped Functions & Unset Value to Preset Env/CLI

Antonio-Laguna:
This takes care of pushing the new package to preset env and the cli and also keeping up to date preset env's site.

If this is good, preset env will be released and then CLI and the site will be both updated and CLI released.


romainmenke:
Can we also add `postcss-unset-value` to preset-env?

_Hadn't done that yet as I was hoping to have some color related work done more quickly._

Antonio-Laguna:
@romainmenke yep, doing that!

cli/csstools-cli/src/plugins/postcss-unset-value.ts
```diff
@@ -0,0 +1,18 @@
+import plugin from '@csstools/postcss-unset-value';
+import { cli, helpTextLogger } from '@csstools/base-cli';
+
+export default function postcssUnsetValue() {
+	cli(
+		plugin,
+		['preserve', 'onInvalid'],
```
romainmenke, line 7:
```suggestion
		['preserve'],
```

romainmenke (commented):
One small thing, rest looks perfect 🎉

romainmenke (approved):
LGTM!

Thank you for both of these 🎉

Issue #363, opened 2022-05-05
`:any-link` issue with nesting


Antonio-Laguna:
### Bug description

There seems to be an interop issue with `postcss-pseudo-class-any-link` and `postcss-nesting`. When combined with an `anchor` it produces a selector that won't really match anything.

Look at output: `aarea[href]`

### Source CSS

```pcss
a  {
	font-size: 1rem;

	&:any-link {
		text-decoration: none;
	}
}
```

### Expected CSS

```pcss
a {
	font-size: 1rem
}

a:link, a:visited {
	text-decoration: none;
}

a:-webkit-any-link {
	text-decoration: none;
}

a:-moz-any-link {
	text-decoration: none;
}

a:any-link {
	text-decoration: none;
}
```

### Actual CSS

```pcss
a {
	font-size: 1rem
}

a:link, a:visited, aarea[href] {
	text-decoration: none;
}

a:-webkit-any-link {
	text-decoration: none;
}

a:-moz-any-link {
	text-decoration: none;
}

a:any-link {
	text-decoration: none;
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

Yes

### Plugin version

7.5.0

### What OS are you experiencing this on?

macOS

### Node Version

16.13.2

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

PR #364, opened 2022-05-05
any-link : improve fallbacks for area

romainmenke:
should fix : https://github.com/csstools/postcss-plugins/issues/363

I am a bit unsure how it will interact with nesting after the fix.


Antonio-Laguna (approved):
Thanks for tackling this so fast, looks solid!

PR #374, opened 2022-05-12
postcss-preset-env : add cascade layers

romainmenke:


plugins/postcss-nesting/test/layer.css
```diff
@@ -0,0 +1,14 @@
...
+
+.contained {
+	background-color: blue;
+	height: 100px;
+	width: 100px;
+
+	@layer A {
+		/* There is no definition for this behavior */
+		background-color: red;
+	}
```
romainmenke, line 13: I have no idea if this should be allowed and how it should work exactly.

Mirroring `@media` and `@supports` handling seemed logical and should be ok for now.

_Not handling it here causes issues in the `cascade-layer` plugin if it encounters `@layer` in a style rule._

plugins/postcss-cascade-layers/src/index.ts
```diff
@@ -22,7 +22,7 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 
 	return {
 		postcssPlugin: 'postcss-cascade-layers',
-		Once(root: Container, { result }: { result: Result }) {
+		OnceExit(root: Container, { result }: { result: Result }) {
```
romainmenke, line 25: Better interop with `:has`, `:is`, `@custom-selector`, nesting, ...

romainmenke:
Size diff with and without some cascade layers : 

with cascade layers :

```
+--------------------------------------------------------------+
| Size         | 33.82 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 5.26 KiB                                      |
+--------------------------------------------------------------+
```

without :

```
+--------------------------------------------------------------+
| Size         | 27.07 KiB                                     |
|--------------------------------------------------------------|
| Gzipped      | 5.04 KiB                                      |
+--------------------------------------------------------------+
```

plugins/postcss-nesting/test/layer.css
Antonio-Laguna: That makes sense, I did a brief integration test and saw the issues!

Antonio-Laguna (approved):
This LGTM.  I'm kind of nervous with this plugin because of how drastic the transformations are! Thanks for checking this so thoroughly

romainmenke:
updated docs : https://github.com/csstools/postcss-plugins/blob/preset-env-add-cascade-layers--compassionate-giant-panda-bear-bf23b5ebc4/plugins/postcss-cascade-layers/README.md#how-it-works

Antonio-Laguna (approved):
Great work, love the docs, I think this definitely helps with #376

Issue #257, opened 2022-02-15
review : `postcss-progressive-custom-properties`


romainmenke:
Before this is used to widely we should review the implementation and usage of `postcss-progressive-custom-properties`.

My intentions was/is to extract custom property fallbacks of out individual plugins.

- custom properties have specific behaviour
- trying to find ways to limit code duplication across plugins

In practice each individual plugin still needs a bit of logic to prevent infinite loops.
Each transform needs to detect if it is inside a relevant `@support` rule and skip transformations when true.

This bit is still useful on its own and is not directly related to custom properties.
It also allows stylesheet authors to write fallbacks and manually wrap enhancements in `@supports`.

At the moment I like the result of the whole, but I don't like how all the code is split and shared.

I also like that `postcss-preset-env` can take control here and run this for all plugins in a single step.

**This is an open call for ideas and opinions on how to solve issues like these.**

- `postcss-progressive-custom-properties` : https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-progressive-custom-properties
- example usage : https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-oklab-function/src/index.ts#L68-L97
- `hasSupportsAtRuleAncestor` : https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-oklab-function/src/has-supports-at-rule-ancestor.ts

romainmenke:
Going to close for now.
The plugin does what it needs to do well enough.
I can't think of a better solution.

PR #377, opened 2022-05-14
is guarded by at supports

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/257

When working on `@layer` I realised that the value and/or selector parsers we use are suited to process `AtRule` params.

Extending `postcss-progressive-custom-properties` so that it uses existing value matchers for `@supports` params was trivial and appears to work great.

It always bugged me that we were not using a parser to check the contents of `@supports` and this looks like a good way to resolve this.

-------------

This does redefine the scope of `postcss-progressive-custom-properties` which might also be a good thing. We were never really happy with it.

Maybe this would be better :

- package dedicated to matching values and patterns
- `postcss-progressive-custom-properties` remains but uses the matching package
- `declarationIsGuardedByAtSupports` used by other plugins also comes from the matching package

-------------


romainmenke:
Too much time has passed.
I will restart this after `postcss-preset-env v8`

Issue #382, opened 2022-05-16
[postcss-cascade-layers] `:not()` selector order conflicts with pseudo elements


rbndelrio:
### Bug description

Hey! Running into an issue where `postcss-cascade-layers` appends the `:not()` selector after all pseudo elements. This causes blocks using `:before`, `:after`, etc., to be ignored by browsers. Re-ordering the selector would fix this in any case I'm aware of.

### Source CSS

```css
@layer top, bottom;

@layer top {
  h1:before { content: '👍' }
}

@layer bottom {
  h2:before { content: '👎 bug' }
}

h3:before { content: '🤔 bug' }
```

### Expected CSS

```css
h1:before { content: '👍' }
h2:not(#\#):before { content: '👎 bug' }
h3:not(#\##\#):before { content: '🤔 bug' }
```

### Actual CSS

```css
h1:before { content: '👍' }
h2:before:not(#\#) { content: '👎 bug' }
h3:before:not(#\##\#) { content: '🤔 bug' }
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

N/A

### Extra config

Default options, minimal config: `plugins: [require('@csstools/postcss-cascade-layers')()]`

### What plugin are you experiencing this issue on?

PostCSS Cascade Layers

### Plugin version

1.0.0

### What OS are you experiencing this on?

macOS

### Node Version

18.1.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@rbndelrio Thank you for reporting this.
I've prepared a fix for this issue and we will try to ship this soon.

Antonio-Laguna:
Hey @rbndelrio !

Thanks again for taking the time to report this! We've just released 1.0.1 with the fix @romainmenke did to pseudo elements, can you test with your code to ensure it now behaves as expected?

Thanks!

romainmenke:
@rbndelrio Can you confirm if this is fixed for you?
Happy to take another look at this if it's still causing issues.

PR #383, opened 2022-05-16
fix selector ordering with pseudo elements

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/382


packages/selector-specificity/test/test.mjs
```diff
@@ -7,6 +7,11 @@ function calculate(selector) {
 	return selectorSpecificity(selectorAST);
 }
 
+assert.deepEqual(calculate(':before'), { a: 0, b: 0, c: 1 });
+assert.deepEqual(calculate('::before'), { a: 0, b: 0, c: 1 });
+
+assert.deepEqual(calculate(':focus'), { a: 0, b: 1, c: 0 });
```
romainmenke, line 13: This package lacked sufficient coverage for pseudo elements vs classes.

plugins/postcss-cascade-layers/.tape.mjs
```diff
@@ -6,12 +6,6 @@ postcssTape(plugin)({
 	basic: {
 		message: "supports basic usage",
 	},
-	'basic:color': {
-		message: "supports { color: '<a color>' }",
-		options: {
-			color: 'purple'
-		}
-	},
```
romainmenke, line 14: Left over from the base plugin

plugins/postcss-cascade-layers/src/compound-selector-order.ts
```diff
@@ -67,9 +67,6 @@ function selectorTypeOrder(selector, type) {
 		return selectorTypeOrderIndex.pseudoElement;
 	}
 
-	if (type === 'pseudo' && selector && selector.indexOf('::') === 0) {
```
romainmenke, line 70: `selectorParser.isPseudoElement(selector)` on line 66 expects a `Node` and this didn't work before because it was a string.

This code on line 70 was insufficient but did work with strings.

Same change in all other plugins with similar logic.

plugins/postcss-cascade-layers/src/index.ts
```diff
@@ -22,7 +22,7 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 
 	return {
 		postcssPlugin: 'postcss-cascade-layers',
-		Once(root: Container, { result }: { result: Result }) {
+		OnceExit(root: Container, { result }: { result: Result }) {
```
romainmenke, line 25: see : https://github.com/csstools/postcss-plugins/pull/374

This change will likely be merged and released before we can add it to `preset-env` so adding it here as it is an obvious improvement.

Antonio-Laguna (approved):

Antonio-Laguna:
Awesome work! Getting this out !

PR #386, opened 2022-05-19
color functional notation : percentage values

romainmenke:
Modern browsers support using modern channel values in the legacy (comma separated) syntax.
Older browsers obviously do not support this :)


Antonio-Laguna (approved):
LGTM! Thanks for tackling this and finding a good solution!

PR #388, opened 2022-05-21
design tokens : unit conversion

romainmenke:
reading : https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

---------

When values and units are fully determined by designers and/or design tools we take away control from developers.

In some cases this is fine as developers (will soon) have tools to manipulate values in native CSS. (e.g. `calc` or relative color syntax).

But some units are different.
`px` vs. `rem` is meaningful and they aren't equivalent.

Developers should have the freedom and tools to convert values from design tokens to the correct unit for each rule and property.

```css
.unit-conversion {
	padding-bottom: design-token('space.small' to px);
	padding-bottom: design-token('space.small' to rem);
}
```

Without a conversion mechanic designers and developers will have to declare each value twice, once for each unit.

- designers might not know the root font size
- you lose the connection between components/views and used design tokens

```css
.unit-conversion {
	padding-bottom: design-token('space.small.px');
	padding-bottom: design-token('space.small.rem');
}
```

Alternatively designs and developers do nothing and settle on inaccessible results :/

------

I hope this won't explode later into a lot of conversions.
But I think it will be fine as I can only think of `rem` vs. `px` having this complexity.

If there are more units please let me know.

### DIFF
--- a/plugins/postcss-design-tokens/.tape.mjs
+++ b/plugins/postcss-design-tokens/.tape.mjs
@@ -5,10 +5,20 @@ import postcssImport from 'postcss-import';
 postcssTape(plugin)({
 	basic: {
 		message: "supports basic usage",
-		options: {},
 		plugins: [
 			postcssImport(),
 			plugin()
+		]
+	},
+	'basic:rootFontSize-20': {
+		message: "supports basic usage with { unitsAndValues { rootFontSize: 20 } }",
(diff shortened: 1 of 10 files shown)


Antonio-Laguna (approved):
Love this! Thanks for handling this!

PR #397, opened 2022-05-23
fix stage select not working

romainmenke:


sites/postcss-preset-env/src/static/js/features/index.js
```diff
@@ -5,6 +5,11 @@
 		return;
 	}
 
+	const hashElements = [];
```
romainmenke, line 8: `hashElements` was not yet defined on the first call to `onParsed`

PR #425, opened 2022-06-03
Add migrated plugins to CLI and update PostCSS Preset Env

Antonio-Laguna:
This adds the following plugins to the CLI:

- `postcss-attribute-case-insensitive`
- `postcss-custom-media`
- `postcss-custom-selectors`
- `postcss-selector-not`

It also updates CSSDB on Preset Env so the `FEATURES` doc is kept up to date. After this I will update the site so it's all completed.


cli/csstools-cli/src/plugins/postcss-custom-media.ts
```diff
@@ -0,0 +1,20 @@
+import plugin from 'postcss-attribute-case-insensitive';
```
romainmenke, line 1:
```suggestion
import plugin from 'postcss-custom-media';
```

cli/csstools-cli/src/plugins/postcss-custom-selectors.ts
```diff
@@ -0,0 +1,20 @@
+import plugin from 'postcss-attribute-case-insensitive';
```
romainmenke, line 1:
```suggestion
import plugin from 'postcss-custom-selectors';
```

romainmenke (commented):
two typos :)

romainmenke (approved):
LGTM!

PR #431, opened 2022-06-04
housekeeping after migration

romainmenke:
- see : https://github.com/csstools/postcss-plugins/projects/1#card-74855220
- spotted some files with incorrect permissions
  - set correct permissions on all files
  - added a CI check for regressions


Antonio-Laguna (approved):
LGTM! Thanks for cleaning this up!

PR #436, opened 2022-06-06
prefers color scheme

romainmenke:
> - Breaking : removed old CDN urls
> - Breaking : remove `color-depth` queries fallback
> - Breaking : remove 'no-preference' support as this was dropped from the spec
> - Breaking : remove old global object

- cleaned up old and deprecated parts of the plugin
- converted the PostCSS plugin to typescript
- added browser tests

All docs have also been moved to a single [readme](https://github.com/csstools/postcss-plugins/tree/prefers-color-scheme--compassionate-alpine-dachsbracke-f5f77b6255/plugins/css-prefers-color-scheme#browser)

This was previously split up between browser and non-browser docs.
I hope that documenting it all together also makes it very clear how it all works.

------

_the individual issues have more details_

- fixes : https://github.com/csstools/postcss-plugins/issues/441
- fixes : https://github.com/csstools/postcss-plugins/issues/444
- fixes : https://github.com/csstools/postcss-plugins/issues/448
- fixes : https://github.com/csstools/postcss-plugins/issues/450


plugins/css-prefers-color-scheme/package.json
```diff
@@ -35,23 +47,21 @@
 		"CHANGELOG.md",
 		"LICENSE.md",
 		"README.md",
-		"browser.js",
-		"browser.min.js",
 		"dist"
 	],
 	"peerDependencies": {
-		"postcss": "^8.4"
+		"postcss": "^8.3"
```
Antonio-Laguna, line 53: Any specific reason for this? Or is it just automatic change?

Antonio-Laguna (approved):
Nice work!

I started this page as a stub for these changes, https://github.com/csstools/postcss-plugins/wiki/PostCSS-Preset-Env-8

It might be worth to both, point to that on the CHANGELOG and also add any info on a per plugin basis. In this case, perhaps it might be even worth to point to the specific spec we're referring to on that wiki page which should be https://www.w3.org/TR/2021/WD-mediaqueries-5-20211218/#prefers-color-scheme (in case it changes again, though unlikely)

plugins/css-prefers-color-scheme/test/_browser.mjs
```diff
@@ -0,0 +1,158 @@
+import puppeteer from 'puppeteer';
```
Antonio-Laguna, line 1: Shouldn't this be listed as `devDependencies` of this package?

plugins/css-prefers-color-scheme/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changes to Prefers Color Scheme
 
+### Unreleased (major)
+
+- Breaking : removed old CDN urls
+- Breaking : remove `color-depth` queries fallback
+- Breaking : remove 'no-preference' support as this was dropped from the spec
+- Breaking : remove old global object
```
Antonio-Laguna, line 8: Very much nit but keeps consistent with this and all of the other CHANGELOG files

```suggestion
- Breaking: removed old CDN urls
- Breaking: remove `color-depth` queries fallback
- Breaking: remove 'no-preference' support as this was dropped from the spec
- Breaking: remove old global object
```

plugins/css-prefers-color-scheme/README.md
```diff
@@ -1,95 +1,321 @@
...
+		--site-color: #222;
+	}
+}
+```
+
+## Browser
+
+```js
+// initialize prefersColorScheme (applies the current OS color scheme, if available)
+const prefersColorScheme = require('css-prefers-color-scheme/browser')();
```
Antonio-Laguna, line 125: Is the immediate invocation correct? Also, should we start advocating for `import`?

plugins/css-prefers-color-scheme/CHANGELOG.md
romainmenke: Good catch!

plugins/css-prefers-color-scheme/test/_browser.mjs
romainmenke: yup!

plugins/css-prefers-color-scheme/package.json
romainmenke: Going to create a separate issue for this

plugins/css-prefers-color-scheme/README.md
romainmenke: > Is the immediate invocation correct?

It is correct, but I don't think it is very clear. (old docs)

I've updated all these examples to use `import`.
The immediate invocation is now also gone.

Antonio-Laguna (approved):
Nice one!

Issue #435, opened 2022-06-03
Using plain breakpoints defintion from root


lhtdesignde:
Hi!
I was going through the documentation but could not get it working so I need to ask. I have breakpoints in root looking like this.
```
:root {
  --break-desktop: 1025px;
  --break-tablet: 1024px;
  --break-tablet-min: 600px;
  --break-mobile: 599px;
  --break-mobile-min: 320px;
}
```
I'm trying to use them just like this:
```
  @media (min-width: var(--break-desktop)) {
    background-color: red;
  }
```

This seems not to work? Am I doing something wrong?
I'm using webpack 5 and scss. Also your webpack configuration in the docu did not work with webpack 5 (options has an unknown property 'plugins') for me so I needed to adjust it a bit.
```
      rules: [
        {
          test: /\.s[ac]ss$/i,
          use: [
            'style-loader',
            {
              loader: 'css-loader',
              options: {
                importLoaders: 1,
                modules: {
                  localIdentName: '[local]-[hash:base64:5]'
                }
              }
            },
            {
              loader: 'sass-loader',
              options: {
                implementation: require('sass')
              }
            },
            {
              loader: 'postcss-loader',
              options: {
                postcssOptions: {
                  plugins: [
                    postcssCustomMedia()
                  ]
                }
              }
            }
          ]
        }
      ]
    }

```

lhtdesignde:
@romainmenke Thanks for this. I'll update my webpack config. 
It looks a bit different. In regards to why i need it in root. I'm running a more complex architecture with microfrontends. I have a shared repo that's acting as a design/pattern library. from there I'm sharing colors etc. I also need to share the breakpoints. I'm able to set it to :root but the children need to consume those somehow and till now I didn't find a good way. The older plugin seemed to be able to do what I wrote in the starting question post so I was hopeful that method could work.

romainmenke:
But why does it need to be in `:root`?

Can you try declaring them at the top level (not in `:root`)

```css
/* shared library */
@custom-media --break-desktop (min-width: 1025px);
@custom-media --break-tablet (min-width: 1024px);
@custom-media --break-tablet-min (min-width: 600px);
@custom-media --break-mobile (min-width: 599px);
@custom-media --break-mobile-min (min-width: 320px);
```

Use them like this :

```css
/* consumer */
@media (--break-desktop) {
  background-color: red;
}
```

lhtdesignde:
[...]

PR #437, opened 2022-06-06
e2e : webpack config

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/435#issue-1261409686

We had no way to test if the Webpack setup we describe in our docs actually works.
The only thing we test is that it builds and has `exit 0`

If we ever need to update the config for a future Webpack version this will also help us migrate.


Antonio-Laguna:
@romainmenke I love this idea ! LGTM but has a conflict here

romainmenke:
conflict resolved :)

Antonio-Laguna (approved):
LGTM!

Issue #439, opened 2022-06-10
`selector-specificity` throws `TypeError for `:nth-child` without arguments


ybiquitous:
### Bug description

From the Stylelint regression report: https://github.com/stylelint/stylelint/issues/6139

I found the `@csstools/selector-specificity` package throws a `TypeError` when parsing a `:nth-child` pseudo class without arguments:

```js
const parser = require("postcss-selector-parser");
const { selectorSpecificity } = require("@csstools/selector-specificity");

console.log(selectorSpecificity(parser().astSync(":nth-child")));
//=> TypeError: Cannot read properties of undefined (reading 'nodes')
```

For details, see also the reproduction on Gist:
https://gist.github.com/ybiquitous/96194528bae6ce9ac4ae1414f20141e8

Maybe, it seems due that `node.nodes[0]` returns `undefined` in the following location:

https://github.com/csstools/postcss-plugins/blob/b755ca7806b52d86276d825968832e957c5c2e8a/packages/selector-specificity/src/index.ts#L74

The package can parse `:is` pseudo classes without arguments, so I think it makes sense that it can also parse `:nth-child` without arguments. What do you think?

### Source CSS

```css
/* without arguments like `(2n+1)` */
:nth-child
```

### Expected CSS

N/A

### Actual CSS

N/A

### Debug output

N/A

### Extra config

N/A

### Plugin version

2.0.0

### What OS are you experiencing this on?

macOS

### Node Version

18.2.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

ybiquitous:
@romainmenke Thanks for the quick fix! 😄

romainmenke:
Thank you for the report!

@Antonio-Laguna will release the update shortly 🎉

Antonio-Laguna:
This has been released as `2.0.1`, thanks a lot for reporting @ybiquitous ! Please let us know if everything goes right on your end.

ybiquitous:
@romainmenke @Antonio-Laguna Thank you so much for your help! 😊

ybiquitous:
Stylelint 14.9.1 is out, including this patch release of `@csstools/selector-specificity`! 🎉 
[...]

PR #440, opened 2022-06-10
selector-specificity : fix exceptions on `:nth-child`

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/439

_I've also added a few more tests for other pseudo classes which might take arguments._


Antonio-Laguna (approved):
LGTM! Glad this is more solid!

PR #471, opened 2022-06-13
Change PostCSS version strategy

### LINKED ISSUE #465: Which PostCSS version should be used for the peer dependency config?
The peer dependency version of PostCSS has started to diverge in plugins.

https://github.com/csstools/postcss-plugins/blob/main/UPDATING_POSTCSS.md?plain=1#L3-L6

>see : `packages/postcss-tape/package.json`
>
>Set the `postcss` version to match the peer dependency of all plugins.
>Set the `postcss-oldest-supported` depending on community needs. This can be the same as the `postcss` version.

After some thought I propose we change this to :

- choose the lowest PostCSS version as the peer dependency
- also test the latest PostCSS version
- validate in CI that all plugins have the same version (part of package.json lint?)

If a plugin of ours does not require PostCSS `8.3` and works perfectly fine with the features in `8.2` it should be possible to use PostCSS `8.2`. Currently we disable usage with `8.2` arbitrarily.

Overal this should make it easier for users to install and use `postcss-preset-env` and plugins if they have little or no control over their PostCSS version. (we've seen these versions pinned by frameworks)

romainmenke:
fixes #465

-----

Should we release all the plugins?
Feels like a lot of work for something that isn't a reported issue.


Antonio-Laguna:
It is indeed a lot of work, I think we can just leave it to be and it will go gradually, if it becomes an issue at any point we can just release that plugin in particular. 

Do you foresee any issues there?

romainmenke:
> Do you foresee any issues there?

Not really.

I think this change is the better versioning strategy but I don't think there currently is an urgent need to have it out there.

Rolling it out one by one will also not cause any issues as we are relaxing the version constraint.

PR #477, opened 2022-06-16
check licenses

romainmenke:
inspired by : https://github.com/csstools/postcss-plugins/pull/475

This action did not work at all but overall I prefer something custom as this is not something that is complex.

- focus on a single ecosystem (npm)
- precisely scan dependencies (we know our own layout)
- precisely scan distributed files (we know our own layout)

-----

> Checks 2 things
> 1. Do all our dependencies have an appropriate license?
> 2. Are there unexpected license comments in build artifacts?
>
> Combined this should give our users sufficient safeguards against license issues.

We are already careful with the dependencies we use and pass on to users and this adds more checks to ensure that we do not accidentally cause issues for the community.

It should give a reasonable guarantee that the combined software our users get is appropriately licensed.


romainmenke:
check the two failing commits and their annotations for examples of issues and how these are reported.

.github/bin/format-package-json.mjs
```diff
@@ -51,6 +51,9 @@ const formatted = {};
 
 	formatted.license = packageJSONInfo.license;
 	delete packageJSONInfo.license;
+	if (formatted.license !== 'CC0-1.0' && formatted.license !== 'MIT') {
+		formatted.license = 'MIT';
+	}
```
Antonio-Laguna, line 56: Is this intended?

Antonio-Laguna (approved):
LGTM! Just a minor question

.github/bin/format-package-json.mjs
romainmenke: Not really, just copy/paste from the last string on line 54.
I will change to `CC0-1.0` which is the most common license in plugins here.

PR #484, opened 2022-06-17
`op_mini` : correctly transform when only `op_mini all` is set

romainmenke:
`op_mini all` was not a valid browserslist.
It is the only one with only `all` and needs special handling.

I've added this change in two commits so that you can see how the test output changes with the change.

_I am hoping that the added test will fail if caniuse ever adds version info for Opera Mini._


Antonio-Laguna (approved):
LGTM! Curious how you found this 😂

romainmenke:
After Andrey talked about browserslist linting on Twitter I was wondering how we handled these browsers.

I started by gathering a list of lowest versions and Opera Mini is `op_mini all` and then I came across the part where we build the browserslist queries which are all with `<` or `>`.

So 50/50 chance and random re-reading of our code base.

Issue #452, opened 2022-06-11
[blank-pseudo] housekeeping


romainmenke:
- [x] rebuild from base plugin
- [x] typescript
- [x] check docs

PR #486, opened 2022-06-18
CSS Blank Pseudo

Antonio-Laguna:
- resolves https://github.com/csstools/postcss-plugins/issues/443
- resolves https://github.com/csstools/postcss-plugins/issues/446
- resolves https://github.com/csstools/postcss-plugins/issues/449
- resolves https://github.com/csstools/postcss-plugins/issues/452


Antonio-Laguna:
@romainmenke what would be the best way to share a portion of code between polyfill and plugin? CJS vs MJS in a mixed world makes me wonder. Perhaps just better to duplicate till mjs is a thing?

romainmenke (approved):
LGTM! awesome work on this!

PR #495, opened 2022-06-21
pseudo elements must be handled like complex selectors

romainmenke:
**TL;DR;**

```css
::before:hover {}
/* not equivalent to */
:hover::before {}
```

**More context**

>  a pseudo-element is effectively a new element, the pseudo-element selector is not equivalent in meaning to any other simple selector, and is instead basically a combinator into the "pseudo-children" of an element.

https://github.com/w3c/csswg-drafts/issues/2284#issuecomment-364218218

---------

This affects every plugin that has any logic around combinators or compound vs complex selectors.

- [x] `any-link`
- [x] cascade layers
- [x] `is-pseudo`
- [x] nesting
- [x] changelogs

-----

I usually prefer to reduce problems to sorting because sorting is easy and algorithms aren't.

This just backfired because the AST we receive from `postcss-selector-parser` doesn't consider `::before` as `<combinator><tag>`.

To make things worse I was not even aware that this was an issue to look out for and I was just trying to place pseudo elements at the end of selectors. (breaking `::before:hover`)

-----

Cascade layers and `any-link` were trivial to fix because they only required an exact insertion of new selector bits. I was lazy before and always just added the new bits at the end and then sorted to repair.

The fixes here remove all this "bad" code and implement exact insertion.

You might see `:not(#\#)` or `[href]` inserted in a different position than before this change.

-----

`is-pseudo` (and thus nesting) are much more complex.
We still need the sorting there, but we must consider pseudo elements as combinators.

I've written a new sorting function for `is-pseudo` and it seems to work.

I avoided making too many changes in nesting as there is no browser implementation yet.
The change here is limited to not re-ordering pseudo elements in a way that changes the selected element.

For `is-pseudo` I've made the plugin a bit more strict and disallowed pseudo elements in `:is()`. These are not allowed by the spec. To preserve the forgiving selector list behaviour any pseudo element is replaced by a custom pseudo element which is known to be invalid. `::before` -> `::-csstools-invalid-before`


Antonio-Laguna (approved):
LGTM, nice and impressive work! So many tests!

PR #502, opened 2022-06-25
case insensitive processing and consistent cloning

romainmenke:
- ensure we handle everything case insensitive when specified as such
- ensure we always clone so that the modified nodes are revisited by all plugins


romainmenke:
_Starting to wonder if I jinxed it when I decided that relaxing the PostCSS version constraint didn't warrant a re-release of all plugins. Probably not coincidental that I find a new error class days later :D_

.github/labeler.yml
```diff
@@ -164,9 +164,9 @@
   - plugins/postcss-pseudo-class-any-link/**
   - experimental/postcss-pseudo-class-any-link/**
 
-"plugins/postcss-rebeccapurple":
-  - plugins/postcss-rebeccapurple/**
-  - experimental/postcss-rebeccapurple/**
+"plugins/postcss-color-rebeccapurple":
```
Antonio-Laguna, line 167: Nice catch!

plugins/postcss-dir-pseudo-class/src/index.js
```diff
@@ -29,19 +27,23 @@ function creator(opts) {
...
+								return;
+							}
+
+							if (!node.nodes || !node.nodes.length) {
 								return;
 							}
 
 							// value of the :dir pseudo-class
-							const value = node.nodes.toString();
+							const value = node.nodes.toString().toLowerCase();
```
Antonio-Laguna, line 39: This is later concatenated below which means this won't preserve developer casing. Is this something we should do?

plugins/postcss-gap-properties/src/index.js
```diff
@@ -11,16 +12,33 @@ function creator(opts) {
 		postcssPlugin: 'postcss-gap-properties',
 		// walk decl shorthand gap, column-gap, or row-gap declaration
 		Declaration(decl) {
-			if (gapPropertyRegExp.test(decl.prop) && decl.parent.some(isDisplayGrid)) {
```
Antonio-Laguna, line 14: ❤️

Antonio-Laguna (approved):
This is really impressive! Thanks for bringing up this so it's even closer to spec. I've left one minor comment but feel free to ignore it if it's fine.

The reasoning was that I'm OK with producing lowercase code but this seemed like transforming the user's code

plugins/postcss-dir-pseudo-class/src/index.js
romainmenke: Good catch! I was indeed trying to avoid changing things and only using lower cased values in conditionals.

I will update this 🎉

romainmenke: Coincidentally this was actually more correct imo.

- `[dir="RTL"]` or `[dir="rtl"]` is not the same.
- we generate `:not([dir="<inverse>"])`.

I think it is best to allow upper case in `:dir` and keywords but only output lowercase.

Issue #460, opened 2022-06-11
[preset-env] `v8` scope - batch 1


romainmenke:
> refresh of all plugins with client side scripts

focussed on :

- code quality
- browser tests
- spec compliance

----

Plugins :

- `css-blank-pseudo`
- `css-has-pseudo`
- `css-prefers-color-scheme`
- `postcss-focus-visible`
- `postcss-focus-within`
- `postcss-preset-env`

With `:has()` support rolling out in multiple browsers the spec is settling.
We have kept tack of these changes in an experimental plugin.

The final issues seem to have been resolved and it should be ready to ship.

As this is a breaking change we want to also do some chores on related plugins.
Over the past few years these have accumulated some cruft.

This is scoped so that migrating to `v8` of `postcss-preset-env` will only affect those users who enable client side polyfills.

romainmenke:
@Antonio-Laguna is there anything you would like to add or omit from this?

Antonio-Laguna:
@romainmenke is there any reason not to include logical and importFrom removal here ?

romainmenke:
We can include more but this will make the upgrade path for users less clear.

I suspect that each breaking change only affects a small part of our users but each time in a big way.

Making it simple to determine if a change affects you and which steps to take might help adoption.

For us it will also make support a bit simpler.
If `v8` includes all the breaking changes at once we will have to answer questions and follow up on potential bugs for a lot more plugins and use cases.

So future major versions might be :

- 9 : logical
- 10 : importFrom + general plugin options/config cleanup

------

Doing everything at once is also totally fine for me :)
As long as we can communicate clearly to users which steps they need to take to upgrade.

Antonio-Laguna:
Given the amount of work and planning around each major version, and also the amount of overhead that creates on other libraries, I'm a bit torn here. I completely see your point and at the same time, I don't want to put too much strain or appear unstable. Given the number of weekly downloads, having _semi-frequent_ major upgrades makes me wonder if that's the right path. 

Sadly I don't have an answer here or a more solid decision and I'm rather full of doubts 😂

romainmenke:
_below some thoughts, not always structured_

----

Part of my concern can be resolved with just really good migration docs :)
The Wiki might be a good place for this.

----

As a user of `postcss-preset-env` I want my projects to work the same before and after the major version bumps.

Verifying this and applying changes becomes more complex if we break more at once.

----

We have a multiple factors that cause breaking changes that are beyond our control.
[...]

PR #513, opened 2022-06-30
Updating Focus Visible

Antonio-Laguna:
This is part of #460 

This converts the plugin to use the scaffolding process we've created so docs are easier to maintain. Here is a quick summary of the changes made:

- Clarified the docs around the `replaceWith` given that W3C's polyfill does not allow to change such selector.
- Made so selectors now include the ready class `.js-focus-visible` which is added to the document so `:not(:focus-visible)` is not applied by default.
- Included browser testing as some quick integration test of the polyfill plus our plugin.

Open questions:

Should we consider the above a breaking change?
Is the selector updating solid? I found postcss-selector-parser's API very confusing to be able to just leverage that. I wonder if the method is too naive.


plugins/css-blank-pseudo/src/index.ts
```diff
@@ -2,7 +2,7 @@ import type { PluginCreator } from 'postcss';
 import parser from 'postcss-selector-parser';
 import isValidReplacement from './is-valid-replacement.mjs';
 
-type pluginOptions = { color?: string, preserve?: boolean };
+type pluginOptions = { preserve?: boolean, replaceWith?: string };
```
Antonio-Laguna, line 5: I realised this was a leftover which would have exported the wrong types, this hasn't been published yet so it's not a big concern

romainmenke (approved):
LGTM!

PR #514, opened 2022-07-01
focus-visible : code review

romainmenke:
```css
.foo:focus-visible {}

/* becomes */

.js-focus-visible.foo.focus-visible,
.js-focus-visible .foo.focus-visible {}
```

This should more reliably handle `html`, `:root`,...
We actually have no way to know if any part of the selector might match the top most element on the page.

```css
html:focus-visible {}

/* becomes */

html.js-focus-visible.focus-visible,
.js-focus-visible html.focus-visible {}
```

----

```css
.some-class-on-html:focus-visible {}

/* becomes */

.some-class-on-html.js-focus-visible.focus-visible, /* matches */
.js-focus-visible .some-class-on-html.focus-visible {}
```

```css
.something-else:focus-visible {}

/* becomes */

.something-else.js-focus-visible.focus-visible, 
.js-focus-visible .something-else.focus-visible /* matches */ {}
```

### DIFF
--- a/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
@@ -248,7 +248,7 @@
 	word-wrap: break-word;
 }
 
-.js-focus-visible .test-focus-visible-pseudo-class.focus-visible {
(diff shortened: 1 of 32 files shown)


Antonio-Laguna (approved):
Looks great! Thanks for taking a look at this and getting it robust!

PR #516, opened 2022-07-02
Updating Focus Within

Antonio-Laguna:
This is part of #460 

This converts the plugin to use the scaffolding process we've created so docs are easier to maintain. Here is a quick summary of the changes made:

- Made so selectors now include the ready class `.js-focus-within`.
- Rewrote browser polyfill to be included within the postcss plugin rather than using JSX Tool's one
- Included browser testing.

This is basically a duplicate of `focus-visible` with some enhancements from blank given that in this case we do control the Polyfill now.


plugins/postcss-focus-visible/README.md
```diff
@@ -59,10 +59,10 @@ instructions for:
 ### preserve
 
 The `preserve` option determines whether the original notation
-is preserved. By default, it is not preserved.
+is preserved. By default, it is preserved.
```
Antonio-Laguna, line 62: This was actually wrong as per https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-focus-visible/src/index.ts#L7

plugins/postcss-focus-visible/src/index.ts
```diff
@@ -46,7 +46,7 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 
 				let containsPseudo = false;
 				selectorAST.walkPseudos((pseudo) => {
-					if (pseudo.value !== PSEUDO) {
+					if (pseudo.value.toLowerCase() !== PSEUDO) {
```
Antonio-Laguna, line 49: Handling lowercase

Antonio-Laguna:
@romainmenke I bet we should add now the `.js-blank-pseudo` to PostCSS Blank Pseudo too?

plugins/postcss-focus-visible/README.md
```diff
@@ -98,14 +93,7 @@ postcssFocusVisible({ replaceWith: '[data-focus-visible-added]' })
...
 
-.foo {
-	color: blue;
-	color: red;
-}
-
-.baz {
-	color: green;
-}
+<example.preserve-true.expect.css>
```
romainmenke, line 96: I think the docs need a poke?

plugins/postcss-focus-within/src/browser.js
```diff
@@ -0,0 +1,75 @@
...
+
+				activeElement = activeElement.parentNode;
+			}
+		}
+	};
+}
+
+export default function focusWithin(opts) {
+	// configuration
+	const options = Object.assign(
```
romainmenke, line 40: `Object.assign` doesn't have good legacy support.
I will do another pass to check for API usage.

romainmenke (commented):
Awesome work!

plugins/postcss-focus-visible/README.md
Antonio-Laguna: Nice catch!

plugins/postcss-focus-within/src/browser.js
Antonio-Laguna: Thanks for flagging this. You're completely right! Thanks for putting me in the right path :)

romainmenke (approved):
LGTM!

> I bet we should add now the .js-blank-pseudo to PostCSS Blank Pseudo too?

Yeah, I think that would be best for consistency

PR #517, opened 2022-07-03
case insensitive matching

romainmenke:
same as before but for plugins which have changed a lot for `v8`


Antonio-Laguna (approved):
LGTM! Thanks!

PR #519, opened 2022-07-03
postcss-blank : browser compat

romainmenke:


plugins/css-blank-pseudo/src/browser.js
```diff
@@ -1,8 +1,14 @@
-/* global document,MutationObserver */
+/* global document,window,self,MutationObserver */
 import isValidReplacement from './is-valid-replacement.mjs';
 
 // form control elements selector
-const BLANK_CANDIDATES = 'input,select,textarea';
+function isFormControlElement(element) {
```
romainmenke, line 5: trivial equivalent for `matches()` that isn't a polyfill and thus doesn't require testing or validation

plugins/css-blank-pseudo/src/browser.js
```diff
@@ -108,39 +114,71 @@ export default function cssBlankPseudoInit(opts) {
 
 	const handler = generateHandler(options.replaceWith);
 
-	document.body.addEventListener('change', handler);
-	document.body.addEventListener('input', handler);
-
-	// observe value changes on <input>, <select>, and <textarea>
-	const window = (document.ownerDocument || document).defaultView;
+	if (document.body) {
```
romainmenke, line 117: if the activation is called in the html head `document.body` will still be `undefined`

plugins/css-blank-pseudo/src/browser.js
```diff
@@ -108,39 +114,71 @@ export default function cssBlankPseudoInit(opts) {
...
-						if (node.nodeType === 1 && node.matches(BLANK_CANDIDATES)) {
+						if (node.nodeType === 1 && isFormControlElement(node)) {
 							handler({ target: node });
 						}
 					},
 				);
 			});
 		}).observe(document, { childList: true, subtree: true });
+	} else {
+		window.addEventListener('load', () => {
```
romainmenke, line 157: If the init is called in the html head we miss the initial state.
Adding a few event callbacks to make sure the initial state is correct even if we have no support for mutation observer

rollup/presets/browser.javascript.js
```diff
@@ -43,7 +43,7 @@ export function browserJavascript() {
 					rootDir: path.join(process.cwd(), '..', '..'),
 				}),
 				babel(babelConfig),
-				terser(),
+				// terser(),
```
Antonio-Laguna, line 46: leftover?

romainmenke: yup :)

Antonio-Laguna (approved):
LGTM! Thanks for doing this!

PR #520, opened 2022-07-03
postcss-focus-within : browser compat

romainmenke:


plugins/postcss-focus-within/src/browser.js
```diff
@@ -65,7 +65,11 @@ export default function focusWithin(opts) {
 	const handleFocusChange = generateHandler(options.replaceWith);
 
 	const initialize = function initializeEventListeners() {
-		document.documentElement.classList.add('js-focus-within');
+		if (document.documentElement.className.indexOf('js-focus-within') > -1) {
```
romainmenke, line 68: If the class was already added this function was likely already called.

plugins/postcss-focus-within/src/browser.js
```diff
@@ -65,7 +65,11 @@ export default function focusWithin(opts) {
...
+			return;
+		}
+
+		document.documentElement.className = document.documentElement.className + ' js-focus-within';
```
romainmenke, line 72: Low effort fallback for `classList` which does not exist in IE9

Antonio-Laguna (approved):
Thanks! This looks great!

PR #524, opened 2022-07-07
Adding global class to CSS Blank

Antonio-Laguna:
This PR adds the `js-blank-pseudo` to the `documentElement` as well as handled via the plugin. This is now consistent across `focus-within` and `focus-visible` which are all doing the same with different pseudos.

As discussed I've also added a warning around `replaceWith` to warn about `classList` usage for old browsers.

### DIFF
--- a/plugins/css-blank-pseudo/src/browser.js
+++ b/plugins/css-blank-pseudo/src/browser.js
@@ -137,6 +137,10 @@ export default function cssBlankPseudoInit(opts) {
 		window.addEventListener('load', bindEvents);
 	}
 
+	if (document.documentElement.className.indexOf(CSS_CLASS_LOADED) === -1) {
+		document.documentElement.className += ` ${CSS_CLASS_LOADED}`;
+	}
+
 	observeValueOfHTMLElement(self.HTMLInputElement, handler);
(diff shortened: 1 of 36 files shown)


Antonio-Laguna:
@romainmenke from your browser compat PRs #519 and #520 I realised that `classList` was being used indistinctly and wanted to check this with you before spreading over to the rest of the plugins. This makes me cringe a bit with the repetitioning but I guess it's not worth the hassle of setting up shared code here.

plugins/css-blank-pseudo/src/browser.js
```diff
@@ -30,8 +33,27 @@ function generateHandler(replaceWith) {
 
 	if (replaceWith[0] === '.') {
 		selector = replaceWith.slice(1);
-		remove = (el) => el.classList.remove(selector);
```
romainmenke, line 33: 🤔 I actually do not mind `classList` here.

The default is to use attributes which works everywhere.
Users that absolutely want a class here can add their own `classList` polyfill.

romainmenke: Maybe we should explain this were we document the option to provide a class?

Antonio-Laguna: ok then!

PR #525, opened 2022-07-07
css has pseudo : pseudo element support and .js-has-pseudo class

romainmenke:
## Pseudo elements

Previous iteration encoded pseudo elements as part of the `[csstools-has-...]`
This pretty much broke them.

`:has(> .foo)::after` is now split before the pseudo element.
This ensures pseudo elements still work.

## `.js-has-pseudo`

Works a bit different from the other polyfills.
We already require access to CSSOM to analyse the stylesheet.
I am using this to rewrite the rules.

If the polyfill is loaded and applied we remove `.js-has-pseudo` from the selectors in CSSOM.

If the polyfill is loaded but there is native support we remove these rules entirely.

Overall this has the least side effects in terms of specificity and interaction with `:not()`.

### DIFF
--- a/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
+++ b/plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
@@ -269,7 +269,7 @@
 	background-color: yellow;
 }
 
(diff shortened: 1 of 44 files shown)


Antonio-Laguna (approved):
LGTM Nice work here!

PR #528, opened 2022-07-08
flip enableClientSidePolyfills

romainmenke:
Haven't added a changelog entry as I noticed we haven't added anything yet for preset-env.
Maybe best to do this in a separate PR for all changes, so that we can focus solely on that?


Antonio-Laguna (approved):
LGTM! Thanks for adding this. I agree with the CHANGELOG

Issue #540, opened 2022-07-11
Postcss-gap-properties not working anymore since 2 days on Angular 13.3.8 build


Alyyc85:
### Bug description

Without any changes on scss files we run into impossibility to serve or build some packages with Angular V 13.3.8 whic use this in webpack process

The error occours in  index.cjs file at line 10

`if (!o.parent.some((e => "display" === e.prop.toLowerCase() && "grid" === e.value.toLowerCase()))) return;`

where e.prop is undefined so it throws error
![image](https://user-images.githubusercontent.com/16958658/178329730-49c4a452-69a4-4bf0-b81d-acf63f7db8c4.png)

Here the objects collected from launching script with debugger attached:
![image](https://user-images.githubusercontent.com/16958658/178330161-db85cc0c-dd21-4143-a092-103c09dd7374.png)

### What plugin are you experiencing this issue on?

PostCSS Gap Properties

### Plugin version

3.0.4

### What OS are you experiencing this on?

Windows

### Node Version

14.18.2

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@Alyyc85 just to confirm, are you using nested CSS around gap properties?

SC-JC:
was about to report the same error, when i debugged it, both e.prop and e.value were undefined, manually adding 
`"postcss-gap-properties": "3.0.3",`
to my package.json temporarily fixes the problem.
The angular-devkit has a dependency on
`"postcss-preset-env": "7.2.3",`
which has a depdendency for
`"postcss-gap-properties": "^3.0.2",`
so will automatically install 3.0.4 when you run a fresh npm install

Antonio-Laguna:
@SC-JC and @Alyyc85  this has just been published as 3.0.5. Could you do a fresh install (or manually bump the dependency of `postcss-gap-properties`) and test? Hopefully it should be fixed :)

SC-JC:
@Antonio-Laguna it looks like 3.0.5 resolved the issue :)

Alyyc85:
@Antonio-Laguna yep with 3.0.5, downloaded in fresh npm install, it works, Thanks!

PR #541, opened 2022-07-11
postcss-gap-properties: fix detection of `display: grid`

romainmenke:
fixes: https://github.com/csstools/postcss-plugins/issues/540

### DIFF
--- a/plugins/postcss-gap-properties/src/index.js
+++ b/plugins/postcss-gap-properties/src/index.js
@@ -29,6 +29,10 @@ function creator(opts) {
 
 			const replacement = `grid-${decl.prop.toLowerCase()}`;
 			const hasGridPrefixedDeclaration = decl.parent.some((node) => {
+				if (node.type !== 'decl') {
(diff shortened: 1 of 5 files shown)


plugins/postcss-gap-properties/src/index.js
```diff
@@ -17,6 +17,10 @@ function creator(opts) {
 			}
 
 			const isNotDisplayGrid = !(decl.parent.some((node) => {
+				if (node.type !== 'decl') {
```
romainmenke, line 20: When this is a `rule` or `at-rule` `prop` and `value` won't exist

iain-s-h: With this change I'm getting a runtime error on line 32 because node.prop is undefined.

romainmenke: thank you @iain-s-h !

Antonio-Laguna (approved):
LGTM!

PR #554, opened 2022-07-22
conditional values

romainmenke:
Space toggle hacks are hard to read and write but very powerful.
Can we add some sugar for them?

```css
.fancy-container {
	--is-fancy: true;
}

.block {
	color: csstools-if(--is-fancy pink else red);
}
```

I think this is useful and can help others but feel free to shoot this down :)

-----

- I couldn't find any other plugin that does this, but maybe I missed it?
- [sass has `if`](https://sass-lang.com/documentation/modules#if)

```scss
if($condition, $if-true, $if-false)
```

vs. 

```pcss
csstools-if(--condition 1 else 2);
```

1. I prefer having `else` an explicit part because this really is a ternary and can not be used with just a single value. 
2. `csstools-if-else(--condition, 1, 2)` could work but I prefer the space separated syntax with the `else` keyword to split truthy and falsy values.

-----

The name also needs some bikeshedding.

Maybe just `@csstools/postcss-if-else`?


plugins/postcss-conditional-values/package.json
```diff
@@ -0,0 +1,75 @@
+{
+	"name": "@csstools/postcss-conditional-values",
+	"description": "TODO: Add description for Conditional Values",
```
Antonio-Laguna, line 3: Flagging this

Antonio-Laguna:
Love this! Just left a minor comment on the description just to make sure is not forgotten.

I prefer the coma-less syntax as it does seem more in line with the nature of CSS and with the changes to all color functions for example.

Can/should we add some way to pass a friendlier name so you could write `if` vs `csstools-if` ? I understand why you didn't add `if` by default but I'm wondering at the same time if we should help adoption (at the user's risk), this could be a plugin option with a huge warning

romainmenke:
> Can/should we add some way to pass a friendlier name so you could write if vs csstools-if ? I understand why you didn't add if by default but I'm wondering at the same time if we should help adoption (at the user's risk), this could be a plugin option with a huge warning

I was considering a more general purpose find/replace postcss plugin for this.
We have similar requests for `design-token()` and there are to many plugins out there that do this find/replace based on regexp instead of AST parsing.

This would give anyone the ability to create handy aliases for anything based on their preference.

plugins/postcss-conditional-values/package.json
romainmenke: Updated

Antonio-Laguna (approved):
🎉 LGTM!

plugins/postcss-conditional-values/README.md
```diff
@@ -0,0 +1,149 @@
...
+	--is-fancy: initial;
+}
+
+.block {
+	--is-fancy--0: var(--is-fancy) red;
+	color: var(--is-fancy--0,pink);
+}
+```
+
+For more information see one of the articles :
```
Antonio-Laguna, line 38: Feel free to change the suggestion felt odd that it was making you choose.

```suggestion
For more information on how the trick works, you can read more on these articles:
```

romainmenke: Good catch!

PR #559, opened 2022-07-30
Adding new option to control `design-token` function name

Antonio-Laguna:
This PR mimics the work done for `conditional-values`.

I wonder if we should be able to tweak the at-rule too. I guess no harm but wanted to check before doing!


plugins/postcss-design-tokens/src/index.ts
```diff
@@ -65,7 +65,7 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 					}
 				},
 				Declaration(decl, { result }) {
-					if (decl.value.toLowerCase().indexOf('design-token') === -1) {
+					if (decl.value.toLowerCase().includes(options.functionName)) {
```
Antonio-Laguna, line 68: Swapped by `includes` to enhance legibility

plugins/postcss-design-tokens/README.md
```diff
@@ -81,6 +81,34 @@ This allows you to generate multiple themed stylesheets.
 
 By default only `@design-tokens` without any `when('foo')` conditions are used.
 
+### functionName
+
+The `functionName` option allows you to set a custom alias for `design-token`.
```
romainmenke, line 86: Should we anticipate future functions and name this something more specific?

maybe `valueFunctionName`?

Not entirely happy with that but it is the name for the function to replace values.
A bit verbose, but covers it fully : `valueReplacementFunctionName`

romainmenke:
> I wonder if we should be able to tweak the at-rule too. I guess no harm but wanted to check before doing!

I can imagine that users will want to change this to `tokens`, `import-tokens` or something similar :) So I agree that it should be customisable.

plugins/postcss-design-tokens/README.md
Antonio-Laguna: That makes complete sense!

Antonio-Laguna:
@romainmenke added that. I'm unsure if you foresee more at-rules being created for the plugin though!

romainmenke:
> I'm unsure if you foresee more at-rules being created for the plugin though!

Hehe, I should have thought about that and mentioned it earlier :)
We will likely need to add a mixing-like `at rule` for composite tokens.

These are tokens that have multiple key/value pairs and might not always have a CSS shorthand.

I am not a fan of composite tokens and have been arguing for an alternative but you know how things have been going there :/

```pcss
{
  @token ('my.typeography.composite-token');
  color: red;
}

/* becomes : *?

{
  line-height: 2;
  font-size: 18px;
  letter-spacing: 2px;
  /* ... */
  color: red;
}
```

Antonio-Laguna:
Does `fileAtRuleName` sound good to you then? Or perhaps `tokensAtRuleName`?

romainmenke:
`fileAtRuleName` or `importAtRuleName`?

`tokensAtRuleName` might be confusing because everything is related to tokens in some way in this plugin :)

Antonio-Laguna:
@romainmenke as we discussed, I wen't with `importAtRuleName` !

plugins/postcss-design-tokens/README.md
```diff
@@ -74,13 +74,69 @@ Use `style-dictionary3` in `@design-tokens` rules to pick this format.
...
+```
+
 ### is
 
 The `is` option determines which design tokens are used.
 This allows you to generate multiple themed stylesheets.
 
 By default only `@design-tokens` without any `when('foo')` conditions are used.
 
+### valueFunctionName
```
romainmenke, line 112: `### is` should remain right before `#### Example usage`

romainmenke (commented):
One final note :)

plugins/postcss-design-tokens/README.md
Antonio-Laguna: Reason I moved is to keep them alphabetically sorted but I guess `is` is more important?

romainmenke: `### is` is the parent chapter of `#### Example usage`.
So it's `Example usage of "is"` :)

Antonio-Laguna: 🤦

Antonio-Laguna: Thanks for catching that!

romainmenke (approved):
LGTM!

Thank you for those tweaks 🎉

Antonio-Laguna:
Thank YOU for the thorough review!

PR #567, opened 2022-08-12
postcss-text-decoration-shorthand

romainmenke:
readme : https://github.com/csstools/postcss-plugins/tree/postcss-text-decoration-shorthand--diligent-ibis-84ebbbd0bb/plugins/postcss-text-decoration-shorthand#readme

[browser compat data : `css.properties.text-decoration.text-decoration-thickness`](https://github.com/mdn/browser-compat-data/blob/main/css/properties/text-decoration.json#L119)

The `text-decoration` shorthand rollout in browsers was botched in my opinion.
This leaves the feature in a largely unusable state if you want to support older browsers.

The core of the issue is that `text-decoration` became a shorthand, first for 3 and now for 4 different properties.

Doesn't work in all browsers :

```css
.broken {
  /* Safari doesn't support the unprefixed shorthand */
  text-decoration: wavy overline purple;  

  /* Safari doesn't support thickness yet in the shorthand */
  -webkit-text-decoration: wavy overline purple 4px;  

  /* only works in fairly modern browsers */
  text-decoration-line: overline;
  text-decoration-color: purple;
  text-decoration-style: wavy;
  text-decoration-thickness: 4px;

  /* we can not fallback `text-decoration-line` */
  /* in modern browsers we would reset the other constituent properties */
  text-decoration: overline; /* color, style and thickness will become their initial value */
  text-decoration-line: overline;
}
```

The work around is to write everything as shorthand explicitly and construct a multi-tiered fallback :

```css
.example {
	text-decoration: wavy underline purple 25%;
}

/* becomes */

.example {
	text-decoration: underline;
	text-decoration: underline wavy purple;
	text-decoration-thickness: calc(0.01em * 25);
	text-decoration: wavy underline purple 25%;
}
```

Autoprefixer will add fixes for `text-decoration`.
The final result being :

```css
.example {
	text-decoration: underline;
	-webkit-text-decoration: underline wavy purple;
	        text-decoration: underline wavy purple;
	text-decoration-thickness: calc(0.01em * 25);
	-webkit-text-decoration: wavy underline purple 25%;
	        text-decoration: wavy underline purple 25%;
}
```


romainmenke:
note : a separate plugin will be needed to convert percentage values in `text-decoration-thickness`

plugins/postcss-text-decoration-shorthand/README.md
```diff
@@ -0,0 +1,82 @@
+# PostCSS Text Decoration Shorthand [<img src="https://postcss.github.io/postcss/logo.svg" alt="PostCSS Logo" width="90" height="90" align="right">][postcss]
+
+[<img alt="npm version" src="https://img.shields.io/npm/v/@csstools/postcss-text-decoration-shorthand.svg" height="20">][npm-url] [<img alt="CSS Standard Status" src="https://cssdb.org/images/badges/TODO.svg" height="20">][css-url] [<img alt="Build Status" src="https://github.com/csstools/postcss-plugins/workflows/test/badge.svg" height="20">][cli-url] [<img alt="Discord" src="https://shields.io/badge/Discord-5865F2?logo=discord&logoColor=white">][discord]
```
Antonio-Laguna, line 3: Should we remove CSSDB one?

romainmenke: We should :)

Antonio-Laguna (approved):
Nice one! Very convenient for sure!

PR #568, opened 2022-08-14
postcss-individual-transform-properties

romainmenke:
https://www.w3.org/TR/css-transforms-2/#individual-transforms

Particularly tricky feature to provide fallbacks for.
This is not a shorthand/longhand situation.

- `rotate: x;` and `tranform: rotate(y);` stack.
- unknown DOM structure
- `inherit`
- `revert`, `reset`, `initial`, ...
- browser support for `var()`
- doesn't (yet?) work in `@keyframes`

I think there are too many downsides to providing a fallback and that we will break the stylesheets of authors in too many cases and browsers.

I do however have rather strict views of the browsers I want to support and spec correctness.

Maybe the tradeoffs are different for other people and we still want to complete this plugin?

- we can warn on CSS wide keywords (`revert`, `initial`, ...)
- I can get `inherit` to work but this doubles the generated CSS.

Thoughts?

--------

This plugin vs : https://github.com/csstools/postcss-transform-shortcut

The older plugin only works within a single rule.
This completely eliminates one feature of the new spec.

Because the new spec adds individual properties, stylesheet authors can use utility classes _(the sane kind)_ or modifiers to combine and stack transform effects:

```css
.base {
  transform: scale(1.1);
}

.base--upside-down {
  rotate: 180deg;
}

.base--extra-large {
  scale: 1.5;
}
```

I didn't want to give that up.


romainmenke:
Closing for now, might revisit this later.
Anyone is free to pick this up :)

PR #573, opened 2022-08-15
add cssdb id's

romainmenke:


plugins/postcss-nested-calc/test/basic.css
```diff
@@ -38,5 +38,5 @@
 }
 
 .custom-props {
-	--order: calc(1 + var(--foo, calc(2 * 2)));
```
romainmenke, line 41: This test checks that we ignore custom properties but the value was also ignored for other reasons.

Simplified the test to make sure we don't have false signals.

Issue #583, opened 2022-08-22
postcss-design-tokens issues using Style Dictionary JSON


fredboyle:
### Bug description

Maximum call stack size exceeded [postcss-design-tokens] error encountered with the following JSON imported:

```json
{
	"color": {
		"internal": {
			"neutrals": {
				"darkest": "#1E1E1E",
				"lightest": "#FFFFFF"
			}
		}
	}
}
```

<details>
<summary>Expand for the full tokens source</summary>

```json
{
	"color": {
		"brand": {
			"primary": {
				"value": "#ffdc28",
				"filePath": "tokens/color/brand.json",
				"isSource": true,
				"original": {
					"value": "#FFDC28"
				},
				"name": "color-brand-primary",
				"attributes": {
					"category": "color",
					"type": "brand",
					"item": "primary"
				},
				"path": ["color", "brand", "primary"]
			},
			"admin": {
				"value": "#825077",
				"filePath": "tokens/color/brand.json",
				"isSource": true,
				"original": {
					"value": "#825077"
				},
				"name": "color-brand-admin",
				"attributes": {
					"category": "color",
					"type": "brand",
					"item": "admin"
				},
				"path": ["color", "brand", "admin"]
			}
		},
		"marketing": {
			"gradient": {
				"sunrise": {
					"value": "linear-gradient(227.79deg, #300068 38.28%, #E65F35 100%)",
					"filePath": "tokens/color/gradients-marketing.json",
					"isSource": true,
					"original": {
						"value": {
							"type": "linear-gradient",
							"angle": "227.79deg",
							"colors": {
								"1": "#300068 38.28%",
								"2": "#E65F35 100%"
							}
						}
					},
					"name": "color-marketing-gradient-sunrise",
					"attributes": {
						"category": "color",
						"type": "marketing",
						"item": "gradient",
						"subitem": "sunrise"
					},
					"path": ["color", "marketing", "gradient", "sunrise"]
				},
				"sunset": {
					"value": "linear-gradient(48.24deg, #EF13AD -3.39%, #300068 66.25%)",
					"filePath": "tokens/color/gradients-marketing.json",
					"isSource": true,
					"original": {
						"value": {
							"type": "linear-gradient",
							"angle": "48.24deg",
							"colors": {
								"1": "#EF13AD -3.39%",
								"2": "#300068 66.25%"
							}
						}
					},
					"name": "color-marketing-gradient-sunset",
					"attributes": {
						"category": "color",
[...]

PR #584, opened 2022-08-23
postcss-design-tokens : prevent stack overflow

romainmenke:
see : #583

We can not really fix the issue because the source file doesn't have the expected format, but we can emit a more meaningful warning.


plugins/postcss-design-tokens/src/data-formats/style-dictionary/v3/group.ts
```diff
@@ -7,12 +7,20 @@ export type StyleDictionaryV3TokenGroup = {
...
+				Array.isArray(node[key]) &&
+				node[key] === null
+			) {
+				throw new Error(`Parsing error at "${[...path, key].join('.')}"`);
+			}
+
 			const child = Object(node[key]);
 			if (!child) {
-				throw new Error('Parsing error');
+				throw new Error(`Parsing error at "${[...path, key].join('.')}"`);
```
romainmenke, line 23: @Antonio-Laguna we can also choose to ignore these issues.
This will give users more freedom.

At the moment we require that each object has a `"value"` property.
If this is missing we emit a `Parsing error`.

Thoughts?

Antonio-Laguna: What would be an usecase for not having a `value` property? If there's none, I think we should just error since you're probably missing something?

romainmenke: Users might stuff extra meta data into the same JSON file.
Since there is no specification this is "allowed".

Starting with an error makes more sense to me.
If users need to this to be more forgiving they can ask us to change it or make it configurable.

Antonio-Laguna: Yep, agreed

Antonio-Laguna (approved):
LGTM!

PR #586, opened 2022-08-23
browser polyfills : use window or self for the global polyfill

romainmenke:


Antonio-Laguna (approved):
Thanks for the swift update! LGTM

PR #591, opened 2022-08-29
design-tokens : implement `require.resolve`

romainmenke:


Antonio-Laguna:
This is looking good too! Let me know if you want this together for the next minor or if you'd prefer to have two minors

PR #596, opened 2022-08-31
cascade layers fixes

romainmenke:
- fixes an issue with empty duplicate `at-rules`.
- added support for `@container` queries by adding it the list of known conditional rules.
- update `postcss-import` to `15.0.0` : https://github.com/postcss/postcss-import/blob/master/CHANGELOG.md#1500--2022-08-30

### DIFF
--- a/plugins/postcss-cascade-layers/src/constants.ts
+++ b/plugins/postcss-cascade-layers/src/constants.ts
@@ -9,6 +9,7 @@ export const IMPLICIT_LAYER_SUFFIX = 'b147acf6-11a6-4338-a4d0-80aef4cd1a2f';
 export const CONDITIONAL_ATRULES = [
 	'media',
 	'supports',
+	'container',
 ];
 
 export const ATRULES_WITH_NON_SELECTOR_BLOCK_LISTS = [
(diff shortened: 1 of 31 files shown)


plugins/postcss-cascade-layers/src/constants.ts
```diff
@@ -9,6 +9,7 @@ export const IMPLICIT_LAYER_SUFFIX = 'b147acf6-11a6-4338-a4d0-80aef4cd1a2f';
 export const CONDITIONAL_ATRULES = [
 	'media',
 	'supports',
+	'container',
```
romainmenke, line 12: With `@container` shipping in browsers we need to make sure we handle it correctly.

plugins/postcss-cascade-layers/src/sort-root-nodes.ts
```diff
@@ -70,22 +71,6 @@ export function sortRootNodes(root: Container, model: Model) {
...
-			return aIsCharset ? -1 : 1;
-		}
-
-		const aIsImport = a.type === 'atrule' && a.name.toLowerCase() === 'import';
-		const bIsImport = b.type === 'atrule' && b.name.toLowerCase() === 'import';
-		if (aIsImport && bIsImport) {
-			return 0;
-		} else if (aIsImport !== bIsImport) {
-			return aIsImport ? -1 : 1;
-		}
```
romainmenke, line 87: All this code was not needed.
If users have incorrect orders for `@charset` or `@import` in their source we must leave it as it was, not try to fix it.

plugins/postcss-cascade-layers/test/nested-case-insensitive.expect.css
```diff
@@ -9,9 +9,6 @@
 		}
 	}
 
-@KEYFRAMES slide-left {
-	}
-
```
romainmenke, line 14: This was the bug I wanted to fix.
I totally missed this but an empty `@KEYFRAMES` was created before this update.

This obviously breaks a lot of things.

Antonio-Laguna (approved):
This is looking great, thanks a lot for the extended explanation and for chasing this bug down!

PR #599, opened 2022-08-31
bump PostCSS, node and npm versions

romainmenke:
Minimum `PostCSS` is now `8.4`.
For the moment this is equal to the highest version we test in CI, but this can grow apart over time again.

Minimum `node` version is now `14`.

`npm` version change only affects contributors using `volta` to manage their local dev env.


Antonio-Laguna (approved):
Thanks for doing all of these chores! Looking great

Issue #595, opened 2022-08-30
[postcss-custom-properties] issue when used with postcss-html (document node)


lubomirblazekcz:
### Bug description

Custom properties are not transformed correctly with `postcss-html`, this might be problem with document node support as mentioned in https://github.com/ota-meshi/postcss-html/issues/68

### Source CSS

```css
 :root {
    --blue: blue
  }
  
  body td {
      color: var(--blue)
  }
  
  @media all and (max-width: 600px) {
      body {
          color: red;
      }
  }
```

### Expected CSS

```css
:root {
    --blue: blue
  }
  
  body td {
      color: blue
  ;
      color: var(--blue)
  }
  
  @media all and (max-width: 600px) {
      body {
          color: red;
      }
  }
```

### Actual CSS

```css
 :root {
    --blue: blue
  }
  
  body td {
      color: var(--blue)
  }
  
  @media all and (max-width: 600px) {
      body {
          color: red;
      }
  }
```

### What plugin are you experiencing this issue on?

PostCSS Custom Properties

### Plugin version

12.1.8

### What OS are you experiencing this on?

macOS

### Node Version

18.6.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@lubomirblazekcz Can you provide a full reproduction?

The CSS on its own works fine : 

```css
 :root {
    --blue: blue
  }
  
  body td {
      color: var(--blue)
  }
  
  @media all and (max-width: 600px) {
      body {
          color: red;
      }
  }
```

becomes :

```css
 :root {
    --blue: blue
  }
  
  body td {
      color: blue
  ;
      color: var(--blue)
  }
  
  @media all and (max-width: 600px) {
      body {
          color: red;
      }
  }
```

lubomirblazekcz:
It's mentioned in https://github.com/ota-meshi/postcss-html/issues/68, with that html you get the non-working result. 

@ota-meshi mentioned that it's actually a problem with document nodes, which postcss-custom-properties might not support. I don't know if that is something that can be fixed though.

romainmenke:
I have no idea how to add support for document nodes (or what they are) :/
I will need to look into that.

------------------

[...]

PR #602, opened 2022-09-03
add support for `postcss-html` in `postcss-tape`

romainmenke:
[postcss-html](https://www.npmjs.com/package/postcss-html)

see : https://github.com/csstools/postcss-plugins/issues/595

------

To add support for the new `Document` event and node type we need to be able to test this with `postcss-tape`.

The `postcss-html` syntax plugin combined with `html` files as test sources allow us to easily test plugins for compat issues.


packages/postcss-tape/test/_a-plugin.mjs
```diff
@@ -4,25 +4,12 @@ const aPlugin = (options) => {
...
-				// Check the release logs of PostCSS and pick something else that is new to use.
-				// Disabling this test is also fine.
-				try {
-					decl.assign({ prop: 'order-b', value: '2' });
-				} catch (error) {
-					// strip stacktrace to ensure the stderr output is the same across multiple machines
-					console.error('[TAPE DEBUG] ' + error.message);
-				}
-			}
-
```
romainmenke, line 25: 8.4 is the minimum version now

Antonio-Laguna (approved):
LGTM!

PR #603, opened 2022-09-03
refactor `postcss-custom-properties` and cleanup in `postcss-preset-env`

romainmenke:
out take from https://github.com/csstools/postcss-plugins/pull/527

- [x] remove all the logic in `postcss-preset-env` for `importFrom`, `exportTo` and "side effects"
- [x] update docs for `postcss-preset-env`
- [x] refactor `postcss-custom-properties`
  - remove `importFrom`, `exportTo`
  - add support for local custom property declarations
  - correctly handle `initial` as always invalid (ignore it)
  - update docs
  - update change log


plugins/postcss-custom-properties/docs/README.md
```diff
@@ -0,0 +1,62 @@
...
+<example.css>
+
+/* becomes */
+
+<example.expect.css>
+```
+
+**Note:** This plugin only processes variables that are defined in the `:root` or `html` selector.
+
+Locally defined custom properties will be used as fallbacks inly within the same rule, but not elsewhere.
```
Antonio-Laguna, line 34: Minor typo here

```suggestion
Locally defined custom properties will be used as fallbacks only within the same rule, but not elsewhere.
```

Antonio-Laguna (approved):
So much cleaner! Lots of removal here which feels great!

Looking great! Just left a minor comment but good to go!

plugins/postcss-custom-properties/docs/README.md
romainmenke: Good catch! thank you 🙇

Issue #593, opened 2022-08-29
Add import and usage tests for Typescript


romainmenke:
See : https://github.com/postcss/postcss/issues/1771

We have typings for more and more plugins but we don't have any actual usage tests with Typescript.

At the moment it is possible that plugins we author do not actually work for Typescript users.

PR #606, opened 2022-09-05
add e2e tests for typescript

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/593

- I think dependabot will now also manage dependencies in the `e2e` folder
- setup node workspaces for `e2e` -> `install`, `build` and `test` in `e2e` can now be run from the `e2e` root
- Add tests for typescript, we only test one package, if one works, they should all work.


romainmenke:
dependabot indeed picks up now :

Screenshot 2022-09-05 at 14 53 37

PR #612, opened 2022-09-07
cascade-layers : add support for `scope` and `container` at rules

romainmenke:
I think this is the right way to add support for `@scope` and `@container`.
I initially thought it would be complicated to resolve the issue and didn't immediately attempt a fix. But it seems that just adding it them the list of `allowedParentAtRules` was enough.

See : https://github.com/csstools/postcss-plugins/pull/610

These two at rules no longer affect the layer order.

_very low prio as very few people use `@layer` and even less will be using it with `@container` or `@scope`_


Antonio-Laguna (approved):
LGTM!

Thanks for adding this!

Issue #619, opened 2022-09-13
Fallback for css variables doesn't seem to be replaced correctly in combination with logical properties


benjamingries:
Hi,

first of all: this is a great plugin and I appreciate that you are spending your (free) time to help all of us. 

I don't know whether it's a bug or not but I think I found a combination where the fallback for css variables isn't replaced correctly (I'm using webpack 5.74.0 with the dependencies postcss 8.4.16, postcss-loader 7.0.1 and postcss-preset-env 7.8.1).

In the past I wrote a (scss) definition like this:

```
.stage__container {
...
left: var(--size, $spacer);
...
}
```

This works as expected.

Now I am trying to use logical properties in order to use modern css and prevent duplicate code (in the scss file) for the arabic landingpage:

```
.stage__container {
  ...
  inset-inline-start: var(--size, $spacer);
  ...
}
```

But this doesn't seem to work because in the end the fallback is rendered twice - once with the var style and its fallback and once with only the fallback (please take a look at the order):

![image](https://user-images.githubusercontent.com/25054095/189829729-874f7184-b688-4d07-98b8-293d921de6eb.png)

The same behavior occurs when I'm setting the fallback as a root variable instead of the custom property:

```
:root {
  --size: 1rem;
}

.stage__container {
  ...
  inset-inline-start: var(--size);
  ...
}
```

Setting the preserve option doesn't change this behavior for me.

The only way that works for me is this one:

```
.stage__container {
  ...
  inset-inline-start: var(--size);
  ...
}
```

But then there is no fallback for older browsers (we build a global website that is accessed by many older browsers following the usage stats).

I found out that the logical properties plugin seems to handle these (incorrect) replacements and that disabling it prevents the duplicate code. But then again there is no fallback for older browsers.

Can you recreate this behavior?

Best regards,
Benjamin

benjamingries:
For better understand: This is the part of my webpack configuration that handles the postcss replacements:

webpack.config.js

![image](https://user-images.githubusercontent.com/25054095/189833397-afa2d916-12b9-4705-92c6-a90865d5d584.png)

postcss.config.js

![image](https://user-images.githubusercontent.com/25054095/189833483-e4a913d4-dae9-46ac-9ae5-cac27809e890.png)

romainmenke:
Definitely a bug, or more precisely a whole class of bugs.

- plugin creates fallback rules (`.foo {}`)
- plugin looks at declarations (`left: 20px`)
[...]

PR #620, opened 2022-09-13
postcss-custom-properties : reduce duplicate code generation

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/619#issuecomment-1244996380

_This does not resolve the linked issue but removes a part of it and makes the real issue more visible._


Antonio-Laguna (approved):
This looks good too. Will release ASAP

Issue #624, opened 2022-09-16
Check if pre-defined orders for sub layers work.


romainmenke:
https://css-tricks.com/css-cascade-layers/#aa-grouping-nested-layers

https://twitter.com/nileshpraj/status/1570828201194029056

```css
@layer one {
  /* sorting the sub-layers */
  @layer two, three;

  /* styles ... */
  @layer three { /* styles ... */ }
  @layer two { /* styles ... */ }
}
```

nileshprajapati:
Thanks for opening the issue,  the following example above works fine.

 However, I was hoping to get the following result if possible.

@layer one {
    /* sorting the sub-layers */
    @layer two, three;
  
    /* styles */
    @import "two" layer(one);
    @import "three" layer(one);
}
    or 

@layer one {
    /* sorting the sub-layers */
    @layer two, three;

    /* styles */
    @layer  two {
       @import "two-one";
        @import "two-two";
    }

    @layer  three {
        @import "three-one";
        @import "three-two";
    }
}

Become:

@layer one {
    /* sorting the sub-layers */
    @layer two, three;
  
    /* styles */
   @layer three { /* styles ... */ }
    @layer two { /* styles ... */ }
}

romainmenke:
`@import` will never work nested in `@layer`.

You need to use this syntax : `@import ... layer(one.two)`

romainmenke:
_(didn't have much time for a longer answer before)_

```css
@layer one {
    /* sorting the sub-layers */
    @layer two, three;
  
    /* styles */
    @import "two" layer(one);
    @import "three" layer(one);
}
```

This is not supported by `postcss-import` or `postcss-cascade-layers` because it is not allowed by the CSS specification.

https://developer.mozilla.org/en-US/docs/Web/CSS/@import#description

It seems that the MDN docs are not entirely up to date with `@layer` but they do mention the important point. `@import` must be declared early and it can not be placed in other rules.

allowed :

```css
@charset ...;

@layer foo, bar;

@import ...;

/* rest of the styles */
```

⚠️ not allowed :

```css
@media screen {
  @import ...; /* not allowed in `@media` */
}

@import ...; /* not allowed after `@media` */

@layer foo {
  @import ...; /* not allowed in `@layer` */
}
```

----------------

That said, there was a real bug with the example from the article.
This update will resolve this issue : https://github.com/csstools/postcss-plugins/pull/625

Antonio-Laguna:
[...]

PR #625, opened 2022-09-16
correctly clean empty rules

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/624

In `postcss-cascade-layers` we have a couple of de-sugaring and normalisation steps.
One of these steps re-structures rules and removes empty left-over rules after the step.

`@layer foo;` is technically an empty rule in the PostCSS AST.
But there is a difference between `typeof rule.nodes === 'undefined'` and `rule.nodes.length  === 0`.

I've updated the code to preserve at-rules when `typeof rule.nodes === 'undefined'`.

```css
@layer one { /* has nodes */
	@layer two, three; /* nodes is undefined */
}

@layer four {} /* has 0 nodes */
```

The final result is that all `@layer` rules still exist when the layer order is recorded.

### DIFF
--- a/plugins/postcss-cascade-layers/.tape.mjs
+++ b/plugins/postcss-cascade-layers/.tape.mjs
@@ -36,6 +36,9 @@ postcssTape(plugin)({
 	'extensions': {
 		message: "css custom extensions",
 	},
+	'pre-defined-order-for-nested-layer': {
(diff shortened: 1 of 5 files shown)


Antonio-Laguna (approved):
LGTM!

PR #627, opened 2022-09-24
postcss-custom-media & css-tokenizer & media query parser

romainmenke:
- [x] a CSS tokenizer + a Media Query List parser so that we can manipulate `@custom-media` without `regexp`
  - maybe we need more test coverage here but for now it appears to work correctly for `@custom-media` and `@media` params. We can increase coverage as we increase usage.
- [x] converted to Typescript
- [x] updated docs
- [x] removed `importFrom`|`exportTo`
- [x] implement `true`|`false` keywords
- [x] implement logical evaluation of `@custom-media`

`postcss-preset-env` tests didn't need any changes because the plugin still works exactly the same for simple examples.


packages/css-tokenizer/LICENSE.md
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
romainmenke, line 1: MIT Licensed because it is intended as a low level package.
This matches what PostCSS uses and should be more compatible with the ecosystem.

plugins/postcss-custom-media/src/transform-at-media.ts
```diff
@@ -0,0 +1,484 @@
+import { tokenizer, stringify, TokenType, NumberType, TokenIdent } from '@csstools/css-tokenizer';
+import type { CSSToken } from '@csstools/css-tokenizer';
```
romainmenke, line 2: For now I decided not to implement a full parser for `@media`.
This is a lot of work and the patterns we need to detect and mutate are simple.

This does move a lot of complexity into this plugin that would otherwise exist in a parser:
- split media query lists by `,`
- keep track of bracket nesting depth. (`([{(( deeply nested ))}])`)

But it allows us to move a bit more quickly now and revisit this later to optimise.

plugins/postcss-custom-media/src/transform-at-media.ts
```diff
@@ -0,0 +1,484 @@
...
+		tokens.push(t.nextToken());
+	}
+
+	return tokens;
+}
+
+const alwaysTrue: Array<CSSToken> = [
+	[TokenType.Ident, 'max-color', 0, 0, { value: 'max-color' }],
+	[TokenType.Colon, ':', 0, 0, undefined],
+	[TokenType.Number, '2147477350', 0, 0, { value: 9999943, type: NumberType.Integer }],
```
romainmenke, line 24: This is an arbitrarily high number.

.github/workflows/test.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         # - without requiring postinstall scripts from dependencies to run
       - name: build
         run: |
-          npm run build --if-present $MODIFIED_WORKSPACES
+          npm run build --if-present
```
Antonio-Laguna, line 84: Is this intentional or was it just an in-between change while building?

romainmenke: Something changed that affects how the build results are cached and restored.
Those caches speed up the build process, but now they just cause it to fail often.

I thought I could eliminate the failures by always building everything and still get a small speed bump from only testing that which was changed. But it didn't work.

Currently thinking of removing all caches just to get rid of the failures.
Builds will be slower again, but they will pass reliably.

package.json
```diff
@@ -9,6 +9,9 @@
 		"node": "^14 || ^16 || >=18"
 	},
 	"workspaces": [
+		"packages/css-tokenizer",
+		"packages/css-parser-algorithms",
+		"packages/media-query-list-parser",
 		"packages/*",
```
Antonio-Laguna, line 15: Given L15, do we need to specify in particular? I notice we have progressive too so I wasn't sure.

.github/workflows/test.yml
Antonio-Laguna: Makes sense!

package.json
romainmenke: This is a "hack" to specify the build and test order.
npm workspaces do not build a dependency graph between local packages.

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
```diff
@@ -0,0 +1,621 @@
...
+	value: Array<ComponentValue>;
+
+	constructor(name: TokenFunction, endToken: CSSToken, value: Array<ComponentValue>) {
+		this.name = name;
+		this.endToken = endToken;
+		this.value = value;
+	}
+
+	get nameTokenValue(): string {
+		return this.name[4].value;
```
Antonio-Laguna, line 68: `4` here seems like a magic number, should it be abstracted into its own constant?

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
```diff
@@ -0,0 +1,621 @@
...
+
+		if (aborted) {
+			return false;
+		}
+	}
+
+	toJSON() {
+		return {
+			type: this.type,
+			name: this.name[4].value,
```
Antonio-Laguna, line 139: Perhaps we could/should reuse `nameTokenValue` getter here?

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
```diff
@@ -0,0 +1,621 @@
...
+	// eslint-disable-next-line no-constant-condition
+	while (true) {
+		const token = tokens[i];
+		if (!token || token[0] === TokenType.EOF) {
+			ctx.onParseError({
+				message: 'Unexpected EOF while consuming a function.',
+				start: tokens[0][2],
+				end: tokens[tokens.length - 1][3],
+				state: [
+					'5.4.9. Consume a function',
```
Antonio-Laguna, line 177: ❤️ how this is searchable in the spec!

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
```diff
@@ -0,0 +1,621 @@
...
+	let i = 1;
+
+	// eslint-disable-next-line no-constant-condition
+	while (true) {
+		const token = tokens[i];
+		if (!token || token[0] === TokenType.EOF) {
+			ctx.onParseError({
+				message: 'Unexpected EOF while consuming a simple block.',
+				start: tokens[0][2],
+				end: tokens[tokens.length - 1][3],
```
Antonio-Laguna, line 330: We get here more of these numbers too which tie into how these are stored. Flagging in case you want to tackle with the `4` too.

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
romainmenke: yes, and `nameTokenValue` should not be a getter.
I think it is the only getter in all the three new packages.

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
romainmenke: Is indeed a bit of a magic number, but best to keep it as is for now.
I've tried to keep the tokenizer API really small so we have a bit more freedom later.
I also hope that using plain numbers is a bit faster, but this might be wishful thinking :)

All tokens are arrays, but are also fully typed. So by narrowing the type of each token to it's specific token type you can see which index holds which value/type.

0. token type
1. representation (what was found in the source)
2. start index
3. end index
4. value (what was parsed)

- representation can contain escape sequences, value will not
- value is structured and has things like `unit`

Screenshot 2022-11-06 at 16 50 09

I do a agree that this is a bad API for plugin creators unless they also have a lot of knowledge about CSS tokenizing.

Maybe something like this is easier for plugin creators to use :

```ts
const myToken = [TokenType.Ident, 'foo', 0, 2, { value: 'foo' }];
const tokenType = typeFromToken(myToken); // TokenType.Ident
const tokenValue = valueFromToken(myToken); // { value: 'foo' }
```

```ts
const myToken = [TokenType.Dimension, '10px', 0, 3, { value: 10, unit: 'px' }];
const tokenType = typeFromToken(myToken); // TokenType.Dimension
const tokenValue = valueFromToken(myToken); // { value: 10, unit: 'px' }
```

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
romainmenke: updated

packages/css-parser-algorithms/src/consume/consume-component-block-function.ts
Antonio-Laguna: Thanks for the explanation!

Antonio-Laguna (approved):
@romainmenke this has been a lengthy review but I've learned so much about this and the spec. 

I wouldn't be able to really phrase how amazed I am by this and impressed by this work!

Thanks a lot!

My only point (and not needed here) is to add more literature about the changes in custom-media. I'm unsure if it belongs to the README or the wiki (leaning towards the README) but explaining what's new. Essentially these:

- Fixed: implement logical evaluation of custom media queries.
- Added: Support for `true` and `false` keywords in `@custom-media`.

Issue #614, opened 2022-09-11
Adopt a "do not cause harm" policy for "future css" plugins


romainmenke:
The more time I spend here and try to align plugins with their respective specifications the more I notice the harm we can and do cause by moving to fast.

I really want to avoid that we continue to poison the well on proposed features.

Adopting a solid policy and guidelines might help.

see : https://github.com/w3c/csswg-drafts/issues/7150

romainmenke:
Some aspects and thoughts :

- moving fast and making "future css" available to stylesheet authors is what made `postcss-preset-env` popular.
- providing polyfills is a completely different problem than providing "future css". Mixing these is a mistake imho.
- "vendor prefixing" features might help (`@csstools-custom-media`)

romainmenke:
rough proposal : 

- we will not add a plugin to `postcss-preset-env` before CSS WG invites implementations
- we will not create plugins with features that look and feel native but don't have a spec (jquery `:has()` issue)

PR #631, opened 2022-10-01
add authoring guidelines

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/614

This formalises core principles which we have adopted in the past few months.

### DIFF
--- a/AUTHORING_GUIDELINES.md
+++ b/AUTHORING_GUIDELINES.md
@@ -1,0 +1,14 @@
+# Authoring guidelines
+
+We want to make it easier to write awesome CSS that works great in as many browsers as possible.
+Some of these browsers are a bit older, others don't yet exist.
+
+We will not place undue burden on anyone writing or maintaining CSS.
+
+We will only add a plugin to `postcss-preset-env` when the CSS WG invites implementations for the relevant features.
+
+We will not create plugins with features that look and feel native but don't have a specification.
+
+We will make it clear that non-standard, pre-processor features are just that.
+
+We will only implement that which we know is correct and avoid side effects.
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -26,7 +26,7 @@ instructions:
 ### Quick start
 
 1. fork the repo and clone it locally
-2. `npm run get-me-going` to install and build dependencies
+2. `npm run setup-workspace` to install and build dependencies
```
romainmenke, line 29: This wording is a bit more clear I think

AUTHORING_GUIDELINES.md
```diff
@@ -0,0 +1,14 @@
+# Authoring guidelines
+
+We want to make it easier to write awesome CSS that works great in as many browsers as possible.
+Some of these browsers are a bit older, others don't yet exist.
+
+We will not place undue burden on anyone writing or maintaining CSS.
+
+We will only add a plugin to `postcss-preset-env` when the CSS WG invites implementations for the relevant features.
+
+We will not create plugins with features that look and feel native but don't have a specification.
```
Antonio-Laguna, line 10:
```suggestion
We will not create plugins with features (or add features to plugins) that look and feel native but don't have a specification.
```

Feel free to tweak the wording, an example would be #671

CONTRIBUTING.md
Antonio-Laguna: Agreed, naming things is hard ™

Thanks for coming up with this!

Antonio-Laguna (approved):
This LGTM! Feel free to discard my comment if you think it's already covered. I just wanted to be extra explicit about features too.

Thanks for adding this!

AUTHORING_GUIDELINES.md
romainmenke: I agree, good catch :)

It is something we have seen in the past.
That small productivity features find their way into plugins that intended to "polyfill" native features.

Best to keep those separate.

PR #647, opened 2022-10-15
experimental : postcss-custom-shorthands

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/7879

------

Notes :

1) This plugin has an internal versioning mechanic because it allows people to rapidly try out a different variant/flavour. This doesn't really have anything to do with the semver version of the npm package.

> The `experimentalVersion` option determines which proposal you are using.
> The latest proposal is mostly the default. (currently `2`)
> 
> - 1 : `<link to issue>`
> - 2 : `<link to issue>`
> 
> ```js
> postcssCustomShorthandsExperimental({ experimentalVersion: 1 })
> ```

2) This plugin introduces `css-tree` as a dependency.
I am fine with having it as a workspace dependency but if this feature ever becomes a real thing I prefer writing our own syntax definition parser from scratch. I don't like the bloat of having all of `css-tree` for this single parser. see : https://www.w3.org/TR/css-properties-values-api-1/#parsing-syntax

3) This package was quickly written and lacks sufficient test coverage and coding style is not ideal in places. This is fine for now I think, as we will end up trashing it later. Better to start from scratch if we ever build this for real.

4) Custom data in VSCode has this effect :

Screenshot 2022-10-15 at 11 52 44
Screenshot 2022-10-15 at 11 57 40

A real plugin would not have this feature as we would rely on editors to do this work themselves. But it is useful to get a glimpse of the real author experience.


rollup/configs/babel-presets.js
```diff
@@ -1,9 +1,8 @@
 export const packageBabelPreset = [
 	['@babel/preset-env', {
-		corejs: 3,
```
romainmenke, line 3: I had already fixed this in the `v8` branch.
We don't have `corejs` as a dependency.
This worked because we only wrote JS that didn't need any polyfills from `corejs`.

`css-tree` might require a polyfill for typed arrays. (these matches can have false positives)

Removing `corejs` from our config is the right fix as we never actually had it in the first place.

romainmenke:
This is likely unimplementable : https://github.com/w3c/csswg-drafts/issues/7879#issuecomment-1279825086

rollup/configs/babel-presets.js
Antonio-Laguna: Agreed!

experimental/postcss-custom-shorthands/package.json
```diff
@@ -0,0 +1,77 @@
+{
+	"name": "@csstools/postcss-custom-shorthands-experimental",
+	"description": "Research plugin for a Custom Shorthands proposal",
+	"version": "1.0.0",
```
Antonio-Laguna, line 4: Should this be in the `0.0.x` realm?

Antonio-Laguna:
So I like this a lot! Love the versions too 

I'm wondering a couple of things

- Should this be tested with custom-properties? Could there be any clashes?
- Could/should there be an easy way to leverage the `default`. I've been looking at this and I sort of would expect that no value leads to `default` however I can also see how this could be problematic

romainmenke:
Closing because it can not be implemented.

Thank your for reviewing @Antonio-Laguna 🙇

PR #662, opened 2022-10-20
custom-properties : fix how preserve false interacts with duplicate code

romainmenke:


plugins/postcss-custom-properties/test/basic.import-override.expect.css
```diff
@@ -112,12 +112,10 @@
 
 .ignores-declarations-that-have-an-exact-fallback-a {
 	left: 1rem;
-	left: var(--does-not-exist, 1rem);
```
romainmenke: We can not preserve `var()` because `left: 1rem;` will never be used by browsers that understand `var()`;

Because we strip the `:root` of all custom props we also have to remove these.

Antonio-Laguna (approved):
@romainmenke this looks good to me, let's get this out!

PR #670, opened 2022-10-26
Feat/rgba css variables

dufemeng:
Add a new plugins.


romainmenke:
Hi @dufemeng 👋 

Thank you for contributing!

Can you tell us a little bit more about this plugin and why we should it add it here?

Have you also seen this plugin : https://www.npmjs.com/package/postcss-split-color-vars
It seems to do similar things.

dufemeng:
> 

Hi @romainmenke ,

I read `postcss-split-color-vars` carefully just a moment ago. Then here is  `postcss-split-color-vars`'s example:
```pcss
:root {
  @split-colors {
    --color-primary: rgb(14, 14, 14);
    --color-foo: rgb(128 0 0 / 80%);
  }
}
````

It seems like the plugin support specific rgb number but not css variables. In contrast to the `rgba css variables ` plugin can support it such as:

```pcss
:root {
  @split-colors {
    --color-primary: rgb(14, 14, 14);
    --color-foo: rgba(var(--color-primary), .8);
  }
}
````

And this plugin provide three output mode for selection.

Looking forward to your reply.

romainmenke:
@dufemeng I am closing this pull request because we do not think this plugin fits within our scope. As I said before this isn't in any way a judgement about the usefulness or the quality of your work.

We do hope that you finish and publish it so that you and others can use it!

PR #671, opened 2022-10-26
Allow custom media queries to be used in @container rules

gabrielgrant:
Container queries have a syntax nearly identical to media queries, and can replace them in many cases, while adding extra functionality. But as of right now, the custom-media plugin doesn't allow using the defined media queries in `@container` rules.  

This PR adds causes the plugin to process `@container` at-rules in the same way `@media` at-rules are handled

FWIW, container queries are now [natively supported in Blink and WebKit](https://caniuse.com/css-container-queries), and supported via [polyfill](https://github.com/GoogleChromeLabs/container-query-polyfill) in Gecko/Firefox

### DIFF
--- a/plugins/postcss-custom-media/src/index.js
+++ b/plugins/postcss-custom-media/src/index.js
@@ -32,7 +32,7 @@ const creator = opts => {
 			await writeCustomMediaToExports(helpers[customMediaHelperKey], exportTo);
 		},
 		AtRule: (atrule, helpers) => {
-			if (atrule.name !== 'media') {
+			if (!(atrule.name === 'media' || atrule.name === 'container')) {
 				return;
 			}
 
--- a/plugins/postcss-custom-media/test/basic.css
+++ b/plugins/postcss-custom-media/test/basic.css
@@ -8,6 +8,12 @@
 	}
 }
 
+@container (--mq-a) {
+	body {
+		order: 1;
+	}
+}
+
 @media (--mq-b) {
 	body {
 		order: 1;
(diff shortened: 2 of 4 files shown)


Antonio-Laguna:
@gabrielgrant thanks for this PR. However we won't be integrating this within `postcss-custom-media`.

The reason is that that's not part of any spec and it hasn't even been drafted. For whatever is worth, none of the proposals related to custom-media or anything in that line is getting any traction with browsers nor CSSWG so we're hesitant to add tooling without, at least, a draft spec.

This is the closest to what you're suggesting but it didn't get picked either (for now) https://github.com/w3c/csswg-drafts/issues/7622

gabrielgrant:
thanks for taking a look. from the discussion in https://github.com/w3c/csswg-drafts/issues/6698 it sounds like the plan is for the functionality of `env()` to be pretty much a superset of the functionality of custom media (and so seems likely that will move forward instead). Would it make sense to mention that on the [custom-css README](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-custom-media/README.md) that using the [env function plugin](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-env-function) may be a better choice? (I certainly would have appreciated a nudge in that direction)

romainmenke:
Side note :

We are extra careful with additions and new plugins and try to align them very closely to native browser behavior.

But you should not feel limited by this.
You are absolutely free to take the code for `@custom-media` and make it work for `@container` in any way or form you like :)

gabrielgrant:
@romainmenke thanks so much for the follow-up!

PR #674, opened 2022-10-29
postcss nesting experimental

romainmenke:
see : https://drafts.csswg.org/css-nesting/

- no more `@nest`
- relative selector syntax
- everything uses `:is()` now

----

I don't want the complexity of trying to avoid `:is()` at this point.
Using `:is()` makes it easier to avoid certain types of bugs in our code.


Antonio-Laguna (approved):
This is looking great! Thanks for staying ahead of the curve! Going to merge and release this ASAP

PR #693, opened 2022-11-08
update rollup

romainmenke:


romainmenke:
This fails on `importFrom`/`exportTo` on windows.
Since we are removing those feature I am not going to fix that.
Instead I will migrate to rollup v3 on the `postcss-preset-env v8` branch.

Issue #673, opened 2022-10-28
invalid selectors and CSS that is interpreted as an invalid selector throws exceptions


devingfx:
### Bug description

The W3C spec says [@media rules can be nested](https://www.w3.org/TR/css-nesting-1/#conditionals) but this plugin craches when it find a nested @media

### Actual Behavior

```css
.my {
    @media print {
        color: red;
    }
}
```

`Error: Expected a pseudo-class or pseudo-element.`

### Expected Behavior

```css
@media print {
    .my {
        color: red;
    }
}
```

### What plugin are you experiencing this issue on?

PostCSS Nesting

### Plugin version

10

### What OS are you experiencing this on?

Linux

### Node Version

deno 1.27

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@devingfx Thank you for reaching out!

This just works on my end, but this might be related to deno vs node.

I have personally never pushed a new version for deno and I am unsure if anyone has done this beyond the initial release for deno.

I am also unable to verify if this works as intended or not with deno, I don't know anything about the deno ecosystem :/

devingfx:
Hi! Thanks for the reply.
I wanted somewhat to check if this part of the spec is actually coded in the plugin as nothing is written on this in the readme...

Let me describe more precisely my deno pipeline (with reproductible install and versions) with the actual css file for you to reproduce the error...
(Or maybe could you indicate the right deno focussed repo to issue)
I'll make more tests also with a nodejs pipeline...
I'm not sure to be able to spend time quickly though, so let me get in touch asap!

devingfx:
PS: One more thing: in my css @media is a custom media query, but this plugin is not deno ready so it's not configured in my pipeline!
Is it possible that the non transformation of this mq cause the error in postcss-nesting?

```css
.my {
    some-other: props here;
    @media (--mobile) {
        color: red;
    }
}
```

romainmenke:
> Or maybe could you indicate the right deno focussed repo to issue

This is the right repo for that :)
But we only published a couple of versions specifically for deno.
In practice we don't have the capacity (time & knowledge) to really support deno.
We only actively support npm.

I have been hoping that native npm support in deno would take over.
[...]

PR #698, opened 2022-11-10
wrap calls to selector parser in try/catch

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/673


plugins/css-blank-pseudo/src/index.ts
```diff
@@ -48,8 +48,8 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 
 				try {
 					selectorAST = parser().astSync(selector);
-				} catch (_) {
-					rule.warn(result, `Failed to parse selector : ${selector}`);
+				} catch (err) {
+					rule.warn(result, `Failed to parse selector : "${selector}" with message: "${err.message}"`);
 					return selector;
```
romainmenke, line 53:
```suggestion
					rule.warn(result, `Failed to parse selector : "${selector}" with message: "${err.message}"`);
					return [selector];
```

PR #699, opened 2022-11-11
postcss-media-queries-aspect-ratio-number-values

romainmenke:
https://github.com/csstools/postcss-plugins/tree/postcss-mediaqueries-aspect-ratio-number-values--passionate-cheetah-aeb6398989/plugins/postcss-media-queries-aspect-ratio-number-values#readme

`<ratio>` typed was relaxed to include floats, `calc()` and to allow a single value instead of the mandatory `<integer> / <integer>`.

Browsers first clamp each component of `<ratio>` to `<integer>` before evaluating `<ratio>`.
This means that we can multiply each component by an arbitrary large number.

`<number>/0` needs special handling and this is changed to a very large number to mimic infinity.

browser support : https://mrhenry.github.io/web-tests/#de023e11-f61a-4635-a74c-e4a152251bdf

-------

side note :

Naming becomes sucky when `-` is the only separator but I am assuming that no one is typing these by hand.
Either you are using `postcss-preset-env` and you don't think about it, or you are copy-pasting from our readme/install docs.


Antonio-Laguna (approved):
This looks great to me!

PR #707, opened 2022-11-16
postcss-nesting: group declarations

romainmenke:
I missed that [section 2.3 was rewritten](https://drafts.csswg.org/css-nesting/#mixing)

This is a really nice change for us because it means that we can closely follow the spec without breaking things like mixins.

```css
article {
  color: green;
  & { color: blue; }
  color: red;
}

/* equivalent to */
article {
  color: green;
  color: red;
  & { color: blue; }
}
```

### DIFF
--- a/experimental/postcss-nesting/package.json
+++ b/experimental/postcss-nesting/package.json
@@ -26,7 +26,6 @@
 		"mod.js"
 	],
 	"dependencies": {
-		"@csstools/selector-specificity": "^2.0.0",
 		"postcss-selector-parser": "^6.0.10"
 	},
 	"peerDependencies": {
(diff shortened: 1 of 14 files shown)


experimental/postcss-nesting/package.json
```diff
@@ -26,7 +26,6 @@
 		"mod.js"
 	],
 	"dependencies": {
-		"@csstools/selector-specificity": "^2.0.0",
```
romainmenke, line 29: left over from the main plugin, not used anymore

Antonio-Laguna (approved):
This is great! Good catch and I'm glad the spec covers that aspect!

Issue #705, opened 2022-11-16
postcss-custom-media (9.0.0)


nileshprajapati:
### Bug description

The recently updated postcss plugin version 9 and have noticed unusual CSS output compared to version 8.0.2.

Issue 1:- Double media queries nested within.
Issue 2:- output contains `@media screen and (color: ...  )` & `@media screen and (max-color: ... )`.

I have reverted the plugin to v8.0.2 and confirmed that the plugin works as expected.

### Source CSS

```css
@custom-media --screen-md-n-above (width >= 48em);
@custom-media --screen-lg-n-above (width >= 64em);
@custom-media --screen-xl-n-above (width >= 80em);

:where(h1, h2, h3, h4, h5, h6) {
	font-size: 20px);
	font-weight: 600;
	line-height: 1.5;

	@media screen and (--screen-md-n-above) {
		font-size: 22px;
	}

	@media screen and (--screen-lg-n-above) {
		font-size: 26px;
	}

	@media screen and (--screen-xl-n-above) {
		font-size: 28px
	}
}
```

### Expected CSS

```css
:where(h1, h2, h3, h4, h5, h6){
	font-size: 20px;
	font-weight: 600;
	line-height:1.5;
}
@media screen and (min-width: 48em){
:where(h1, h2, h3, h4, h5, h6){
		font-size: 22px;
}
	}
@media screen and (min-width: 64em){
:where(h1, h2, h3, h4, h5, h6){
		font-size: 26px;
}
	}
@media screen and (min-width: 80em){
:where(h1, h2, h3, h4, h5, h6){
		font-size: 28px;
}
	}
```

### Actual CSS

```css
:where(h1, h2, h3, h4, h5, h6) {
	font-size: 20px;
	font-weight: 600;
	line-height: 1.5;
}
@media (min-width: 48em) {
	@media screen and (max-color: 2147477350) {
		:where(h1, h2, h3, h4, h5, h6) {
			font-size: 22px;
		}
	}
}
@media not all and (min-width: 48em) {
	@media screen and (color: 2147477350) {
		:where(h1, h2, h3, h4, h5, h6) {
			font-size: 22px;
		}
	}
}
@media (min-width: 64em) {
	@media screen and (max-color: 2147477350) {
		:where(h1, h2, h3, h4, h5, h6) {
			font-size: 26px;
		}
	}
}
@media not all and (min-width: 64em) {
	@media screen and (color: 2147477350) {
[...]

PR #709, opened 2022-11-17
postcss-custom-media: avoid complex generated CSS in more cases

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/705

### DIFF
--- a/plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
+++ b/plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
@@ -1,5 +1,19 @@
 import { alwaysTrue, neverTrue } from './always-true-or-false';
-import { isGeneralEnclosed, isMediaAnd, isMediaConditionList, isMediaFeature, isMediaFeatureBoolean, isMediaNot, isMediaOr, isMediaQueryInvalid, isMediaQueryWithType, MediaFeature, MediaQuery, newMediaFeaturePlain, parse } from '@csstools/media-query-list-parser';
+import {
+	MediaFeature,
+	MediaQuery,
+	isGeneralEnclosed,
+	isMediaAnd,
+	isMediaConditionList,
+	isMediaFeature,
+	isMediaFeatureBoolean,
+	isMediaNot,
+	isMediaOr,
+	isMediaQueryInvalid,
+	isMediaQueryWithType,
+	newMediaFeaturePlain,
+	parse,
+} from '@csstools/media-query-list-parser';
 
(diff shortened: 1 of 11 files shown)


plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
```diff
@@ -1,5 +1,5 @@
 import { alwaysTrue, neverTrue } from './always-true-or-false';
-import { isGeneralEnclosed, isMediaAnd, isMediaConditionList, isMediaFeature, isMediaFeatureBoolean, isMediaNot, isMediaOr, isMediaQueryInvalid, isMediaQueryWithType, MediaQuery, newMediaFeaturePlain, parse } from '@csstools/media-query-list-parser';
+import { isGeneralEnclosed, isMediaAnd, isMediaConditionList, isMediaFeature, isMediaFeatureBoolean, isMediaNot, isMediaOr, isMediaQueryInvalid, isMediaQueryWithType, MediaFeature, MediaQuery, newMediaFeaturePlain, parse } from '@csstools/media-query-list-parser';
```
Antonio-Laguna, line 2: [nit] Should we make this a line per import so it's easier to see/maintain?

plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
```diff
@@ -104,6 +104,29 @@ export function transformComplexMediaQuery(mediaQuery: MediaQuery, replacements:
 
 		const replacement = replacements.get(name);
 		if (replacement) {
+
+			if (replacement.truthy.length === 1 && mediaQueryIsSimple(replacement.truthy[0])) {
+				let mediaFeature: MediaFeature | null = null;
+				const replacementMediaQuery = replacement.truthy[0];
```
Antonio-Laguna, line 110: [nit] Should this use destructuring?

Antonio-Laguna (approved):
This does look great to me!

Left really minor comments but feel free to discard them if you don't really agree

plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
romainmenke: Good catch!

The auto import feature in typescript makes me lazy.
It just keeps adding things to the list and I never really look at it :D

plugins/postcss-custom-media/src/transform-at-media/transform-at-media.ts
romainmenke: I liked the parallel with the condition : `(replacement.truthy.length === 1 && mediaQueryIsSimple(replacement.truthy[0]))`

PR #710, opened 2022-11-17
more install instructions

romainmenke:
- next.js
- postcss-load-config

example : https://github.com/csstools/postcss-plugins/blob/more-install-instructions--adaptable-pike-9b74e871a5/plugins/postcss-design-tokens/INSTALL.md


Antonio-Laguna (approved):
This looks good to me! Thanks for doing this!

PR #715, opened 2022-11-19
Update handling of ampersand in functional pseudo classes

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/5745#issuecomment-1320204844


romainmenke:
99% sure that this is more correct.
_Doesn't have to be released any time soon._

Issue #680, opened 2022-10-31
postcss-custom-media - Support variables for setting media rules


Preen:
### What would you want to propose?

Be able to set custom-media with variables such as
```postcss
$medium_up: 767px;
@custom-media --medium-up-viewport (min-width: $medium_up);
@custom-media --medium-up-viewport (min-width: design-token("media.breakpoints.medium-up", to px));
```

### Suggested solution

I dont have any :) 

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
@Preen Thank you for reaching out.

We can not add this to the `custom-media` plugin because this feature request deviates from the specification.

This plugin has a very clear and well defined scope that must align with what browsers will implement in the future.

------

```css
@custom-media --medium-up-viewport (min-width: design-token("media.breakpoints.medium-up", to px));
```

We can investigate how design tokens can be used in `at-rules`.
This would include `@custom-media` and `@media`.

For now I prefer to keep this unresolved because this is a feature that has been requested in the standardised design token format.

Inventing some syntax now would also imply a breaking change later if the two are not compatible.

Preen:
Thank you for your quick response and it sounds like a good way forward :)

romainmenke:
Hi @Preen 

We have released an update to our design tokens plugin : https://www.npmjs.com/package/@csstools/postcss-design-tokens

You can now use design tokens in any at rule :

```css
@media (min-width: design-token('viewport.medium')) {
	.foo {
		padding-bottom: design-token('size.spacing.medium' to rem);
	}
}

/* becomes */

.foo {
	color: #fff;
	padding-top: 16px;
	padding-left: 16px;
	padding-bottom: 1rem;
}

@media (min-width: 35rem) {
	.foo {
		padding-bottom: 1.1rem;
	}
}
```

romainmenke:
@mantismamita Do you have a concrete example?
I don't immediately understand how this would look like.

The design tokens plugin supports any and all at rules.

This will just work :

```css
@something-made-up design-token('foo') {}
```

So will this :

```css
@custom-media --foo (min-width: design-token('viewport-tablet' to px));
```

But it is important to keep in mind that there is no magic here.
[...]

PR #717, opened 2022-11-20
design-tokens : add support for at rules

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/680

`AtRule` params and `Declaration` values can both be parse with our tokenizer and parsing algorithms.

Because we do not need to consider the context we can use the same transform function for both.


plugins/postcss-design-tokens/src/transform.ts
```diff
@@ -0,0 +1,135 @@
...
+		if (!('walk' in componentValue)) {
+			return;
+		}
+
+		{
+			const replacements = transformComponentValue(componentValue, tokens, result, postCSSNode, opts);
+			if (replacements) {
+				componentValues.splice(index, 1, ...replacements);
+				didChangeSomething = true;
+				return;
```
romainmenke, line 22:
```suggestion
				return false;
```

Antonio-Laguna (approved):
This LGTM!

Thanks for the update. I think it adds a lot of value here!

Issue #721, opened 2022-11-21
postcss-custom-media v9 not replacing variables


gpoitch:
### Bug description

Since updating to v9+ (9.0.1) postcss-custom-media is not compiling and replacing custom media variables. Have been using v8 for a while with no issue.

### Source CSS

```css
@custom-media --tabletUp (min-width: 768px);

@media (--tabletUp) {
  a { color: red; }
}
```

### Expected CSS

```css
@media (min-width: 768px) {
  a { color: red; }
}
```

### Actual CSS

```css
@media (--tabletUp) {
  a { color: red; }
}
```

### Extra config

Used via webpack:

```js
{
  loader: 'postcss-loader',
  options: {
    postcssOptions: {
      plugins: [postcssCustomMedia(), autoprefixer()]
    }
  }
}
```

### What plugin are you experiencing this issue on?

PostCSS Custom Media Queries

### Plugin version

9.0.1

### What OS are you experiencing this on?

macOS

### Node Version

19.1.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@gpoitch Thank you for reaching out!

I did a quick check and it works on our end : https://github.com/csstools/postcss-plugins/pull/722/files

Are there maybe other plugins that can be interfering?

romainmenke:
Also see : https://github.com/csstools/postcss-plugins/issues/705#issuecomment-1316647241

It seems that for some users something is "sticky" and after updating the `@custom-media` are not transformed at first.

I don't know that user's or your setup.
Maybe there is a cached config or a watcher still running?

gpoitch:
Same issue with postcss-custom-media as the only plugin and cleaning any possible caches and reinstalling all node_modules.

I didn't see a changelog but maybe something changed with how separate files are combined? My setup has all the variables in its own file.

```css
@custom-media --tabletUp (min-width: 768px);
@custom-media --desktopUp (min-width: 1025px);
--primaryColor: blue;
--textColor: black;
```

romainmenke:
🤔 We did indeed change something could affect that.

We now use `prepare` to make it safe to re-use the same plugin instance multiple times.
[...]

PR #724, opened 2022-11-21
document how parallel builders can affect plugins

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/721

I don't think we can do anything except for warn users about potential issues.

example : https://github.com/csstools/postcss-plugins/blob/af1efe833aa93172dc20173c078bb3a2b95d0804/plugins/postcss-cascade-layers/INSTALL.md


Antonio-Laguna (approved):
Looks great to me! Thanks for adding this!

PR #730, opened 2022-11-26
Skip ignore comment checks for files with no ignore comments

marvinhagemeister:
Previously we'd cast every rule to a string and then check if an ignore comment is present. That constant casting and repeatedly calling the regex costs a lot of time. We can avoid that work if we know that the input source doesn't contain any ignore comments at all.

This speeds up postcss compilation time in one of my projects by about 4.6s. In that project there are no css files with postcss comments anywhere.

Before: 

postcss

### DIFF
--- a/plugins/postcss-custom-properties/src/lib/get-custom-properties-from-root.ts
+++ b/plugins/postcss-custom-properties/src/lib/get-custom-properties-from-root.ts
@@ -1,5 +1,5 @@
 import valuesParser from 'postcss-value-parser';
-import { isBlockIgnored } from './is-ignored';
+import { isBlockIgnored, hasIgnoreComment } from './is-ignored';
 
 // return custom selectors from the css root, conditionally removing them
 export default function getCustomPropertiesFromRoot(root, opts): Map<string, valuesParser.ParsedValue> {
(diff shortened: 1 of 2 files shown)


romainmenke:
@marvinhagemeister Thank you this!
I wasn't aware that this feature was causing performance issues.

-----

@Antonio-Laguna Maybe we should reconsider removing this feature?
We've discussed this in the past but we decided to keep it because it was causing any harm and it wasn't adding that much complexity.

Given that we are doing breaking changes for `postcss-preset-env v8` anyway I would prefer removing it entirely. That would make this plugin significantly faster even when users have one or more comments.

Thoughts?

marvinhagemeister:
Dropping support for ignore comments all together works for me 👍

Arborescent:
Another optimization is reusing saving the `RegExp` instance for later use.
Most of the cost of this code is compiling the regular expression while initializing the instance. Once it's compiled you can reuse it for a lower cost.

elijahrdorman:
> Another optimization is reusing saving the `RegExp` instance for later use. Most of the cost of this code is compiling the regular expression while initializing the instance. Once it's compiled you can reuse it for a lower cost.

Because regex literals don't change during runtime (like `RegExp()` might), JITs can and do choose to cache them if heuristics indicate they should (I may be wrong, but I believe this is mostly a factor of how many times it was used). This would be worth trying, but may not do much.

I'd certainly rather see it behind a plugin for the few people who want it and are okay with the performance hit.

romainmenke:
see : https://github.com/csstools/postcss-plugins/pull/737

The check for the comment is impossible to make really fast.
But we can make sure we do it as few times as possible.

marvinhagemeister:
@romainmenke Awesome sounds good to me! Thanks again for taking a look 👍

Issue #685, opened 2022-11-06
Can we make CI faster without shooting ourselves in the foot too much?


romainmenke:
The current approach was nice in theory but something changed and now it is backfiring.

Current system :

1. build and test everything
2. cache build output
3. cache current commit hash

subsequent commits :

1. take commit hash from last ±10 commits
2. use those commit hashes as cache restore keys
3. restore cache with build outputs when available
4. check what changed since the last build was stored
5. build only those things
6. test only those things

I am assuming that this is going wrong because the cache mechanic in GitHub workflows is not intended for this.

- are the restored files not ok or complete in some way?
- are we sure there is no cache pollution between branches?

--------------

Another way to have a similar system is to store build outputs in git.
This has it's own negative aspects but maybe it also has benefits.

store build outputs in git :

pro's :

- git repo is a complete and functional archive
- easier to optimize CI
- unpkg and friends can read from the GitHub repo

con's :

- noisy commits
- git conflicts when multiple editors are working on the same plugin
- updating dependencies is more work (need to rebuild everything and push changed output files)

---------------

Are there other alternatives I am not thinking of?

romainmenke:
_I always forget to explicitly mention this bit :_ 

A change to a plugin can affect `postcss-preset-env` or the `cli` package.
It is the major benefit of the monorepo. We can detect that a plugin change would be breaking for `postcss-preset-env` without needing to release it first.

This also means that we need to re-run the tests for `postcss-preset-env` and the `cli` each time a constituent plugin is altered.

To run those tests we need to have the contents of the `dist` folder for each constituent plugin.

Even when a change only concerns a single plugin we still need to build everything in the current setup.

Having a way to store the contents of the `dist` folder makes it possible to build and test subsets of the monorepo. Storing them in `git` is one such way.

jonathantneal:
Does the monorepo currently have the ability to create iterative builds?

If so, could the compiled files be stored in a separate `dist` branch? If so, perhaps a GitHub action on commit could kick off the iterative build in that branch and then test against that?

romainmenke:
> Does the monorepo currently have the ability to create iterative builds?

No :)

Storing `dist` files in `git` is one way to have this.

--

> If so, could the compiled files be stored in a separate dist branch?

Wouldn't this be very complex?
We would need a "dist" branch for `main` and for each feature branch that has sub branches.

We would also need automation to force push the exact write contents to these branches.
Which involves write-capable GitHub access tokens.

[...]

PR #731, opened 2022-11-27
Speed up and simplify CI by checking `dist` files into git

romainmenke:
proposal for : https://github.com/csstools/postcss-plugins/issues/685


jonathantneal:
Forgive me if I misunderstand this PR.

From my own experience, it can be quite troublesome to check in compiled files, as mismatches between those files and source can result in false positives or negatives in the tests themselves. If the goal is CI speed,  are there other viable solutions?

Could tests be more discriminatory in what is checked after a change? Could CI be configured to cache previously compiled files?

romainmenke:
> From my own experience, it can be quite troublesome to check in compiled files, as mismatches between those files and source can result in false positives or negatives in the tests themselves.

Do you have a concrete example of a case/scenario?

------

I've only encountered this one case :

- multiple developers are working on the same project
- that project "bundles" to one or more outputs
- person A is working on sub-feature X
- person B is working on sub-feature Y
- sub-feature X is merged
- person B needs to resolve conflicts locally by:
  - merge in all changes from the base branch
  - re-build the project
  - push

The merge conflicts arise mostly because the bundled outputs are also minified.

I am not overly concerned about this case because it is very rare for multiple people to be working on the same plugin at the same time.

If there was a way 

> Could tests be more discriminatory in what is checked after a change

There are two issues that need to be solved for "fast CI" :
- a way to reduce the amount of work that needs to be done
- a way to store/re-store past work

Without a way to store/re-store the contents of `dist` folders we can not reduce the amount of work that needs to be done.

_This aspect is not the issue._

>  Could CI be configured to cache previously compiled files?

We previously did exactly this :)
We cached the contents of all `dist` folders via GitHub actions caching and then restored them. This approach abused the git log and how cache keys work in GitHub actions.

This worked well for a few months and then it started breaking.
To make this work well we need a much more reliable service to store build artifacts.

-------

If there is an alternative that has the benefits but no drawbacks then obviously I am all ears :) But maybe the linked issue is more suitable for that discussion?

https://github.com/csstools/postcss-plugins/issues/685

romainmenke:
🤔 This does however mean that we lose the ability to edit source code via GitHub UI.
Could bring that back with a dedicated GitHub workflow that can do these steps :
- build all needed plugins
- commit changed files
- push back to same branch

Antonio-Laguna (approved):
This does look good to me.

I agree with the objective and we can always pivot if we find out that conflicts start to happen very often.

Thanks (as usual) for thinking ahead on this and taking the lead!

romainmenke:
Thank you @Antonio-Laguna and @jonathantneal for the feedback on this!

PR #734, opened 2022-11-29
drop-in replacement for `importFrom`/`exportTo`

romainmenke:
What I did :

- copy/paste the existing plugins as they are now on the `main` branch
- removed all the "polyfill/fallback" behaviour
- added `importedStylesOverrideDocumentStyles` as plugin option to all plugins
- removed `overrideImportFromWithRoot` from custom-properties

I picked a bad name in `overrideImportFromWithRoot`.
It is very confusing what is overriding what purely from the name.
I hope `importedStylesOverrideDocumentStyles` is clearer.


Antonio-Laguna:
After thorough consideration, we don't think this is the right way to move forward. We still strongly believe that there are better ways to solve the issues where `importFrom` and `exportTo` are used today. We will keep our decision to get rid of `importFrom` and `exportTo` from the postcss-preset-env ecosystem starting with version 8 and upwards. 

Our main concern is that we have limited resources to create and maintain all the plugins here.

We have separated the removed functionality into separate plugins and the source code for this is available to anyone who wants to take over ownership.

We'll keep this visible for some reasonable time and then we will just close it.

romainmenke:
Going to close this.
We can always restore if needed later.

Issue #481, opened 2022-06-16
custom selector and custom media do not play nice with cascade layers


romainmenke:
I think this behaviour is largely undefined at the moment but declaring custom selectors and custom media inside `@layer` does not work.

Should it work?

Antonio-Laguna:
wow, that's a big question. I'm not sure. On one hand it's like, nope, shouldn't work and the other hand is a big _mayybeee?_

romainmenke:
> mayybeee

Hehe, after typing this up I kept coming back to that :D

--------

I think in theory they could work, but them being nested in `@layer` just doesn't have any effect.

It will be safest if we keep the current (broken) behaviour and wait until the relevant specs are updated.

------

Forgot to mention that I encountered this with imported styles where it is a bit more unexpected.

```css
/* file-alpha.css */

mirisuzanne:
I'm pretty sure we did [spec this](https://drafts.csswg.org/css-cascade-5/#layering) - but in a short mention nested between examples:

> Name-defining at-rules such as `@keyframes` or `@font-face` that are defined inside cascade layers also use the layer order when resolving name collisions.

I think that would apply to any sort of custom-property/selector/media at-rules, which are generally name-defining. So that means they should:

1. Work as expected even inside `@layer`
2. If there are name conflicts, the higher layer should win (this can be achieved by ordering them, I think?)

romainmenke:
Thank you @mirisuzanne for finding this issue and sharing this.

I've been working on an updated version which ensures this works as you describe : https://github.com/csstools/postcss-plugins/pull/543

The main issue was that the plugins that handle custom selectors and custom media transformed the CSS before the cascade layers plugin.

This is part of the upcoming major release of `postcss-preset-env` where we hope to align everything much more closely with the CSS specifications.

------

_I got side-tracked a bit recently taking on maintenance for the `polyfill-library` and 9to5 work. Hopefully more time will be available soon!_

romainmenke:
1. we want to ship v8 with a version of nesting that will land in browsers, we don't want to ship v9 a short time after only because of nesting
2. we want to have solved `@media` param parsing
3. we want to have an implementation of logical properties that is much simpler and more correct at the same time.

I am currently working on 1 and 2.
I would not count on v8 being ready within the next few weeks :)

romainmenke:
[...]

PR #739, opened 2022-12-02
fix cascade layers in combination with nesting and name defining at rules

romainmenke:
- [x] `postcss-preset-env`
- [x] `cascade-layers`
- [x] `custom-media`
- [x] `custom-selectors`
- [x] `custom-properties`

see : https://github.com/csstools/postcss-plugins/issues/481#issuecomment-1335678359

By running `cascade-layers` before any other plugin we forced that plugin to support everything. This included things like nested css and all future css.

The `cascade-layers` plugin needed to understand the most modern syntax.

This pull request rolls back the recent changes and implements :
- `cascade-layers` is always last
- all plugins for name defining at rules needs to have a basic understanding of Cascade Layers

We shift the burden from the `cascade-layers` plugin to a select number of other plugins.
Those plugins do not need to be capable of the entire Cascade Layers feature. Only just enough to know which declaration has priority if there are naming conflicts.

------

This issue did made me realize that Cascade Layers will complicate future plugins.
We have to assume that anything can be layered and that both layers and source order needs to considered.

Plugins that only provide fallbacks for declarations or selectors aren't affected.
But anything that interacts with the structure will need to take Cascade Layers into account.


Antonio-Laguna (approved):
This LGTM!

Thanks for adding this!

PR #740, opened 2022-12-03
PostCSS Logical revamp

### LINKED ISSUE #736: Proposal for `postcss-logical`
The current plugin re-writes selectors with `:dir()`.
But it is incorrect because it only considers `ltr` and `rtl` and it also changes selector specificity which can lead to bugs.

I don't think the goal must be to have a true polyfill.
This will never work without causing issues.

We can however create a really good fallback.

https://drafts.csswg.org/css-logical/#intro

The plugin inputs are :
- block direction
- inline direction

For English this is :

- block: top to bottom
- inline: left to right

For Chinese :

- block: right to left
- inline: top to bottom

--------

For this CSS in English

```css
magin-block-start: 20px;
```

- block -> top to bottom
- start -> top
- top -> `margin-top: 20px;`

In Chinese :

- block -> right to left
- start -> right
- right -> `margin-right: 20px;`

-------

The plugin doesn't do anything more than this :

- user configures their fallback mode
- plugin maps to that fallback

------

The downside is that you can not switch modes in the same document.
But here it is important to remember that it is a fallback.
The aim is to have a readable document in both old and new browsers, not to match the full feature.

Antonio-Laguna:
Fixes #736 

See [README](https://github.com/csstools/postcss-plugins/tree/feature/logical-revamp-v8/plugins/postcss-logical#readme)
See [Wiki](https://github.com/csstools/postcss-plugins/wiki/PostCSS-Preset-Env-8/#postcss-logical-breaking-changes)
See [CHANGELOG](https://github.com/csstools/postcss-plugins/blob/feature/logical-revamp-v8/plugins/postcss-logical/CHANGELOG.md)

I hope everything is covered here. I feel tests aren't as thorough but often not sure how deeper I should get with those!


plugins/postcss-logical/src/index.ts
```diff
@@ -0,0 +1,251 @@
...
+			// 2.1 Caption
+			'caption-side': makeTransform(transformValue(directionConfig)),
+			// 2.3 Text Align
+			'text-align': makeTransform(
+				directionConfig.inlineIsHorizontal
+					? transformTextAlign(directionConfig.inline)
+					: null,
+			),
+			// 2.4 Resize
+			'resize': makeTransform(transformResize(directionConfig)),
```
romainmenke, line 105: This is not yet supported by Blink and we don't track in CSSDB

mherchel:
Note that the link to the CHANGELOG above is broken.

romainmenke:
I've added the updated changelog location above : https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-logical/CHANGELOG.md#600-january-24-2023

Issue #750, opened 2022-12-12
`:has` pseudo selector not working properly on hover


nlhkabu:
### Bug description

I have an interface where I have items nested inside other items.

Each item has a handle for drag and drop (on the left) and a menu (on the right).

i want to show the handle and menu on hover, but only for the item that I am directly hovering on, **not** it's parent (or parents)

This is working well on Chrome (without polyfill):

![Screenshot from 2022-12-12 11-06-10](https://user-images.githubusercontent.com/3323703/207030360-d833b199-b886-4a81-a97c-0e95c0d4e98f.png)

But this is what I get with Firefox (107):
![Screenshot from 2022-12-12 11-06-09](https://user-images.githubusercontent.com/3323703/207030390-60f7fd7b-1809-4086-880e-035f3053bc56.png)

My setup

package.json
```
"postcss-preset-env": "8.0.0-alpha.1",
```

index.html
```
<script src="https://unpkg.com/css-has-pseudo@4.0.1/dist/browser-global.js"></script>
<script>cssHasPseudo(document)</script>
```

postcss.config.js
```
const postcssPresetEnv = require("postcss-preset-env");

module.exports = {
  plugins: [
    postcssPresetEnv({
      features: {
        "nesting-rules": true,
        "has-pseudo-class": true,
      },
    }),
  ],
};
```

### Source CSS
```
.draggable:hover:not(:has(.nested-list:hover)) {
  & > .entry .grabber {
    display: flex;
  }

  & > .entry .entry-menu {
    display: inline;
  }
}
```

### Expected CSS

??

### Actual CSS

??

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

N/A

### What plugin are you experiencing this issue on?

CSS Has Pseudo

### Plugin version

8.0.0-alpha.1

### What OS are you experiencing this on?

Linux

### Node Version

v16.15.1

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@nlhkabu Do you also have an html snippet we could use to test with?

nlhkabu:
@romainmenke here you go:
```html
<div class="draggable" role="listitem" >
    <div class="entry">
[...]

PR #751, opened 2022-12-12
has pseudo : fix cleanup of rules in browsers with native support

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/750

-------

We have an unpublished breaking change for this plugin : minimum node version bump.

To make this patch available in `postcss-preset-env` `v8` we need to ship a major of this plugin and a new alpha for `postcss-preset-env`.


plugins/css-has-pseudo/src/browser.js
```diff
@@ -4,18 +4,9 @@ import '@mrhenry/core-web/modules/~element-qsa-has.js';
...
 	try {
-		// Chrome does not support forgiving selector lists in :has()
-		document.querySelector(':has(*, :does-not-exist, > *)');
-		document.querySelector(':has(:has(any))');
-
-		// Safari incorrectly returns the html element with this query
-		if (document.querySelector(':has(:scope *)')) {
-			return false;
-		}
-
```
romainmenke, line 17: The specification changed so we need to update the feature detection to match.
Functional changes were done in the `querySelector(':has()')` polyfill from `core-web`

plugins/css-has-pseudo/src/browser.js
```diff
@@ -4,18 +4,9 @@ import '@mrhenry/core-web/modules/~element-qsa-has.js';
...
-		if (!('CSS' in self) || !('supports' in self.CSS) || !self.CSS.supports(':has(any)')) {
+		if (!('CSS' in self) || !('supports' in self.CSS) || !self.CSS.supports('selector(:has(div))')) {
```
romainmenke, line 9: Bug fix : needed to check for `supports('selector(...)')`

plugins/css-has-pseudo/src/browser.js
```diff
@@ -250,7 +241,7 @@ export default function cssHasPseudo(document, options) {
 						}
 
 						if (!options.mustPolyfill) {
-							rule.deleteRule();
+							styleSheet.deleteRule(index);
```
romainmenke, line 244: This was the actual bug.
The fallback rules were never removed but the polyfill wasn't active because Chrome has native support.

Combined with `:not()` you can get funky results.

This changes correctly cleans up the fallback rules for browsers with native support.

plugins/css-has-pseudo/src/index.ts
```diff
@@ -24,7 +24,7 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 	const specificityMatchingNameTag = ':not(' + options.specificityMatchingName + ')';
 
 	return {
-		postcssPlugin: 'css-has-pseudo-experimental',
+		postcssPlugin: 'css-has-pseudo',
```
romainmenke, line 27: Left over from experimental track.

PR #754, opened 2022-12-14
specificity : add support for `::slotted` / `:host` / `:host-context`

romainmenke:
see :
- https://github.com/bramus/specificity/commit/12557b78948bb4cb6438ac51e4c81c850cde5a4e
- https://github.com/bramus/specificity/commit/02c05afcb2bca5f77f860d7f99d5bd7474737205

`bramus/specificity` tries to enforce the specification by throwing away information if someone uses a selector list or a complex selector.

I think it is better to follow the most common form in CSS which is to take the most specific list item (i.e. the same as `:is()`, `:has()` ... )

The selectors are invalid so it is a bit of a hypothetical.
I want to avoid users getting confused if they copy/paste a selector into `:host` and see the specificity drop.


packages/selector-specificity/src/index.ts
```diff
@@ -37,7 +37,37 @@ export function selectorSpecificity(node: Node): Specificity {
...
+
+		switch (node.value.toLowerCase()) {
+			case '::slotted':
+				// “The specificity of ::slotted() is that of a pseudo-element, plus the specificity of its argument.”
+
+				{
+					c += 1;
+
+					if (node.nodes && node.nodes.length > 0) {
+						// We are more forgiving that the specification and use the most specific complex selector.
```
Antonio-Laguna, line 50: very much [nit]

```suggestion
						// We are more forgiving than the specification and use the most specific complex selector.
```

packages/selector-specificity/src/index.ts
```diff
@@ -114,6 +144,30 @@ export function selectorSpecificity(node: Node): Specificity {
...
+				// “The specificity of :host is that of a pseudo-class. The specificity of :host() is that of a pseudo-class, plus the specificity of its argument.”
+				// “The specificity of :host-context() is that of a pseudo-class, plus the specificity of its argument.”
+
+				{
+					b += 1;
+
+					if (node.nodes && node.nodes.length > 0) {
+						const mostSpecificListItem = specificityOfMostSpecificListItem(node.nodes);
+
+						// We are more forgiving that the specification and use the most specific complex selector.
```
Antonio-Laguna, line 158: [nit] same as above

```suggestion
						// We are more forgiving than the specification and use the most specific complex selector.
```

Antonio-Laguna (approved):
This looks good to me and also makes sense!

I've caught what I think are some typos in some comments but it does look good to me!

packages/selector-specificity/src/index.ts
romainmenke: Good catch! thank you 🙇

PR #755, opened 2022-12-17
cascade-layer-name-parser

romainmenke:
Since we are adding support for Cascade Layers into individual plugins we need a better way to parse layer names.

This should be integrated into `postcss-cascade-layers` but that is a larger refactor.
Adding it to these 3 plugins at least keeps our tech debt in check.


Antonio-Laguna (approved):
This is looking great!

Thanks for extracting this and also the thorough tests and meaningful errors that point to spec parts!

PR #760, opened 2022-12-24
css-tokenizer and parser-algorithms : fixes and performance improvements

romainmenke:
## css-tokenizer :

Large refactor to improve performance.
I've gone over the source and replaced many aspects with slightly faster variants that produce the same result.

## parser-algorithms :

I've removed the `UnclosedFunctionNode` and `UnclosedSimpleBlockNode`.
They were not needed

see : https://drafts.csswg.org/css-syntax/#consume-simple-block

Screenshot 2022-12-24 at 11 48 27

I initially implemented this differently but there was no reason for that.
There are no benefits to parsing as an `UnclosedFunctionNode`.

Also added a bunch of tests from WPT.

-----

These are breaking changes but the bits of the API that were changed aren't in use.

-------

I've also polished the public API a bit and added some more helpers for common operations.
(e.g. getting the name of a media feature in the media query parser.)


Antonio-Laguna (approved):
This is great!

The code looks sane! I love how simpler it got in some areas and also I really like the addition of more community tests for the tokenizer.

Thanks a lot for adding this and also all of the explanations and care you put into the code and the PR! 🙇🏻

PR #773, opened 2023-01-08
feature/logical-revamp-v8 : feedback

romainmenke:
These are mainly some formatting nits and a few superficial changes.
This should not affect any of the logic of the plugin.


plugin-packs/postcss-preset-env/test/basic.css
```diff
@@ -354,7 +354,7 @@
 	right: round(nearest, 2.5px, 1px);
 	bottom: round(up, 2.5px, 1px);
 	left: round(down, 2.5px, 1px);
-	inset: round(to-zero, 2.5px, 1px);
```
romainmenke, line 357: This test is for `round()` but logical transforms `inset`.
This confused me so I changed the test to `padding-left`

plugins/postcss-logical/src/lib/clone-declaration.ts
```diff
@@ -4,6 +4,12 @@ export function cloneDeclaration(
...
 	value: string,
 	prop: string,
-) : Declaration {
-	return declaration.cloneBefore({ value, prop });
+): void {
+	if (declaration.parent && declaration.parent.some((x) => {
+		return x.type == 'decl' && x.prop === prop && x.value === value;
+	})) {
+		return;
+	}
```
romainmenke, line 12: I am unsure about this.
It is a nice addition that prevents duplicate declarations like those added in `css-has-pseudo`.

But it can also slow down the plugin.
I don't mind removing this again.

plugins/postcss-logical/src/lib/transform-inset.ts
```diff
@@ -20,24 +21,30 @@ export function transformInset(
...
 
+		if (hasVariables && values.length !== 4) {
+			// A variable can "resolve" to multiple separate values.
+			// This makes it impossible to statically map tokens to longhand properties.
+			return;
+		}
+
 		if (values.length === 1) {
 			newRules.top = valueParser.stringify(values[0]);
-			newRules.right = valueParser.stringify(values[0]);
```
romainmenke, line 25: `valueParser.stringify` can be slow and expensive.
Better to re-use work.

plugins/postcss-logical/src/lib/types.ts
```diff
@@ -1,18 +1,18 @@
...
 
 export const DirectionValues = {
-	BlockStart: `${Direction.Block}-${DirectionValue.Start}`,
-	BlockEnd: `${Direction.Block}-${DirectionValue.End}`,
-	InlineStart: `${Direction.Inline}-${DirectionValue.Start}`,
-	InlineEnd: `${Direction.Inline}-${DirectionValue.End}`,
+	BlockStart: 'block-start',
+	BlockEnd: 'block-end',
+	InlineStart: 'inline-start',
+	InlineEnd: 'inline-end',
```
romainmenke, line 15: Writing this as strings gives a better experience navigating the code.
On hover it shows `'block-start'` instead of `${Direction.Block}-${DirectionValue.Start}`

But I don't mind changing this back.

plugins/postcss-logical/src/lib/clone-declaration.ts
Antonio-Laguna: I don't think speed would be much impacted in the average CSS file, let's keep this :)

plugins/postcss-logical/src/lib/transform-inset.ts
Antonio-Laguna: Nicely done! I also think this is more clear!

plugins/postcss-logical/src/lib/types.ts
Antonio-Laguna: That's fine!

plugins/postcss-logical/src/utils/parse-value-couple.ts
```diff
@@ -1,19 +1,25 @@
...
 	let valueB;
 
 	if (values.length === 1) {
+		if (hasVariables) {
+			// A variable can "resolve" to multiple separate values.
+			return;
+		}
+
 		valueA = valueParser.stringify(values[0]);
 		valueB = valueParser.stringify(values[0]);
```
Antonio-Laguna, line 24:
```suggestion
		valueA = valueParser.stringify(values[0]);
		valueB = valueA;
```

plugins/postcss-logical/src/utils/parse-value-couple.ts
```diff
@@ -1,19 +1,25 @@
...
+		const errorMessage = `[postcss-logical] Invalid number of values for ${declaration.prop}. Found ${values.length} values, expected 1 or 2.`;
+		throw declaration.error(errorMessage);
 	}
 
 	let valueA;
 	let valueB;
 
 	if (values.length === 1) {
+		if (hasVariables) {
+			// A variable can "resolve" to multiple separate values.
```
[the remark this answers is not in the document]
romainmenke: I agree, it could be a setting maybe, but we can add that later in a feature relaase.

PR #775, opened 2023-01-09
add a few abstract tests with various configs

romainmenke:
This adds a few tests which are not focused a specific CSS feature but rather on how we handle plugin options.


Antonio-Laguna (approved):
LGTM! Thanks for adding these

PR #776, opened 2023-01-10
logical : border radius

romainmenke:
If I understand the feature correctly there is some complex double/triple mapping for border radius.

I am 20% certain this is correct.


plugins/postcss-logical/src/lib/transform-borders.ts
```diff
@@ -56,25 +57,44 @@ export function transformBorderShorthand(
...
 	return (declaration: Declaration) => {
 		let prop;
 
 		switch (declaration.prop.toLowerCase()) {
-			case 'border-start-start-radius':
-				prop = `border-top-${start}-radius`;
+			case 'border-start-start-radius': {
+				const replacement = config.inlineIsHorizontal ?
+					`${config.block[0]}-${config.inline[0]}` :
+					`${config.inline[0]}-${config.block[0]}`;
```
romainmenke, line 69: - First `start` is always `block`.
- So we have to take the value at index `0` from `config.block`.
- Second `start` is always `inline`.
- So we have to take the value at index `0` from `config.inline`.

The result might be : `left-bottom`.
So we check if the orientation is different and inverse the parts if needed.

plugins/postcss-logical/src/lib/transform-borders.ts
```diff
@@ -56,25 +57,44 @@ export function transformBorderShorthand(
...
+
+				prop = `border-${replacement}-radius`;
 				break;
-			case 'border-start-end-radius':
-				prop = `border-top-${end}-radius`;
+			}
+			case 'border-start-end-radius': {
+				const replacement = config.inlineIsHorizontal ?
+					`${config.block[0]}-${config.inline[1]}` :
+					`${config.inline[1]}-${config.block[0]}`;
```
romainmenke, line 77: - First `start` is always `block`.
- So we have to take the value at index `0` from `config.block`.
- Second `start` is always `inline`.
- So we have to take the value at index `1` from `config.inline`.

The result might be : `left-bottom`.
So we check if the orientation is different and inverse the parts if needed.

Antonio-Laguna (approved):

PR #778, opened 2023-01-12
postcss-custom-properties: Enable usage of function for preserve option

bjentsch:
[postcss-css-variables](https://github.com/MadLittleMods/postcss-css-variables) has a neat configuration option that enables devs to preserve custom properties under certain circumstances, using a function.

What it basically does is that it passes a declaration object to the configured function, which does its filtering magic and then returns a boolean indicating if the custom property should be preserved alongside the compiled output. Possible criteria for the function could be the declarations content, filename, ... you name it.

This PR is my go at implementing this functionality as an addition to only specifying `true`/`false` on the `preserve` configuration option.

WDYT?


romainmenke:
Hi @bjentsch 👋 

What is the motivation for preserving some properties but not others?

bjentsch:
Hi @romainmenke :v: 

With postcss-css-variables, we used injection from the TypeScript/JS side to change the variables at runtime while still providing a fallback for older browsers that don't natively support custom properties yet. We would like to replicate this behaviour with postcss-custom-properties.

When preserving some of the variables, those can then be dynamically altered by scripting during runtime, e.g. for a darkmode switch or dynamic color state transitions.

romainmenke:
But then why not preserve all CSS variables?
I am mainly curious why this much control is needed.

Is it to minify the CSS bundle?

romainmenke:
If the variables that should be inlined are more like constants it might be interesting to use something like design tokens : https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-design-tokens#readme (there are other plugins that do similar things) That plugin inlines final values and doesn't produce CSS variables. 

You can also spin of your own variant of `postcss-custom-properties` dedicated to your specific needs.

But I don't think this feature request aligns with the current goal of the plugin, which is to polyfill custom properties in old browsers like IE. We try to keep these plugins focussed on a single use case.

We have just finished removing `importFrom` and `exportTo` from this plugin for this exact reason :)
People had major issues with these plugin options when their browserslist updated and `postcss-preset-env` stopped running the plugin. Then suddenly all their values were gone.

romainmenke:
Yes, that is why a fork would be interesting.
It would allow you to specialize this plugin to suite you needs.

I am sure there will be others with the same needs and we are happy to advice on and share your version (if you decide to build and publish it).

bjentsch:
Closing the PR as I don't see a chance of this getting merged, thanks for your input @romainmenke. 🤗

romainmenke:
I was just about to comment here too 🤗 

Out of curiosity is there anything in particular that would make it easier for you to maintain and publish a fork?

We sometimes have to say no to a feature request like this, but it is always a tough choice.
We do want you to have the feature even if we do not think it fits within our scope.

PR #779, opened 2023-01-13
css-tokenizer : remove `commentsAreTokens`

romainmenke:
I've never used `commentsAreTokens: false` and I am unsure why anyone would.
This was originally added because browsers throw away all comments and I wanted a flag to preserve them.

Removing it simplifies everything a little bit.


Antonio-Laguna (approved):
This looks good to me and also makes sense to have it always `true`!

Issue #496, opened 2022-06-23
[postcss-logical] Add logical units


thewilkybarkid:
### What would you want to propose?

Add support for the `vb` and `vi` units from the [CSS Values and Units Module Level 4](https://www.w3.org/TR/css-values-4/) spec (the logical equivalents of `vh` and `vw`).

This would allow for the likes of `block-size: 100vb`.

### Suggested solution

Expand `postcss-logical` to support logical units.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
This can be done at a token level for both `at-rules` and `declaration` values.

- tokenize with `@csstools/css-tokenizer`
- iterate over all tokens
- replace `dimension` token units when needed

romainmenke:
Hi @thewilkybarkid 

There is now a dedicated plugin for this feature : https://www.npmjs.com/package/@csstools/postcss-logical-viewport-units

It will be part of the upcoming major release of `postcss-preset-env`.

Thank you again for requesting this.

niklasgrewe:
@romainmenke 
> It will be part of the upcoming major release of postcss-preset-env

do you mean `8.0.0` or `9.0.0`? I am currently using `8.0.1` and can confirm that the additional plugin `postcss-logical-viewport-units` is not yet part of `postcss-preset-env`

romainmenke:
@niklasgrewe that is weird 🤔 
It definitely is part of `postcss-preset-env` : https://github.com/csstools/postcss-plugins/blob/main/plugin-packs/postcss-preset-env/scripts/plugins-data.json#L148-L150

You can also see this on the playground : https://preset-env.cssdb.org/playground/

```css
body {
	width: 20vi
}
```

<img width="660" alt="Screenshot 2023-02-18 at 19 09 18" src="https://user-images.githubusercontent.com/11521496/219881624-38b24c2f-b386-44a9-b383-e08930c48899.png">

-----------

Can you check a few things?

1) the result of `npm why` to check if the plugin is present:
```
npm why @csstools/postcss-logical-viewport-units
```

[...]

PR #780, opened 2023-01-14
postcss-logical-viewport-units

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/496


plugins/postcss-logical-viewport-units/src/index.ts
```diff
@@ -0,0 +1,97 @@
...
+				const lowerCaseValue = decl.value.toLowerCase();
+				if (!(lowerCaseValue.includes('vb') || lowerCaseValue.includes('vi'))) {
+					return;
+				}
+
+				// Declaration already has a fallback
+				const prev = decl.prev();
+				if (prev && prev.type === 'decl' && prev.prop === decl.prop) {
+					return;
+				}
```
Antonio-Laguna, line 52: Seeing this makes me wonder a couple of things:

Should we have some global util for this? We have this in the form of checking from the parent and also checking previous. 
Should it be standard?
Could there be any use case in which you (as a user) wouldn't want this behaviour?

Antonio-Laguna (approved):
This LGTM!

Nice one!

plugins/postcss-logical-viewport-units/src/index.ts
romainmenke: > Could there be any use case in which you (as a user) wouldn't want this behaviour?

I don't think this will be observable to users.
Either they have added a manual fallback and it will work or the plugins will add the needed fallbacks automatically.

The only thing that will be omitted are some unneeded intermediate values.

I am indeed being extra cautious with this plugin in particular because it is for a unit.
With multiple plugins generating fallbacks it can quickly escalate because a single declaration can contain multiple features.

plugin 1 : `alpha` -> `one`
plugin 2: `beta` -> `two`

```css
.foo {
  bar: alpha beta;
}
```

plugin 1 results in :

```css
.foo {
  bar: alpha beta;
  bar: one beta;
}
```

plugin 2 results in :

```css
.foo {
  modern: alpha beta;
  modern: alpha two;
  modern: one beta;
  modern: one two;
}
```

plugin 1 runs again :

```css
.foo {
  modern: alpha beta;
  modern: alpha two;
  modern: one two;
  modern: one beta;
  modern: one two;
}
```

------

By short circuiting we get :

plugin 1 results in :

```css
.foo {
  bar: alpha beta;
  bar: one beta;
}
```

plugin 2 results in :

```css
.foo {
  modern: alpha beta;
  modern: one beta;
  modern: one two;
}
```

Ideally we wouldn't have `one beta` but that is the cost of having separate plugins and not being lightningcss :)

--------

> Should it be standard?

Maybe it should.
Wouldn't be complex to add.

--------

> Should we have some global util for this

Yes, but I don't like the progressive custom properties thing I made.
It is still a pain to setup and it makes things more complex, not less so.

So it can be a shared thing but it must be a real reduction in complexity :)

romainmenke: I also already documented this :)

https://github.com/csstools/postcss-plugins/wiki/Plugin-best-practices#manual-fallback-declarations

PR #787, opened 2023-01-16
fix(postcss-cascade-layers): fix example of `postcss-cascade-layers`

CGQAQ:
in `#/#`, `/` WILL NOT WORK， should be `\`


romainmenke (approved):
Thank you for this @CGQAQ 🙇

PR #788, opened 2023-01-16
chore(postcss-cascade-layers): Change `#\#` to `_`

CGQAQ:
according to <https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#tag-name>, html tag name can only consists characters in the range `0–9`, `a–z`, and `A–Z`.
So `_` should be enough


yisibl:
Yes, we need to ensure that Specificity is added using the ID selector. this is the rationale for this plugin.

romainmenke:
> `#\#` is to fake an ID selector which increases the specificity

correct :)

PR #794, opened 2023-01-21
text-decoration : fix shorthand in Safari with single node values

romainmenke:
See : https://github.com/postcss/autoprefixer/issues/1473#issuecomment-1243370592

In Safari `text-decoration: underline;` is still interpreted as a long hand property.
This causes some issues.

```css
.foo {
  text-decoration-color: pink;
}

.foo--modifier {
  text-decoration: underline;
}
```

In Chrome and Firefox `.foo--modifier` will be applied as a shorthand and will be equivalent to :

```css
.foo--modifier {
  text-decoration-color: currentColor;
  text-decoration-line: underline;
  text-decoration-style: solid;
}
```

In Safari however it will be equivalent to :

```css
.foo--modifier {
  text-decoration-line: underline;
}
```

In other words `.foo--modifier` will have a different color in Safari than it will have in Chrome/Firefox. `pink` vs. `currentColor`.

This can be fixed by adding the vendor prefixed `-webkit-text-decoration`.
This is a shorthand property.

This fix is too specific for autoprefixer and might even be a breaking change for them.
We are doing major versions anyway and it's easier for us to do hacky things.
I don't think it will break anything for anyone, but I appreciate that autoprefixer is very careful with changes like these.

A simplified illustration that you can test in multiple browsers : https://romainmenke.github.io/text-decoration-prefix-validation/simplified.html

-----

I also had to fix some version info in `package.json`'s.
We did that one time release of `css-has-pseudo` but hadn't yet changed the version in our local `package.json`. This finally backfired today :)


plugins/postcss-text-decoration-shorthand/src/index.ts
```diff
@@ -145,9 +145,25 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 
 					const nonShortHandValue = valueParser.stringify(data.line);
 					if (decl.value.toLowerCase() === nonShortHandValue) {
+						const next = decl.next();
+						if (!next || next.type !== 'decl' || next.prop.toLowerCase() !== 'text-decoration') {
```
romainmenke, line 149: Have to look at `next`, not previous because we only want to do this for declarations written by authors. Not declarations produced by this plugin.

plugins/postcss-text-decoration-shorthand/src/index.ts
```diff
@@ -145,9 +145,25 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
...
+							decl.cloneBefore({
+								prop: '-webkit-text-decoration',
+								value: nonShortHandValue,
+							});
+						}
+
 						return;
 					}
 
+					decl.cloneBefore({
```
romainmenke, line 162: This re-order is a no-op but this groups related bits of code better.

Antonio-Laguna (approved):
Nice catch! I didn't really know this was an issue in Safari! 

Thanks also for putting that test page together!

PR #795, opened 2023-01-21
remove deno support from postcss-nesting

romainmenke:
We never released updates for Deno and are not helping users by documenting this as supported.

If we want to publish to Deno, it must be setup in a more maintainable way and for all plugins, not just nesting.

Deno is actively working on better support for JavaScript hosted on NPM so I prefer we don't do anything for Deno ourselves : https://deno.land/manual@v1.29.4/node/npm_specifiers

Can we remove packages we published to Deno's registry?


plugins/postcss-nesting/CHANGELOG.md
```diff
@@ -1,11 +1,10 @@
 # Changes to PostCSS Nesting
 
-⚠️ We will most likely not finish this release and switch over to the plugin being worked on in the experimental track.
```
romainmenke, line 3: Because the specification will not be stable in time we will continue using this plugin for a bit longer.

Antonio-Laguna (approved):
This LGTM! Thanks for adding this

PR #798, opened 2023-01-22
Wire logical into Preset Env and CLI

Antonio-Laguna:


plugin-packs/postcss-preset-env/src/lib/get-options-for-browsers-by-feature.mjs
```diff
@@ -45,6 +45,17 @@ export default function getOptionsForBrowsersByFeature(browsers, feature, cssdbL
...
+			return {};
+		case 'logical-properties-and-values':
+		case 'float-clear-logical-values':
+		case 'logical-resize':
+		case 'logical-viewport-units':
+			{
+				if ('logical' in options) {
+					return options.logical;
+				}
+			}
```
Antonio-Laguna, line 57: After a lot of thinking about where this should live, I thought this was the best place but let me know if you disagree!

plugin-packs/postcss-preset-env/src/lib/shared-options.mjs
```diff
@@ -1,11 +1,7 @@
 export function initializeSharedOptions(opts) {
 	if ('preserve' in opts) {
 		const sharedOptions = {};
-
-		if ('preserve' in opts) {
-			sharedOptions.preserve = opts.preserve;
-		}
-
```
Antonio-Laguna, line 8: I don't think this was actually needed?

plugin-packs/postcss-preset-env/FEATURES.md
```diff
@@ -19,6 +19,7 @@ The `ID` listed is the key for PostCSS Preset Env configuration in your project.
 | `dir-pseudo-class` | `:dir` Directionality Pseudo-Class | [example](https://preset-env.cssdb.org/features/#dir-pseudo-class) | [docs](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-dir-pseudo-class#readme) |
 | `display-two-values` | Two values syntax for `display` | [example](https://preset-env.cssdb.org/features/#display-two-values) | [docs](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-normalize-display-values#readme) |
 | `double-position-gradients` | Double Position Gradients | [example](https://preset-env.cssdb.org/features/#double-position-gradients) | [docs](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-double-position-gradients#readme) |
+| `float-clear-logical-values` | Logical Values in float and clear | [example](https://preset-env.cssdb.org/features/#float-clear-logical-values) | [docs](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-logical#readme) |
```
Antonio-Laguna, line 22: The rest is missing from CSSDB I think I jumpstarted on this too soon. 

Once we start to release and we make V8 go into `main` it should be safer to add the rest

Antonio-Laguna:
@romainmenke I was hoping tests would start to pass but it looks like base-plugin had other plans?  😂

plugin-packs/postcss-preset-env/src/lib/shared-options.mjs
romainmenke: Correct, this function had more parts when `importFrom` and `exportTo` still existed.

This simplified version is better for now 🎉

plugin-packs/postcss-preset-env/src/lib/get-options-for-browsers-by-feature.mjs
romainmenke: At first I disagreed because I thought we had a thing for shared plugin options.
But these options are not intended for all plugins, only a select few. This is different from `preserve`.

So I think this is indeed the best way to do this and the best place for this code 🎉

romainmenke:
> I was hoping tests would start to pass but it looks like base-plugin had other plans

I think something went wrong with `package-lock.json`.
No idea what 🤷 

CI should be all green now 🎉

plugin-packs/postcss-preset-env/FEATURES.md
romainmenke: I think we can do a minor version in CSSDB where we add the plugins as polyfills?

romainmenke: On second thought, let's leave this as if for now.
`FEATURES.md` can be updated without requiring a release.

Doing `v8` first and then adding the links makes more sense :)

romainmenke (approved):
LGTM!

Thank you for making it all work together 🎉

PR #799, opened 2023-01-24
design tokens : fix documentation and unit conversion

romainmenke:


plugins/postcss-design-tokens/src/data-formats/style-dictionary/v3/value.ts
```diff
@@ -131,15 +131,17 @@ function formatFloat(value: number): string {
...
 		}
 
-		if (fixedPrecision[i] !== '0') {
-			fixedPrecision = fixedPrecision.slice(0, i + 1);
+		if (fixedPrecision[i] === '0') {
+			fixedPrecision = fixedPrecision.slice(0, i);
 			continue;
 		}
+
+		break;
```
romainmenke, line 144: Not sure what happened when I initially wrote this.
It was obviously incorrect.

Noticed it when the docs updated and I saw `1.1` where I expected `1.125`

Antonio-Laguna (approved):
This LGTM! Thanks for fixing this!

PR #800, opened 2023-01-24
chore(postcss-logical-viewport-units): move @csstools/css-tokenizer to dependencies

simPod:
`@csstools/css-tokenizer` is used directly in src so it has to be in dependencies. 

E.g. Yarn PnP complains about missing dependency on v8.0.0 since it's not declared now.


romainmenke (approved):
Thank you @simPod !

especially for making sure CI is all green 🎉

PR #806, opened 2023-01-26
Compatible with ts NodeNext module resolution

otakustay:
In `"moduleResolution": "NodeNext"` it requires a `types` property in `exports` field, otherwise we need `./dist/index.d.mts` instead of `.d.ts`

### DIFF
--- a/plugin-packs/postcss-preset-env/package.json
+++ b/plugin-packs/postcss-preset-env/package.json
@@ -32,6 +32,7 @@
 		".": {
 			"import": "./dist/index.mjs",
 			"require": "./dist/index.cjs",
+			"types": "./dist/index.d.ts",
 			"default": "./dist/index.mjs"
 		}
 	},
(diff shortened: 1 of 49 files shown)


romainmenke:
Hi @otakustay !

Thank you for this change 🙇 
We were able to confirm the issue and that your change works.
We've added an extra test to make sure this won't regress in the future : https://github.com/csstools/postcss-plugins/pull/807

This change however isn't complete yet.
Not only `postcss-preset-env` needs this change, but each plugin that is part of the `preset-env` pack needs it too.

Is this something you are interested in adding?
Or shall we take it from here?

otakustay:
I updated my patch to modify every `package.json` inside `plugins` directory, but I don't familiar with adding e2e tests for them, if any more tasks needed, I'll continue

romainmenke (approved):
This is awesome, thank you so much @otakustay

PR #808, opened 2023-01-26
release plan helper command

romainmenke:
Context :

We have a whole lot of packages and now that we are also authoring our own tokenizers and parsers the dependency graph can easily become complex.

Publishing everything is a lot of work when you have to get these things right consistently :
- correct version increment
- correct `CHANGELOG.md` edit
- correct publish order
- what to publish
- ...

-----

What if we use the `CHANGELOG.md` as a signal to create and execute a release plan.

Each `CHANGELOG.md` for an unreleased update has : `Unreleased (patch|minor|major)`.
Anything else is ignored.

When publishing you first run : `npm run release-plan --dry-run`.
This will tell you :
- what will be released
- in what order
- the increment

```
Release plan:
  - @csstools/css-tokenizer (patch)
  - @csstools/selector-specificity (patch)
  - @csstools/postcss-progressive-custom-properties (patch)
  - css-blank-pseudo (patch)
  - ...
```

When publishing for real you run : `npm run release-plan`.

------

How it works :

- list all workspaces
- build a dependency graph
- check the `CHANGELOG.md` file for each
- create a list of only those packages that have up to date local dependencies
- increment the version
- re-generate docs
- publish
- commit the version increment

The `publish` sub process inherits stdio from the main process.
So a prompt for 2FA will just be passed to you.

After all releases :
- increment the versions of freshly released dependencies in other packages
- run `npm install` to update the lock file

These changes are not committed to make it easier to review what has changes.

_"up to date local dependencies" means that this package does not depend on another local package that needs to be released_

------

Thoughts?


.github/bin/release-plan/date-format.mjs
```diff
@@ -0,0 +1,32 @@
+export function nowFormatted() {
+	const now = new Date();
+
+	switch (now.getMonth()) {
+		case 0:
+			return `January ${now.getUTCDate()}, ${now.getUTCFullYear()}`;
```
romainmenke, line 6: There definitely is something in node that just gives the month, I didn't spend time looking.

Antonio-Laguna:
I'd say let's give it a spin! 

THANKS for adding this :) 

I think it's missing the `docs --if-exist`  part for packages that have version number but other than that is looking solid!

.github/bin/release-plan/release-plan.mjs
```diff
@@ -0,0 +1,84 @@
...
+	// Update the documentation
+	await updateDocumentation(workspace.path);
+
+	// Publish to npm
+	await publish(workspace.path, workspace.name);
+
+	// Commit changes
+	await commit(newVersion, workspace.path, workspace.name);
+
+	// TODO : remove this after the script proves to work ok.
```
romainmenke, line 82: I want to take it slow and try it one by one first.
So the script will stop after a single package.

If that works you can simply remove these lines and it will do the whole batch.

romainmenke:
_forgot to mention explicitly_

Because packages with unreleased dependencies are skipped you will need to run it multiple times. In theory I can let it bump dependency versions and continue from there.

But I think it is better if it only releases things that were prepared by a human and doesn't start to alter packages on it's own.

for example : 
1. release an update for `@csstools/css-tokenizer`
2. go to `@csstools/css-parser-algorithms`
3. bump the version for dependency `@csstools/css-tokenizer` 
4. release an update for `@csstools/css-parser-algorithms`

I can make it capable of this.
But it feels a bit too magical at this this time :D

The downside is that we will still need to alter a few things manually in between chunks of releases.

PR #809, opened 2023-01-26
todo-or-die

romainmenke:
replaces : https://github.com/csstools/postcss-plugins/pull/757

readme : https://github.com/csstools/postcss-plugins/tree/todo-or-die--generous-long-eared-owl-f30a120073/plugins/postcss-todo-or-die#readme


plugins/postcss-todo-or-die/docs/README.md
```diff
@@ -0,0 +1,97 @@
...
+| `if` | when `true` or `unknown` | when `false` |
+| `not` | when `false` or `unknown` | when `true` |
+| `before-date` | when "now" is before the date | when "now" is after |
+| `browserslist` | when browsers match those of your project | when no browsers match |
+| `issue-open` | when the issue is open | when closed |
+
+```pcss
+@todo-or-die if(10 > 8);
+@todo-or-die not(10 < 8);
+@todo-or-die before-date(2006 01 31); /* year, month, day */
```
Antonio-Laguna, line 76: Probably a silly suggestion but should the commas in the comment be removed? Unless it accepts them

plugins/postcss-todo-or-die/src/match/issue-open.ts
```diff
@@ -0,0 +1,104 @@
...
+	const cacheKey = `github-issue-open:${repository}:${issue}`;
+	let state;
+
+	if (!cache.has(cacheKey)) {
+		try {
+			state = await fetch(repository, issue);
+		} catch (_) {
+
+			// Any network errors are ignored.
+			// We do not want builds to fail when GitHub is offline.
```
Antonio-Laguna, line 14: Nice one!

plugins/postcss-todo-or-die/src/match/issue-open.ts
```diff
@@ -0,0 +1,104 @@
...
+		const options = {
+			headers: {
+				'User-Agent': 'PostCSS TODO or Die',
+			},
+			timeout: 15000,
+		};
+
+		if (process.env.GITHUB_TOKEN) {
+			options.headers['Authorization'] = `Bearer ${process.env.GITHUB_TOKEN}`;
+		}
```
Antonio-Laguna, line 50: I feel this should be documented

plugins/postcss-todo-or-die/src/match/issue-open.ts
```diff
@@ -0,0 +1,104 @@
...
+		const request = https.request(url, options, (response) => {
+			if (rejected) {
+				return;
+			}
+
+			if (response.statusCode !== 200) {
+				rejected = true;
+				reject(new Error('Failed to fetch issue status'));
+				return;
+			}
```
Antonio-Laguna, line 61: I was wondering what happens if you don't have visibility to an issue because you're lacking the token or whatever I feel this should at least be surfaced somehow since you might think it's working but perhaps it isn't

Antonio-Laguna (commented):
@romainmenke this is looking great! I left some questions even if this wasn't ready to be reviewed :P

plugins/postcss-todo-or-die/src/match/issue-open.ts
romainmenke: I am going to remove the "issue open" feature entirely :)

`todo-or-die` for Rust has this feature, but you compile Rust much less frequently than you build CSS during development. When working in Rust or Go I don't use a file watcher to re-compile.

Doing so many builds so frequently is bound to cause false errors because it depends on the network. You also should be able to work offline without seeing errors.

I think it makes more sense to remove this, and maybe later add something that allows people to easily add custom checks.

plugins/postcss-todo-or-die/src/match/issue-open.ts
romainmenke: it's gone :)

plugins/postcss-todo-or-die/src/match/issue-open.ts
romainmenke: it's gone :)

plugins/postcss-todo-or-die/docs/README.md
romainmenke: Good catch! edited

romainmenke:
>  is looking great! I left some questions

Thank you for those!

I was hoping to find some solution to the "issue-open" feature, but as I said above I don't see a way to make work well within a CSS toolkit.

I'd rather streamline the plugin a bit now and ship it 🚀

Antonio-Laguna (approved):

PR #810, opened 2023-01-27
text-decoration-shorthand: multiple line values

romainmenke:
```css
text-decoration: 5px underline overline line-through wavy pink;
```

This is valid and I didn't know about it.
From what I can observe in browsers the line keywords need to come after each other.

I've also added a few more tests for things that should be ignored by this plugin.
We don't want it to start processing unknown future CSS.


Antonio-Laguna (approved):
LGTM! I didn't know this was possible, so many values!

PR #811, opened 2023-01-27
nesting : specification rework

romainmenke:
- [x] not breaking, everything that worked before will still work
- [x] support for relative selectors
- [x] browser tests
- [x] documentation

I don't want to give a warning on anything that would be valid in today's plugin.
People tend to panic a bit when they get these warnings.

I propose that we do this :
1. ship something that allows authors to write "standard nesting"
2. let advocates, MDN, blog post writers do their thing
3. authors are familiar with "standard nesting"
4. new release with warnings for `@nest`
5. remove `@nest` completely in a next major
6. optionally switch to a much simplified plugin that only uses `:is()` in a next major. (the work we did in the experimental plugin)

-----

Technically this is a breaking change.
But adoption of the current major is still tiny and it is only breaking for the order of declarations after nested rules. This is a bad practice anyway and very few people do this.

I don't expect anyone to be impacted by this at this time.


romainmenke:
I think we can and should ship this now.
Browsers haven't finished implementing yet, so there is some risk but I think either the current version of the specification will remain largely as is or it will change dramatically. No one wants the latter.

Advocates and DevRel are starting to communicate this feature, now that there are working prototypes in alpha/preview versions of browsers.

I don't want us to miss this moment by being to cautious.
Shipping or not shipping this update has no impact on us needing to do a massive breaking change in the small chance that the specification is re-written.

I would have preferred if everyone waited a bit more until some last open issues were resolved, but that is not going to happen.

Antonio-Laguna:
@romainmenke thanks for putting this together. I agree with releasing this!

I think we should communicate this though so there's a clear path ahead and nobody is caught off-guard. Is there a good resource for the differences? I'm thinking about putting together a page that outlines this clearly, both the differences with what it was and also what's expected in the short/mid-term. Any thoughts on this?

Happy to start this myself if you have some pointers handy

Antonio-Laguna (approved):

PR #813, opened 2023-01-28
playground

romainmenke:
_bringing it back is finally possible and relatively easy because all plugins work in browsers_


sites/postcss-preset-env/src/static/js/playground.js
```diff
@@ -0,0 +1,197 @@
...
+
+// hack to make browserslist work in a browser. (irony...)
+self.process = { env: {} };
+
+const currentConfig = {
+	browsers: ['> 0.2% and not dead'],
+	minimumVendorImplementations: 0,
+	stage: 2,
+	preserve: null,
+};
```
Antonio-Laguna, line 18: Should we add the logical options too? Could do that myself after too

Antonio-Laguna (approved):
LGTM! Thanks for bringing this back!

sites/postcss-preset-env/src/static/js/playground.js
romainmenke: We should!

PR #814, opened 2023-01-29
try edge functions

romainmenke:


Antonio-Laguna:
Did it work though? 😂

romainmenke:
It did work, but I couldn't do anything meaningful.
Only Deno modules work in edge functions and there are no packages for Deno, or at least not the once I needed.

I wanted to run PostCSS and Preset Env in an edge function.
This would enable sending smaller, more modern CSS to modern browser versions.

But it became overly complicated, not something we can easily demo and explain to users.

PR #818, opened 2023-01-30
Adding Logical Options to the Playground

Antonio-Laguna:
This adds the inlineDirection and blockDirection options to the controls. 

Note that when changing block to right-to-left for example nothing seems to happen. The error, despite the `.catch` goes uncaught. I think this is mostly OK but wanted to point it out.


romainmenke (approved):
Awesome!

Thank you for this 🎉 

------

> Note that when changing block to right-to-left for example nothing seems to happen. The error, despite the .catch goes uncaught. I think this is mostly OK but wanted to point it out.

I forgot about the error path 🙈 
I've made a few tweaks so that the output view resets in cases of errors. This gives some signal to users that the config isn't going to work.

I've also rewritten the process function so that alle exceptions are caught.

Issue #827, opened 2023-02-01
[postcss-is-pseudo-class] :not(:is()) producing incorrect CSS


thewilkybarkid:
### Bug description

I'm looking to upgrade to postcss-preset-env 8 (https://github.com/PREreview/prereview.org/pull/616). Some of the failures there are the result of specificities changing to match the specs, which I can resolve. I can't, however, seem to resolve the use of `:not(:--some-custom-selector)`: it's now changed to separate rather than chained selectors (resulting in incorrect CSS rather than one with inflated specificity).

### Source CSS

```css
:not(:is(h1, h2, h3, h4, h5, h6)) {
  color: red;
}
```

### Expected CSS

```css
:not(h1):not(h2):not(h3):not(h4):not(h5):not(h6) {
  color: red;
}
```

### Actual CSS

```css
:not(h1) {
  color: red;
}
:not(h2) {
  color: red;
}
:not(h3) {
  color: red;
}
:not(h4) {
  color: red;
}
:not(h5) {
  color: red;
}
:not(h6) {
  color: red;
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

Yes

### What plugin are you experiencing this issue on?

PostCSS Preset Env, PostCSS Is Pseudo Class

### Plugin version

postcss-preset-env 8.0.1

### Node Version

18.13.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@thewilkybarkid  Thank you for this report!
We will look into this!

romainmenke:
@thewilkybarkid We have just released an update that should fix this.
You can try it out on the playground

```css
@custom-selector :--headings h1, h2, h3, h4, h5, h6;

:not(:--headings) {
  color: red;
}

/* becomes */

:not(h1):not(h2):not(h3):not(h4):not(h5):not(h6) {
  color: red;
}
```

https://preset-env.cssdb.org/playground/

<img width="1470" alt="Screenshot 2023-02-02 at 10 38 03" src="https://user-images.githubusercontent.com/11521496/216287586-2f059899-6df3-4ba6-a115-a47f8096722e.png">


---

can you let us know if everything is working as expected again?

[...]

PR #828, opened 2023-02-01
postcss-is-pseudo-class : fix `:not` pseudo interaction

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/827

`:not(is(a, b, c))` can not be rewritten as a forgiving selector.
It is either the wrong selector (current behavior) or an unforgiving list (new behavior).

------

The only reason to split all these selectors so aggressively is to be a 100% sure that every rule will be valid or invalid individually.

But we actually know for a lot of these that they will always be valid.
Only new pseudo's, nesting or unknown combinators can trigger invalid selectors.
To be super cautious I also included case insensitive attribute matching.

We can tweak the logic as needed.


Antonio-Laguna (approved):
This LGTM! Thanks for doing this!

PR #830, opened 2023-02-02
bump site

romainmenke:


sites/postcss-preset-env/src/static/js/playground.js
```diff
@@ -234,7 +234,7 @@ for (const control of Object.values(controls)) {
 			blockDirection: controls.blockDirection.value || 'top-to-bottom',
 		};
 
-		processCss(inputState.doc, currentConfig).then((output) => {
+		processCss(inputView.state.doc, currentConfig).then((output) => {
```
romainmenke, line 237: After a config state the playground would render incorrectly.
It would use the demo CSS as the input, not whatever a visitor is currently trying.

PR #835, opened 2023-02-04
Reduce the amount of generated fallback css (batch 1)

romainmenke:
There are multiple issues, we can't fix them all but we can do a little bit more here and there to improve things a lot overal.

1. PostCSS re-visits nodes if a child node was modified, inserted, ..
2. CSS Authors might write their own fallbacks
3. Multi-feature values generate a bunch of intermediate fallbacks 

1: We can keep track of which nodes we generated fallbacks for and skip them
2: We can look at the previous declaration
3: Same as 2

For users this change must be unnoticeable.

----

example issue :

```css
:is(input, button):is(:hover, :focus) {
	color: rebeccapurple;
}

/* becomes */

input:hover, input:focus, button:hover, button:focus {
	color: #639;
	color: rebeccapurple;
}
input:hover, input:focus, button:hover, button:focus {
	color: #639;
	color: #639;
	color: rebeccapurple;
}
:is(input, button):is(:hover, :focus) {
	color: #639;
	color: rebeccapurple;
}
```


Antonio-Laguna (approved):
This is so clever! 

Love this a lot and I'm glad this improves the output!

Issue #838, opened 2023-02-06
Nested var() breaks when using a plugin


kay32:
### Bug description

The plugin handles variables that are not found in :root and then the nested var() will not be output correctly.

### Source CSS
```css
:root {
    --var-color: red;
}

.example {
    background: var(--var-other, var(--var-other-global));
    color: var(--var-color);
}
```

### Expected CSS

```css
:root {
    --var-color: red;
}

.example {
    background: var(--var-other, var(--var-other-global));
    color: red;
}
```

### Actual CSS

```css
:root {
    --var-color: red;
}

.example {
    background: var(--var-other-global);
    color: red;
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

Yes

### Extra config

```json
{"preserve": false}
```

### What plugin are you experiencing this issue on?

PostCSS Custom Properties

### Plugin version

13.1.1

### What OS are you experiencing this on?

Linux

### Node Version

18.14.0

### Validations

- [x] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
@kay32 Thank you for reaching out.
This output is actually technically correct but it isn't helpful.

If the fallback isn't another variable the output makes much more sense :

```css
:root {
    --var-color: red;
}

.example {
    background: var(--var-other, green);
    color: var(--var-color);
}
```

becomes :

```css
.example {
    background: green;
    color: red;
}
```

This issue illustrates really well why you should avoid setting `preserve` to `false`.
With `preserve: true` :

```css
:root {
	--var-color: green;
}

.example {
	background: var(--var-other-unknown);
	background: var(--var-unknown-in-root, var(--var-other-unknown));
	color: green;
	color: var(--var-color);
}
```

--------

I think we can special case fallbacks that contain `var()` and ignore those.

kay32:
Thanks for the answer, I think this will solve my problem.

It's a little misleading: `Note: This plugin only processes variables that are defined in the :root or html selector.`
It would be great to have an option for this behavior so that the plugin completely ignores variables that are not in: root, html and does not replace them with a fallback value.

romainmenke:
I am trying a slightly different perspective :)

[...]

PR #839, opened 2023-02-06
custom-properties : do not apply fallback values when these contain unresolvable custom properties

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/838

----

This looks like a more radical rewrite than it actually is.

It uses existing logic from custom-media and custom-selectors to handle cyclic dependencies between custom properties.

That change makes it possible to use a simpler walk.
What actually happens in the walker is largely the same as it was before but we can now walk inside out.

Walking inside out was key to correctly detecting that a variable fallback is another variable and that it is unresolvable.


Antonio-Laguna (approved):
Thanks for this!

Yeah looks like a rewrite yet tests don't lie :P 
Feels clear and definitely good to me!

PR #840, opened 2023-02-06
Reduce the amount of generated fallback css (batch 2)

romainmenke:
same as before : https://github.com/csstools/postcss-plugins/pull/835


Antonio-Laguna (approved):
LGTM! Awesome improvements here!

PR #841, opened 2023-02-06
PostCSS Global Data

### LINKED ISSUE #834: I'm using Astro. Being able to use `importFrom` was a big deal. What is the workaround besides not upgrading to postcss-preset-env v8?
I'm using Astro. Being able to use `importFrom` was a big deal. What is the workaround besides not upgrading to postcss-preset-env v8? 

My postcss config: 
```js
const postcssPresetEnv = require("postcss-preset-env");

module.exports = {
  plugins: [
    postcssPresetEnv({
      stage: 0,
      features: {
        "custom-media-queries": {
          importFrom: ["./src/styles/common/queries.css"],
        },
      },
      autoprefixer: true
    })
};
```

_Originally posted by @djmtype in https://github.com/csstools/postcss-plugins/issues/822#issuecomment-1415918516_

Antonio-Laguna:
Fixes #834

This creates a plugin that injects Global Data into the CSS. The plugin only injects nodes into the root to then, once it exits, remove them having allowed other plugins to use this data to generate output.

The tests test it specifically with custom-media and ensures that no rules are staying after it finished


plugins/postcss-global-data/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+						}
+
+						const newCSS = parseImport(file, importedFiles);
+
+						if (newCSS) {
+							newCSS.each((node) => {
+								root.append(node);
+								importedCSS.add(node);
+							});
+						}
```
romainmenke, line 40:
```suggestion
						const newCSS = parseImport(file, importedFiles);
						if (!newCSS) {
							return;
						}

						newCSS.each((node) => {
							root.append(node);
							importedCSS.add(node);
						});
```

early return

plugins/postcss-global-data/test/basic.css
```diff
@@ -0,0 +1,35 @@
...
+	body {
+		order: 1;
+	}
+}
+
+@media (--mq-b) {
+	body {
+		order: 2;
+	}
+}
```
romainmenke, line 11: Maybe sufficient to use simpler `@custom-media` examples?

Nothing incorrect about these, but a simpler example is less likely to require changes if the `postcss-custom-media` produces slightly different output.

romainmenke (commented):
This is awesome!

_Looking at the simplicity I am a bit surprised that no one created exactly this before now. The feature felt really complex to me because it was always entangled in other plugins._

plugins/postcss-global-data/test/basic.css
Antonio-Laguna: This is a great suggestion, I've simplified these even further

plugins/postcss-global-data/docs/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+[<humanReadableName>] lets you inject data into PostCSS that then gets removed from the final CSS. This is useful for 
+plugins might use data to generate CSS. For example, in the case of CSS Modules with [PostCSS Custom Media](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-custom-media), rules
+are usually not imported by every single file so Custom Media cannot generate any code based on the available data. By
+providing a list of files, this plugin will inject the data into the CSS so that the plugin can generate the code.
+
+It is important that [<humanReadableName>] is used before the plugin that actually uses the data. The plugin runs 
+syncronously so it will inject the data before the other plugin runs.
+
+Please note that [<humanReadableName>] does not produce any CSS. It only injects data into PostCSS so that other plugins
```
romainmenke, line 27:
```suggestion
[<humanReadableName>] lets you inject CSS that is removed again before the final output. This is useful for  plugins that use global CSS as data.

For example, in the case of CSS Modules with [PostCSS Custom Media](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-custom-media), rules are usually not imported by every single file, so PostCSS Custom Media cannot generate fallbacks.
By providing a list of files, this plugin will inject the global CSS as data so that PostCSS Custom Media can generate fallbacks.

It is important that [<humanReadableName>] is used before the plugin that actually needs the data.

Please note that [<humanReadableName>] does not add anything to the output of your CSS. It only injects data into PostCSS so that other plugins
```

Made a few edits. I tried to use `global` `data` and `css` as much as possible in the context and meaning that we intend with this plugin :D

plugins/postcss-global-data/test/examples/example.css
```diff
@@ -0,0 +1,7 @@
+.foo {
```
romainmenke, line 1: I don't think the examples are used?
Can we delete these?

plugins/postcss-global-data/package.json
```diff
@@ -0,0 +1,77 @@
...
+	"homepage": "https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-global-data#readme",
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/csstools/postcss-plugins.git",
+		"directory": "plugins/postcss-global-data"
+	},
+	"bugs": "https://github.com/csstools/postcss-plugins/issues",
+	"keywords": [
+		"postcss-plugin"
+	],
```
romainmenke, line 69: Maybe a few more keywords?
For the todo or die plugin I added `productivity`, I think that can also be used here.

romainmenke (commented):
Still awesome!

Left some final comments :)

plugins/postcss-global-data/test/examples/example.css
Antonio-Laguna: Yep will do!

.github/bin/generate-docs/readme.mjs
```diff
@@ -6,7 +6,13 @@ const template = await fsp.readFile(path.join('docs', './README.md'), 'utf8');
...
 const packageJSONInfo = JSON.parse(await fsp.readFile('./package.json', 'utf8'));
 
-const exampleFilePaths = await fsp.readdir(path.join('test', 'examples'));
+let exampleFilePaths = [];
+
+try {
+	exampleFilePaths = await fsp.readdir(path.join('test', 'examples'));
+} catch(error) {
+	// No examples
+}
```
Antonio-Laguna, line 15: Made examples non-mandatory

romainmenke: nice

romainmenke (approved):
LGTM 🚀

PR #843, opened 2023-02-08
Adding CSS Modules disclaimer to custom-media

Antonio-Laguna:
I think this will be mostly a copy from plugin to plugin but I thought I'd kickstart the revision on one of them and then carry on. 

Check [PostCSS Custom Media updated README](https://github.com/csstools/postcss-plugins/blob/feature/global-data-docs/plugins/postcss-custom-media/README.md#css-modules)

--- 

Additionally, please check the [Wiki updates](https://github.com/csstools/postcss-plugins/wiki/PostCSS-Preset-Env-8/_compare/511fd78d6b94b7f5561251d041a9928cb9bc26e0...8007d5f2ac05fa9277ead7ee42a93e56744e5cb8). I want to make sure it's clear!


plugins/postcss-custom-media/docs/README.md
```diff
@@ -78,6 +78,33 @@ An example of a very complex (and artificial) use-case :
...
 <envSupport>
 
+## CSS Modules
+
+If you're using CSS Modules, you'll probably notice that custom media queries are not being resolved. This happens
+because each file is processed separately so unless you import the custom media query definition in each file, they
+won't be resolved.
+
+To overcome this, we recommend using the [PostCSS Global Data](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-global-data#readme)
+plugin which allows you to pass a list of files that will globally be available. The plugin won't inject any extra code
```
romainmenke, line 88:
```suggestion
plugin which allows you to pass a list of files that will be globally available. The plugin won't inject any extra code
```

romainmenke (commented):
This is great!

Should we place this at the end of the readme?
After the plugin options.

More people that need to know the general plugin option than people that use css modules :)

plugins/postcss-custom-media/README.md
```diff
@@ -123,6 +123,33 @@ instructions for:
 - [Gulp](INSTALL.md#gulp)
 - [Grunt](INSTALL.md#grunt)
 
+## CSS Modules
+
+If you're using CSS Modules, you'll probably notice that custom media queries are not being resolved. This happens
```
mihkeleidast, line 128: Is it correct for this documentation to be tied to CSS Modules specifically? It is an issue in whatever system where CSS files are modular, i.e one might not b using CSS Modules, rather they might just be importing small CSS files in multiple places in JS. As another example, we are using vanilla-extract, which is not exactly CSS Modules, but still has the same issue and solution.

romainmenke: Is there a good term that can be used to describe the issue or the way CSS is processed so that as many users as possible understand that the solution could apply to them?

Antonio-Laguna: @mihkeleidast this is a good point and perhaps this is just a misunderstanding on my part. CSS Modules is not a specific package but rather a philosophy or way to handle CSS.

See https://github.com/css-modules/css-modules

If there's a better term I'm all up for changing it

romainmenke: Maybe inverse it : `Modular CSS processing`

This covers `CSS Modules` but also other things like `postcss-loader` in Webpack, `vanilla-extract`, ...

mihkeleidast: I would also use "modular css" as the general term, and then list out some of the examples of that in the next paragraph, so users can ctrl+f their specific term.

plugins/postcss-custom-media/docs/README.md
```diff
@@ -97,5 +97,32 @@ is preserved. By default, it is not preserved.
 <example.preserve.expect.css>
 ```
 
+## Modular CSS Processing
+
+If you're using Modular CSS such as, CSS Modules, `postcss-loader` or `vanilla-extract to name a few, you'll probably
```
romainmenke, line 102:
```suggestion
If you're using Modular CSS such as, CSS Modules, `postcss-loader` or `vanilla-extract` to name a few, you'll probably 
```

typo :)

Antonio-Laguna: Thanks!!

romainmenke (approved):
LGTM!

PR #846, opened 2023-02-09
postcss-global-data : set postcss 'from' option

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/680#issuecomment-1424401683


plugins/postcss-global-data/src/index.ts
```diff
@@ -24,13 +24,13 @@ const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 			let importedCSS = new Set<ChildNode>();
 
 			return {
-				Once: (root) => {
+				Once: (root, { postcss }) => {
```
romainmenke, line 27: I missed this before, but we can "extract" PostCSS from here.
This allows us to avoid importing PostCSS directly.

e2e/browserslist/package-json/index.mjs
```diff
@@ -2,7 +2,7 @@ import assert from 'assert';
 import plugin from 'postcss-preset-env';
 import postcss from 'postcss';
 
-const result = await postcss([plugin()]).process(':any-link { color: blue; }').css;
+const result = (await postcss([plugin()]).process(':any-link { color: blue; }')).css;
```
romainmenke, line 5: unrelated, but I noticed that this test never ran and was broken when adding the extra test for all these plugins together

Antonio-Laguna (approved):
Love the interop test! LGTM!

Issue #848, opened 2023-02-10
Document custom syntax things for stylelint


romainmenke:
Stylelint can now enforce value syntax and this is extensible.

For the custom things we have we should document and test stylelint config : https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/#propertiessyntax--property-syntax-

PR #849, opened 2023-02-10
add docs for stylelint rule syntax

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/848

### DIFF
--- a/plugins/postcss-conditional-values/docs/README.md
+++ b/plugins/postcss-conditional-values/docs/README.md
@@ -73,6 +73,25 @@ color: csstools-if(--a-condition yellow else red);
 csstools-if(<custom-property-name> <declaration-value> else <declaration-value>);
 ```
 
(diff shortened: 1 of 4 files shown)


plugins/postcss-conditional-values/docs/README.md
```diff
@@ -73,6 +73,25 @@ color: csstools-if(--a-condition yellow else red);
...
+Setting the correct configuration for this rule makes it possible to check even non-standard syntax.
+
+```js
+	// Disallow unknown values for properties within declarations.
+	'declaration-property-value-no-unknown': [
+		true,
+		{
+			propertiesSyntax: {
+				color: '| csstools-if( <custom-property-name> <\'color\'> else <\'color\'> )',
+				'background-color': '| csstools-if( <custom-property-name> <\'background-color\'> else <\'background-color\'> )',
```
romainmenke, line 88: I hope the repetition makes it clear what is going here.

I don't think we can or should try to document the CSS value syntax notation. So I am only providing a hint for users who already know enough to use this correctly.

PR #851, opened 2023-02-11
fix file watching in postcss cli

romainmenke:
By emitting a custom message in the exact same way as plugins like `postcss-import` the PostCSS CLI watch mode will pick up on these dependencies.

If you make a change in your global data or in your design token files the CSS task will re-run.

There might be other tools that listen to the same signals.


Antonio-Laguna (approved):
LGTM! I always forget about this!

sacom90:
https://store.zidlix.com/

PR #852, opened 2023-02-11
css-calc

romainmenke:
For relative color syntax we need to be able to solve calculations written in CSS.

https://www.w3.org/TR/css-color-5/#:~:text=The%20new%20relative%20color%20syntax,possibly%20modified%20with%20math%20functions).

```
--blue-into-gray: rgb(from blue
                    calc(r * .3 + g * .59 + b * .11)
                    calc(r * .3 + g * .59 + b * .11)
                    calc(r * .3 + g * .59 + b * .11));
```

Untitled Diagram (1)

This diagram illustrates how everything will fit together in the end.

This re-structure is needed because `color-mix` and relative color syntax both need to be able to parse and convert all possible color notations.

They also both need to be able to solve `calc()`.
This change adds the `calc` package.

_At the moment I see `Color Helpers` and `Color Parser` as two distinct packages.
One has all the maths and logic for colors. The other is able to take any CSS value and give back a color._

----

Most of the code in this PR is tests.
There are just so many math functions and units.
Needs a lot of tests to have sufficient coverage.


packages/css-calc/test/basic/test.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+assert.strictEqual(
+	convert('foo(something calc(2 * r + 7 * 5) something-else)', new Map([['r', 3]])),
+	`foo(something ${(2 * 3 + 7 * 5).toString()} something-else)`,
+);
+
+assert.strictEqual(
+	convert('rgb(rgb(50 100 150) calc(r * 2) calc(g / 2) calc(b / 3))', new Map([['r', 50], ['g', 100], ['b', 150]])),
+	`rgb(rgb(50 100 150) ${(50 * 2).toString()} ${(100 / 2).toString()} ${(150 / 3).toString()})`,
+);
```
romainmenke, line 42: This is starting to work 🎉

plugins/postcss-stepped-value-functions/test/basic.expect.css
```diff
@@ -41,6 +41,12 @@
 	right: 8px;
 }
 
+.unit {
+	top: 37.79527559055118px;
```
romainmenke, line 45: It looks weird to round something and to get a decimal, but this is rounded to `1cm`.

Screenshot 2023-02-14 at 19 17 00

plugins/postcss-trigonometric-functions/test/basic.preserve-true.expect.css
```diff
@@ -5,12 +5,11 @@
 	order: sin(15 * 3deg); /* 45deg */
 	order: 0.70711;
 	order: sin(.125turn); /* 45deg */
-	order: 0.70711;
 	order: sin(1 / 8turn); /* 45deg */
 	order: 0.70711;
 	order: sin(50grad); /* 45deg */
 	order: 0.70711;
-	order: sin(10 + 4 * 10grad); /* 45deg */
+	order: sin((1 + 4) * 10grad); /* 45deg */
```
romainmenke, line 12: `10 + 40grad` is invalid.

plugins/postcss-trigonometric-functions/test/basic.expect.css
```diff
@@ -2,7 +2,7 @@
 	order: 0.70711; /* 45deg */
 	order: 0.70711; /* 45deg */
 	order: 0.70711; /* 45deg */
-	order: 0.70711; /* 45deg */
+	order: sin(1 / 8turn); /* 45deg */
```
romainmenke, line 5: I think this is invalid.
But `1turn / 8` would not be invalid.

I can update the source file.

plugins/postcss-stepped-value-functions/test/basic.expect.css
```diff
@@ -41,11 +41,17 @@
...
+	bottom: 940px;
+	left: 0.035s;
+}
+
 .invalid {
 	font-size: mod(18px, 5rem);
 	width: rem(var(--foo), 5px);
 	padding: mod(Fo0px, 5px);
-	left: mod(1px, 0px);
+	left: calc(NaN * 1px);
```
romainmenke, line 54: This is a `NaN` value that still preserves the original unit.
It has no use, but it is technically correct.

plugins/postcss-trigonometric-functions/test/wpt.expect.css
```diff
@@ -98,7 +98,7 @@
 	order: atan2(1deg * ));
 	order: atan2(1deg / ));
 	order: atan2(1deg 2deg));
-	order: 26.57deg);
+	order: atan2(1deg, , 2deg));
```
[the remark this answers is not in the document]
Antonio-Laguna: Agreed!

packages/css-calc/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+### Options
+
+#### `precision` :
+
+The default precision is fairly high.
+It aims to be high enough to make rounding unnoticeable in the browser.
+
+You can set it to a lower number to suite your needs.
```
Antonio-Laguna, line 70: minor typo

```suggestion
You can set it to a lower number to suit your needs.
```

romainmenke: Good catch! 🙇

Antonio-Laguna (approved):
This looks great! I love how everything ties together and that we're able to abstract away logic to be reused in a seemingly unrelated plugin! Kudos for the diagram! Very clear!! 🙇🏻

The only feedback I'd leave is to try to convert some of the checks that are repeated into reusable functions so the code is DRYer and clearer. An example:

```js
	if (
		!(
			aToken[0] === TokenType.Dimension ||
			aToken[0] === TokenType.Number ||
			aToken[0] === TokenType.Percentage
		)
	) {
		return -1;
	}
```

In any case, this is not a blocker and can definitely be something that happens afterwards!

PR #853, opened 2023-02-12
postcss-nesting : ampersand everywhere and mixin related fixes

romainmenke:
This adds support for:

```css
& .foo{}

/* becomes */

:scope .foo {}
```

https://drafts.csswg.org/css-nesting/#nest-selector

> When used in any other context, it represents the same elements as [:scope](https://drafts.csswg.org/selectors-4/#scope-pseudo) in that context (unless otherwise defined).

-------

After you mentioned mixins I had to double check to make sure I hadn't broken it with the latest changes :) But yes, I did.

I don't think `@mixin` is still implementable in browsers because of how nesting now defines the order. But I don't want to advocate for `@mixin` with the CSSWG as I don't use or need mixins.

This change is more of a best effort to support "something".


Antonio-Laguna:
This LGTM, is there any caveat right now with mixins? My only point with mixins is not to position ourselves in a way that we close the ecosystem. If you use postcss-nesting you should still be able to use anything else on top of that

romainmenke:
>  is there any caveat right now with mixins

No, I think it is fixed again.

------

mixins as currently implemented in preprocessors are semantically iffy.
It doesn't make much sense that an `at rule` is a reference to a collection of declarations or rules. There is nothing like them in standard CSS. This is also why they are a constant source of issues in nesting.

Maybe it is time to "invent" something new that provides the same feature set but is semantically correct and will work better with features like standard nesting.

Antonio-Laguna (approved):

PR #854, opened 2023-02-12
postcss-text-decoration-shorthand: set preserve default to true

romainmenke:
This plugin has cases where it doesn't produce an exact equivalent.
Making `preserve: true` the default is better in this case.

```css
.ab {
	text-decoration: wavy overline;
}
```

`preserve: true`

```css
.ab {
	text-decoration: overline;
	-webkit-text-decoration: wavy overline;
	        text-decoration: wavy overline;
}
```

`preserve: false`

_`wavy` is gone_

```css
.ab {
	-webkit-text-decoration: overline;
	text-decoration: overline;
}
```

That `wavy` is gone is not a bug, `text-decoration: overline;` is still a valid fallback.
It is simply not an equivalent.

Making `preserve: true` the default however will make `postcss-preset-env` work better without any config.


Antonio-Laguna (approved):
LGTM and makes sense!

I love one how `false` to `true` in two files cascade changes to 30 files 😂

PR #856, opened 2023-02-13
css-tokenizer : fix `isToken()`

romainmenke:
Before this change : 

```
// true
isToken([TokenType.Number, (3).toString(), -1, -1, { value: 3, type: NumberType.Number }]);
// false
isToken(['token-number', (3).toString(), -1, -1, { value: 3, type: NumberType.Number }]);
```

After both are `true`


Antonio-Laguna (approved):
LGTM!

PR #859, opened 2023-02-15
Adding support for max/min-inline/block-size on logical

### LINKED ISSUE #858: max- and min- logical sizes not being transformed
### Bug description

`(max|min)-(block|inline)-size` aren't being transformed like `(block|inline)-size`.

### Source CSS

```css
.foo {
  block-size: 1px;
  inline-size: 2px;
  min-block-size: 3px;
  max-block-size: 4px;
  min-inline-size: 5px;
  max-inline-size: 6px;
}
```

### Expected CSS

```css
.foo {
  height: 1px;
  width: 2px;
  min-height: 3px;
  max-height: 4px;
  min-width: 5px;
  max-width: 6px;
}
```

### Actual CSS

```css
.foo {
  height: 1px;
  width: 2px;
  min-block-size: 3px;
  max-block-size: 4px;
  min-inline-size: 5px;
  max-inline-size: 6px;
}
```

Yes

### Debug output

_No response_

### Extra config

_No response_

PostCSS Logical

### Plugin version

6.0.1

Linux, macOS

### Node Version

18.14.0

### Validations

- [ ] I'm willing to open a PR

Antonio-Laguna:
Fixes #858 

--- 

Browser compat only differs on Samsung Internet. See [min-block-size](https://developer.mozilla.org/en-US/docs/Web/CSS/min-block-size#browser_compatibility) vs [block-size](https://developer.mozilla.org/en-US/docs/Web/CSS/block-size#browser_compatibility). This is already tracked on CSSDB so no updates needed there. I just somehow completely missed this while scanning the spec!


romainmenke (approved):
LGTM!

PR #864, opened 2023-02-17
[postcss-global-data] Fix type for files option

gpoitch:
The type for the `files` option current only accepts an empty array or undefined. It should be an array of strings or undefined.

Screenshot 2023-02-17 at 2 32 13 PM

### DIFF
--- a/plugins/postcss-global-data/src/index.ts
+++ b/plugins/postcss-global-data/src/index.ts
@@ -4,7 +4,7 @@ import { parseImport } from './parse-import';
 /** postcss-global-data plugin options */
 export type pluginOptions = {
 	/** List of files to be used as context */
-	files?: string[],
+	files?: Array<string>,
 };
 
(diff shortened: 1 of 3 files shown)


plugins/postcss-global-data/src/index.ts
```diff
@@ -4,7 +4,7 @@ import { parseImport } from './parse-import';
 /** postcss-global-data plugin options */
 export type pluginOptions = {
 	/** List of files to be used as context */
-	files?: [],
+	files?: string[],
```
romainmenke, line 7:
```suggestion
	files?: Array<string>,
```

gpoitch:
`"@typescript-eslint/array-type": ["error", { default: "generic" }]` if you want to enforce that style

romainmenke (approved):
Thank you for your contribution @gpoitch 🎉 
We will try to release this shortly.

PR #866, opened 2023-02-18
css tokenizer : helper function

romainmenke:
We always read the entire source and then process it, we never take advantage of the "streaming" aspect.

So our code base is starting to contain a whole lot of this : 

```ts
	const t = tokenizer(input, options);

	const tokens: Array<CSSToken> = [];

	{
		while (!t.endOfFile()) {
			const token = t.nextToken();
			if (token) {
				tokens.push(token);
			}
		}

		const token = t.nextToken(); // EOF-token
		if (token) {
			tokens.push(token);
		}
	}

	return tokens;
```

This adds a helper function in the tokenizer :

```ts
const tokens = tokenize({
	css: '@charset "utf-8";',
})
```


Antonio-Laguna (approved):
This LGTM!

PR #861, opened 2023-02-15
postcss-nesting : better handling of nested selectors that begin with a letter

romainmenke:

PR #868, opened 2023-02-19
mention restriction about nested rules starting with an identifier

mayank99:
the latest css nesting spec disallows this for technical reasons:
```pcss
.foo {
  span {
    color: hotpink;
  }
}
```

when i first upgraded to postcss-nesting 11.0, i was under the impression that such syntax would be supported, because those technical limitations do not exist outside browsers. so i was very surprised when https://github.com/csstools/postcss-plugins/pull/861 was released in 11.2.1 and my styles broke.

this note will help avoid such unexpected surprises for users of `postcss-nesting`.

### DIFF
--- a/plugins/postcss-nesting/README.md
+++ b/plugins/postcss-nesting/README.md
@@ -81,7 +81,7 @@ Future versions of this plugin will first warn and then error if you use `@nest`
 
 We advice everyone to migrate their codebase **now** to nested CSS without `@nest`.
 
-## ⚠️ Nested selectors cannot start with a identifier (tag name)
+## ⚠️ Nested selectors cannot start with a letter
 
 The [CSS Nesting specification](https://www.w3.org/TR/css-nesting-1/#example-34e8e94f) disallows nested selectors to start with an identifier (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
 
--- a/plugins/postcss-nesting/README.md
+++ b/plugins/postcss-nesting/README.md
@@ -83,26 +83,7 @@ We advice everyone to migrate their codebase **now** to nested CSS without `@nes
 
 ## ⚠️ Nested selectors cannot start with a letter
 
-The [CSS Nesting specification](https://www.w3.org/TR/css-nesting-1/#example-34e8e94f) disallows nested selectors to start with an identifier (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
-
-```pcss
-.foo {
-	/* ❌ invalid */
-	span {
-		color: hotpink;
-	}
-
-	/* ✅ valid */
-	& span {
-		color: hotpink;
-	}
-
-	/* ✅ valid */
-	:is(span) {
-		color: hotpink;
-	}
-}	
-```
+The current version of the [CSS Nesting specification](https://www.w3.org/TR/css-nesting-1/#example-34e8e94f) disallows nested selectors to start with an identifier (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
 
 ## Options
 
--- a/plugins/postcss-nesting/README.md
+++ b/plugins/postcss-nesting/README.md
@@ -83,7 +83,26 @@ We advice everyone to migrate their codebase **now** to nested CSS without `@nes
 
 ## ⚠️ Nested selectors cannot start with a letter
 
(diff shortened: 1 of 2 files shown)


romainmenke:
Hi @mayank99, thank you for contributing!

------

> when i first upgraded to postcss-nesting 11.0, i was under the impression that such syntax would be supported, because those technical limitations do not exist outside browsers

The readme starts with a note that this plugin follows the specification.
Is there anything in the docs that gives the impression that this plugin aims to do more than what browsers can do?

It must have also given a warning to make it clear that this was not supported.
Was there no such warning?

------

I am a bit hesitant to add this to the readme for multiple reasons :

- we detect this pattern and emit an actionable warning.
- the specification is still in flux and the CSSWG really wants to get rid of this restriction.
- we already document that we follow the specification and we don't want to document the specification in detail.

We do sometimes have docs about such aspects but mostly to avoid confusion when the "transpiled css" is not what people might expect :

- https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-custom-media/README.md#true-and-false
- https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-cascade-layers#how-it-works

mayank99:
> Was there no such warning?

@romainmenke You're right, there is a warning. I think I missed that, because the build still passes.

> the specification is still in flux and the CSSWG really wants to get rid of this restriction.

hmm, based on recent discussions, i thought the nesting spec was mostly finalized. Maybe not.

> we already document that we follow the specification and we don't want to document the specification in detail.

i agree, it does not make sense to document it in detail, but this _does_ directly affect the use of the plugin. plus, this nuance is very easy to miss in the spec.

would you accept the PR if i trim it down to just a couple lines?

romainmenke:
> would you accept the PR if i trim it down to just a couple lines?

Hehe no, the issue is not that it is long or short :)

It is just that I am hesitant to take on the role of explaining and documenting the feature itself. Especially given the state of the feature/specification and the complexity of the feature.

I think it might make sense to add some extra docs up until there are good docs on MDN and enough blog posts explaining this.

romainmenke:
@mayank99 I've made a few tweaks and copied it over to the readme template.

- included the warning so that people can find this when they google the warning text
- mentions both types of selectors where this issue will surface

What do you think?

plugins/postcss-nesting/README.md
```diff
@@ -81,6 +81,36 @@ Future versions of this plugin will first warn and then error if you use `@nest`
 
 We advice everyone to migrate their codebase **now** to nested CSS without `@nest`.
 
+## ⚠️ Nested selectors must start with a symbol
+
+The [CSS Nesting specification](https://www.w3.org/TR/css-nesting-1/#example-34e8e94f) disallows nested selectors to start with a letter (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
```
mayank99, line 86:
```suggestion
The current version of the [CSS Nesting specification](https://www.w3.org/TR/2023/WD-css-nesting-1-20230214/#example-34e8e94f) disallows nested selectors to start with a letter (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
```

plugins/postcss-nesting/docs/README.md
```diff
@@ -42,6 +42,36 @@ Future versions of this plugin will first warn and then error if you use `@nest`
 
 We advice everyone to migrate their codebase **now** to nested CSS without `@nest`.
 
+## ⚠️ Nested selectors must start with a symbol
+
+The [CSS Nesting specification](https://www.w3.org/TR/css-nesting-1/#example-34e8e94f) disallows nested selectors to start with a letter (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
```
mayank99, line 47:
```suggestion
The current version of the [CSS Nesting specification](https://www.w3.org/TR/2023/WD-css-nesting-1-20230214/#example-34e8e94f) disallows nested selectors to start with a letter (i.e. a tag name or element selector). To write such selectors, they need to be prefixed with `& ` or wrapped with `:is()`.
```

mayank99 (commented):
> included the warning so that people can find this when they google the warning text

This makes perfect sense! Including the warning makes it even more relevant to this project (rather than simply explaining the spec), and users know what to expect before installing the plugin.

> I think it might make sense to add some extra docs up until there are good docs on MDN and enough blog posts explaining this.

Exactly. I think docs should actively try to educate on the concepts that lie at the boundaries of what the code/package is designed for, or at the very least provide a pathway for the user to learn. In the future when there are better resources on this topic, the note could be removed and a link might suffice.

---

Only suggestion I have is to pin to the current version of the spec, in case the link to "Example 4" changes.

romainmenke (approved):
Thank you for this @mayank99 🎉

Issue #870, opened 2023-02-20
[postcss-custom-properties] TypeError: Cannot read properties of undefined (reading 'nodes')


Hypnosphi:
### Reproduction link

https://github.com/Hypnosphi/postcss-custom-properties-repro

### Bug description

Postcss crashes with an exception

### Can you reproduce it with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

Yes

### `npx` Output

given the following css

```css
.grid {
    --column-width: auto;
    --first-column: [start] var(--column-width);

    grid-template-columns: var(--first-column) auto;
}
```

the output is
```
Cannot read properties of undefined (reading 'nodes')
```

Inlining any of the vars or moving `[start]` fixes the issue

### What plugin are you experiencing this issue on?

PostCSS Custom Properties

### Plugin version

13.1.3

### What OS are you experiencing this on?

macOS

### Node Version

18.13.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
Hi @Hypnosphi,

Thank you for reaching out and for the detailed description and repro.
This was extremely helpful 🙇 

We've identified the cause and will try to release a fix as soon as possible.

romainmenke:
@Hypnosphi We have just released an updated version that should resolve your issues.
Thank you again for reporting this!

PR #871, opened 2023-02-20
custom-properties : fix exception on chained declarations

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/870


plugins/postcss-custom-properties/src/transform-value-ast.ts
```diff
@@ -63,6 +63,7 @@ export default function transformValueAST(root: valueParser.ParsedValue, customP
 
 			if (nodes.length) {
 				parent.nodes.splice(index, 1, ...nodes);
+				parent.nodes.forEach((x) => ancestry.set(x, parent));
```
romainmenke, line 66: When inserting new nodes the ancestry also needs to be updated to contain the newly inserted nodes.

Antonio-Laguna (approved):
LGTM!

PR #872, opened 2023-02-21
css-calc : abstract some conditions

romainmenke:
much cleaner this way :)
thank you for the suggestion 🙇


Antonio-Laguna (approved):
❤️ love this! 

Indeed much more readable now!

Antonio-Laguna:
@romainmenke I think we're missing the labeler here. Want to add as part of this PR?

PR #874, opened 2023-02-21
postcss-trigonometric-functions : changelog

romainmenke:
_I missed this one in : https://github.com/csstools/postcss-plugins/pull/872, not urgent_


Antonio-Laguna (approved):
LGTM

PR #877, opened 2023-02-26
`css-color-parser` + `postcss-color-mix-function`

romainmenke:
- [x] `css-color-parser` package
- [x] `postcss-color-mix-function` plugin
- [x] refactored `postcss-hwb-function` plugin
- [x] refactored `postcss-lab-function` plugin
- [x] refactored `postcss-oklab-function` plugin
- [x] refactored `postcss-color-function` plugin
- [x] removed some unused functions from `color-helpers` package
- [x] better checks for lower case matches in some functions [^1] 
- [x] better fallbacks for in/out of gamut colors with smaller overal output as a result

[^1]: `'blacK'.toLowerCase() === 'black'` but `K` isn't actually an upper case `k`. It is a different unicode char that is equivelant to an upper case `k` in certain locales.

Things I skipped/didn't change :
- `postcss-color-functional-notation`, technically `css-color-parser` can help here, but this plugin is more a syntax desugaring and not really a color transform
- relative color syntax because it just doesn't seem ready yet for implementations

--------

I think this is ready to be shipped as a standalone plugin.

There are a few minor open issues I still want to wait on before adding to `postcss-preset-env`. It seems that browsers have incorrect implementations for a wide range of cases and I don't want to add to `postcss-preset-env` before confirmation that our implementation is correct.

-------

Untitled Diagram (1)


packages/color-helpers/types/global.d.ts
```diff
@@ -1,5 +0,0 @@
-export {};
-
-declare global {
-	type Color = [number, number, number];
-}
```
romainmenke, line 5: This worked, but it wasn't technically correct.
`Color` is not a globally available type in node or any other runtime.

Using `import type` all over the place is messier but more accurate.

plugin-packs/postcss-preset-env/test/basic.ff66.expect.css
```diff
@@ -181,7 +173,7 @@
 }
 
 .test-hwb-function {
-	background-color: rgba(0, 195, 255, .5);
+	background-color: rgba(0, 196, 255, 0.5);
```
romainmenke, line 176: Small differences like these are fine.
A single point increase/decrease should be fine in the context of a fallback.

Any color transform from a cylindrical to a rectangular color space will be sensitive to implementation details and we completely overhauled ours.

plugin-packs/postcss-preset-env/test/basic.ff66.expect.css
```diff
@@ -113,9 +107,7 @@
 
 .test-lab-function {
 	background-color: rgb(179, 35, 35);
-	background-color: color(display-p3 0.64331 0.19245 0.16771);
 	color: rgba(179, 34, 35, 0.5);
-	color: color(display-p3 0.64368 0.19188 0.16791 / 50%);
 	order: 17;
 }
```
romainmenke, line 112: These colors were in gamut for `srgb`.
The fallback to `display-p3` was never needed.

plugin-packs/postcss-preset-env/test/basic.ff66.expect.css
```diff
@@ -320,52 +312,35 @@
...
-	color-1: color(display-p3 0.28515 0.27983 0.27246);
-	color-2: rgba(121, 34, 67, 1);
-	color-2: color(display-p3 0.43853 0.16014 0.26019 / 1);
-	color-3: rgba(121, 34, 67, .5);
-	color-3: color(display-p3 0.43853 0.16014 0.26019 / .5);
-	color-4: rgba(121, 34, 67, 1);
-	color-4: color(display-p3 0.43853 0.16014 0.26019 / 100%);
+	color-2: rgb(121, 34, 67);
+	color-3: rgba(121, 34, 67, 0.5);
+	color-4: rgb(121, 34, 67);
```
romainmenke, line 326: The updated serializer inserts spaces at consistent positions and normalizes values. (e.g. `.5` -> `0.5`)

plugins/postcss-text-decoration-shorthand/src/index.ts
```diff
@@ -312,16 +312,17 @@ const thicknessKeywords = [
 ];
 
 const colorFunctions = [
-	'rgb',
-	'rgba',
+	'color',
+	'color-mix',
```
romainmenke, line 316: added `color-mix` and 🔤

plugins/postcss-progressive-custom-properties/test/basic.expect.css
```diff
@@ -225,3 +225,28 @@
...
+	--color-mix-1: color-mix(in lch, purple 50%, plum 50%);
+	--color-mix-2: color-mix(in lch longer, purple 50%, plum 50%);
+}
+}
+
+@supports (color: color(display-p3 0 0 0)) {
+:root {
+	--color-mix-2: color(display-p3 0.64331 0.19245 0.16771);
+}
+}
```
romainmenke, line 252: With `display-p3` fallbacks a single declaration can produce 2 fallback declarations.

Combined with `progressive-custom-properties` it will produce alternating `@supports` rules.

I played around with an updated algorithm but it altered the overal order of CSS.

This test ensures I don't try the same thing again later without thinking about potential issues.

plugins/postcss-oklab-function/test/variables.expect.css
```diff
@@ -1,18 +1,18 @@
 :root {
-	--firebrick: rgb(90, 66, 47);
-	--firebrick-a50: rgba(83, 67, 71, 0.5);
+	--firebrick: rgb(141, 0, 58);
+	--firebrick-a50: rgba(139, 0, 68, 0.5);
```
romainmenke, line 3: With the optimized output these tests now require out of gamut colors to fully test the generated CSS.

packages/color-helpers/types/global.d.ts
Antonio-Laguna: I thought this was meant for these cases but good to know it isn't!

Antonio-Laguna (approved):
This looks **great**!

I had to review this on the editor since this is a huge change and makes the browser suffer a lot.

This is so great to have it finally there! I'm really looking forward to this getting merged and out in the wild.

Thanks for all the work and the changes around the tooling and the other color functions!

romainmenke:
Thank you for working through this massive review 🙇 
Yeah will be awesome to get this out there!

PR #881, opened 2023-02-27
postcss-tape : convert to workspace package

romainmenke:
This is a refactor we planned to do anyway and a few minor tweaks to make the package suitable for 3rd parties.

It can be released so that others can also make use of it.

----

_apparently the labeler also has a hard limit 😂_


Antonio-Laguna (approved):
LGTM! Nice we got this out too before waiting for Node 20 😂

PR #882, opened 2023-03-01
postcss-preset-env : improve debug output

romainmenke:
see : https://github.com/csstools/postcss-plugins/discussions/879#discussioncomment-5158831


Antonio-Laguna (approved):
LGTM. Been following that thread already!

PR #890, opened 2023-03-11
update dependencies

romainmenke:


.github/bin/license/allowed.mjs
```diff
@@ -152,6 +152,11 @@ const licenseExceptions = [
 		file: 'node_modules/caniuse-lite/package.json',
 		license: 'CC-BY-4.0',
 	},
+	{
+		// see : https://github.com/isaacs/path-scurry/issues/7
+		file: 'node_modules/path-scurry/package.json',
+		license: 'BlueOak-1.0.0',
+	},
```
romainmenke, line 159: There is no way to avoid this package

PR #894, opened 2023-03-19
selector-specificity : view-transition pseudos

romainmenke:
https://drafts.csswg.org/css-view-transitions-1/#view-transition-pseudos


Antonio-Laguna (approved):
LGTM!

PR #900, opened 2023-03-22
postcss-logical : `transition`

romainmenke:
We overlooked `transition` when refactoring `postcss-logical`.
I only noticed it now when revisiting all the downsides/benefits of the old plugin.

As far as I can tell we can only safely transform `transition`.
Changing `transition-property` is not possible because other `transition-*` longhands need to have matching indices.

```css
.before {
  transition-property: margin-inline;
  transition-duration: 1s;
}

.after {
  transition-property: margin-left, margin-right;
  transition-duration: 1s; /* missing value for index "1" */
}
```

---

I am now using the same transform functions for real declarations and for `transition`.

For `transition` I am constructing a new `Declaration` with the property taken from `transition` so that it can be passed to the same functions.


Antonio-Laguna (approved):
LGTM!

PR #901, opened 2023-03-24
`postcss-gradients-interpolation-method` : leverage `@csstools/css-color-parser`

romainmenke:
https://github.com/csstools/postcss-plugins/blob/270ddd3f7826ba981272f3f91f47fb1e8dd27981/plugins/postcss-gradients-interpolation-method/README.md

I've made this a breaking change because it is a complete rewrite.
I wouldn't yet add it to `postcss-preset-env`, but I think it is good to have some extra coverage for `color-mix` at this time.

The parser algorithm is more complex/messy than I would like which is why I don't feel fully confident in this yet. But should be ready for release so that people can play with it.


plugins/postcss-color-function/src/index.ts
```diff
@@ -19,16 +19,16 @@ const basePlugin: PluginCreator<basePluginOptions> = (opts?: basePluginOptions)
...
 			}
 
-			if (hasSupportsAtRuleAncestor(decl)) {
+			if (hasFallback(decl)) {
 				return;
 			}
 
-			const originalValue = decl.value;
-			if (!(colorFunctionRegex.test(originalValue.toLowerCase()))) {
+			if (hasSupportsAtRuleAncestor(decl)) {
```
romainmenke, line 31: This order of checks should be faster.

PR #911, opened 2023-03-29
fix(README): link to css-has-pseudo plugin

dpschen:
I used a relative URL which might be the right thing to do for the rest of the README as well.


romainmenke (approved):
Thank you for this @dpschen 🙇

romainmenke:
I've changed it back to full urls because I realized that we need those in case someone else renders these files (npm and other services)

But we indeed removed all `README-BROWSER.md` files and replaced them with chapters in the general `README.md`. I've updated all other incorrect links.

PR #912, opened 2023-03-29
`postcss-media-minmax`

romainmenke:


plugins/postcss-media-minmax/test/browser/basic.html
```diff
@@ -0,0 +1,1216 @@
...
+						assert_equals(
+							window.getComputedStyle(target).color,
+							'rgb(0, 128, 0)',
+							testCase['title'] + ", target '" + target.classList[0] + "'"
+						);
+					}
+
+					styleElement.remove();
+				}
+			}
```
romainmenke, line 1209: Because we are behind browsers we can easily compare native and polyfill behavior. We might have false positives (test passes but there is also a bug) but if tests fail in the fallback and pass in native we are absolutely certain that there is a real bug.

Antonio-Laguna (approved):
This LGTM!

Thanks for doing this!

PR #915, opened 2023-03-30
add `@csstools/postcss-media-minmax` to `postcss-preset-env`

romainmenke:
Ideally we roll this out together with the latest cssdb release so that users have both the updated plugin and updated support stats.


plugin-packs/postcss-preset-env/test/basic.autoprefixer.false.expect.css
```diff
@@ -68,7 +68,7 @@
 	}
 }
 
-@media (min-width: 480px) and (max-width: 767px) {
+@media (min-width: 480px) and (max-width: 767.98px) {
```
romainmenke, line 71: see : https://github.com/postcss/postcss-media-minmax/issues/19

PR #918, opened 2023-04-03
postcss-gradients-interpolation-method : fix same color interpolation

romainmenke:


plugins/postcss-gradients-interpolation-method/src/color-stop-list.ts
```diff
@@ -27,8 +27,11 @@ export function interpolateColorsInColorStopsList(colorStops: Array<ColorStop>,
 		interpolatedColorStops.push(colorStop);
 
 		if (
-			serializeP3(colorStop.colorData, false).toString() !== serializeP3(nextColorStop.colorData, false).toString() &&
-			colorStop.position.toString() !== nextColorStop.position.toString()
+			hueInterpolationMethod ||
+			(
+				serializeP3(colorStop.colorData, false).toString() !== serializeP3(nextColorStop.colorData, false).toString() &&
+				colorStop.position.toString() !== nextColorStop.position.toString()
+			)
```
romainmenke, line 34: With the same color as start and end there isn't any effect when using `shorter hue`, or when interpolating between rectangular color spaces.
So I optimized this, but did so too aggressively.

For `longer hue` there is an actual gradient.
Rendering a color wheel is trivial with two color stops with the same color and using `longer hue`.

Antonio-Laguna (approved):
LGTM!

PR #919, opened 2023-04-03
improve detection of calc expressions

romainmenke:
Most of the plugins and packages only look for `calc`, but they should look for any of the math functions that are allows as top level values.


packages/media-query-list-parser/src/util/component-value-is.ts
```diff
@@ -1,17 +1,42 @@
...
+	'min',
+	'mod',
+	'pow',
+	'rem',
+	'round',
+	'sign',
+	'sin',
+	'sqrt',
+	'tan',
+]);
```
romainmenke, line 38: This package doesn't depend on `css-calc` and I don't want to add it just for this list.

plugins/postcss-media-minmax/test/calc.css
```diff
@@ -109,3 +109,41 @@
 @media (calc(10px / 3) > width >= calc(100px / 3)) {}
 
 @media (calc(10px / 3) >= width >= calc(100px / 3)) {}
+
+/* color */
+/* px */
+@media (color < sin(90deg)) {}
```
romainmenke, line 115: `sin(90deg)` == `1`

_In the results only some outputs have `sin(90deg)` and others have a numeric value. Only when there is a need to solve the math expression will it do so, if the value can be used as is it will just preserve it._

Antonio-Laguna (approved):
LGTM! Didn't know this was even possible!

PR #921, opened 2023-04-05
add detection for `env()` functions

romainmenke:
Stuff like this should be supported in browsers : `@media (height > env(safe-area-inset-top)) {}`

But I don't know how to test this in a browser to verify.
Anyone knows what the magic incantation is to get `env(safe-area-inset-top)` be not `0`?

https://w3c.github.io/csswg-drafts/css-env/#env-function:~:text=Because%20environment%20variables%20don%E2%80%99t%20depend%20on%20the%20value%20of%20anything%20drawn%20from%20a%20particular%20element%2C%20they%20can%20be%20used%20in%20places%20where%20there%20is%20no%20obvious%20element%20to%20draw%20from%2C%20such%20as%20in%20%40media%20rules%2C%20where%20the%20var()%20function%20would%20not%20be%20valid.


plugins/postcss-media-minmax/test/env.expect.css
```diff
@@ -0,0 +1,7 @@
+@media (min-height: calc((safe-area-inset-top) + 0.02px)) {}
```
Antonio-Laguna, line 1: Just curious here, why is it 0.02 and not 0.01 ? Is that the limit?

Antonio-Laguna (approved):
LGTM! Nice one!

plugins/postcss-media-minmax/test/env.expect.css
romainmenke: `0.02` is the smallest fraction that works well for some older Safari versions.
This is something that was tested by bootstrap and reported to `postcss-media-minmax`.

Anything smaller than `0.02` is rounded away.
So `900.01 === 900` but `900.02 !== 900`

romainmenke: https://github.com/postcss/postcss-media-minmax/pull/28#issue-800876301

Antonio-Laguna: Thanks a lot!!

Issue #922, opened 2023-04-06
[postcss-is-pseudo-class] Cannot expand simple statements with complex selectors


WRtux:
### Bug description

I already know that complex selectors can cause problems, but I think the statement below is quite straightforward and can be expanded safely (currently, the complex selector is preserved). Is it possible to implement this feature, or is there an option to force to expand complex selectors?

### Source CSS

```CSS
:is(.frame, .frame> .wrapper)> .target {
  background-color: #ffc;
}
```

### Expected CSS

```CSS
.frame:not(.does-not-exist)> .target, .frame> .wrapper> .target {
  background-color: #ffc;
}
```

### Actual CSS

```CSS
:is(.frame> .wrapper)> .target {
  background-color: #ffc;
}
.frame:not(.does-not-exist)> .target {
  background-color: #ffc;
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

N/A

### Debug output

```
postcss-is-pseudo-class: ***.css:1:1: Complex selectors in ':is(.frame, .frame> .wrapper)> .dialog' can not be transformed to an equivalent selector without ':is()'.
postcss-is-pseudo-class: ***.css:1:1: Complex selectors in ':is(.frame> .wrapper)> .dialog' can not be transformed to an equivalent selector without ':is()'.
```

### Extra config

None. Default config.

### What plugin are you experiencing this issue on?

PostCSS Is Pseudo Class

### Plugin version

2.0.7

### Node Version

16.17.1

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
Thank you for requesting this @WRtux 🎉 

A lot of patterns can be supported relatively easy, but we don't know which patterns are common, adding support for all of them proactively would be too much work :) So it's really helpful to get this feedback.

We've prepared a change that should resolve this : https://github.com/csstools/postcss-plugins/pull/923

romainmenke:
@WRtux We have released this update.
Please let us know if something isn't working as expected 🙇

PR #923, opened 2023-04-06
is-pseudo : handle more complex selector patterns

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/922


plugins/postcss-is-pseudo-class/src/split-selectors/complex/child-adjacent-child.ts
```diff
@@ -1,6 +1,13 @@
 // :-csstools-matches(.a > .b) + :-csstools-matches(.c > .d)
 // equivalent to
 // .a.c > .b + .d
+//
+// and
+//
+// :-csstools-matches(.a > .b) ~ :-csstools-matches(.c > .d)
+// equivalent to
+// .a.c ~ .b + .d
+//
```
romainmenke, line 10: These are actually the same, trivial to support both.

plugins/postcss-is-pseudo-class/src/split-selectors/compound-selector-order.ts
```diff
@@ -61,8 +61,27 @@ export function sortCompoundSelectorsInsideComplexSelector(node) {
...
+							sortedCompoundSelectors[i],
+						],
+					}),
+				],
+			});
+
+			node.prepend(wrapped);
+		} else {
+			node.prepend(sortedCompoundSelectors[i]);
+		}
```
romainmenke, line 84: Sorting will group tag selectors `a` and `b` into `ab` which serializes as a single tag `ab`.

This bit is a "post fix" to wrap it back with `:is()` into `a:is(b)`

plugins/postcss-is-pseudo-class/src/split-selectors/compound-selector-order.ts
```diff
@@ -1,7 +1,7 @@
 import parser from 'postcss-selector-parser';
 
 export function sortCompoundSelectorsInsideComplexSelector(node) {
-	if (!node || !node.nodes) {
+	if (!node || !node.nodes || node.nodes.length === 1) {
```
romainmenke, line 4: `:is(a)` cause issues and doesn't require sorting anyway.
Better to short circuit.

plugins/postcss-is-pseudo-class/test/compound-after-complex-is.css
```diff
@@ -0,0 +1,76 @@
...
+c:is(a b) d {
+	order: 5.1;
+}
+
+/* https: //github.com/csstools/postcss-plugins/issues/922 */
+:is(.frame, .frame > .wrapper) > .target {
+  background-color: #ffc;
+}
+
+:is(:-csstools-matches(.a, .b)) > .c {
```
romainmenke, line 74: I want to be sure that code looking for `:-csstools-matches` can also handle selector lists.

Internally we never produce `:-csstools-matches` with lists, but want to cover it anyway.

(We produce `:-csstools-matches` as an intermediate step with partially transpiled selectors to prevent infinite loops.)

Antonio-Laguna (approved):
LGTM! Thanks al lot for the comments on specific areas of interest! Love it

Issue #930, opened 2023-04-07
using color-mix causes "TypeError: t.replaceComponentValues is not a function"


fallenturtle:
I'm running postcss 8.4.21 through yarn v3 and using postcss-preset-env 8.3.0. When I try to use:
background-color: color-mix(in lch, purple 50%, plum 50%);

and run yarn i see this error in the output and the target file is not compiled:
TypeError: t.replaceComponentValues is not a function

romainmenke:
Can you try updating transient dependencies?
It seems like yarn isn't updating dependencies of dependencies correctly in your case.

romainmenke:
We will rollout a patch with updated dependency versions listed in package.json.
But in the meantime you can work around this by updating transitive dependencies manually.

romainmenke:
@fallenturtle This issue should be resolved now.
Can you try updating to the latest version of the postcss plugins you are using?

fallenturtle:
It's working now. Thanks!

romainmenke:
Thank you for reporting this!

PR #931, opened 2023-04-07
list latest version of all dependencies in all `package.json` files

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/930


Antonio-Laguna (approved):
Thanks for handling this. Should we add some way to the release plan to deal with this?

romainmenke:
https://github.com/csstools/postcss-plugins/blob/main/.github/bin/release-plan/release-plan.mjs#L114-L115

I think this is incorrect and causing the issues.
Currently we only increment the versions if the dependent also needs a release.
But it should always do that.

romainmenke:
Hehe, too many conflicts now 😄 

I am going to close this one and automate it after : https://github.com/csstools/postcss-plugins/pull/938

Issue #914, opened 2023-03-30
Plugin links to external polyfill packages


dpschen:
Kind of a follow up of https://github.com/csstools/postcss-plugins/issues/860

If I'm not mistaken I found out another issue that made it harder for me to use the `focus-within-pseudo-class` polyfill properly:

### Actual Description in READMEs

In the [postcss-preset-env packages README.md](https://github.com/csstools/postcss-plugins/blob/main/plugin-packs/postcss-preset-env/README.md) under

*Plugins list* > *Plugins that need client library* > `focus-within-pseudo-class`

- **Library** links to [`https://github.com/jsxtools/focus-within`](https://github.com/jsxtools/focus-within) and
- **Polyfill** links to [`https://github.com/jsxtools/focus-within/blob/master/README-BROWSER.md`](https://github.com/jsxtools/focus-within/blob/master/README-BROWSER.md). 

### Expected description in READMEs

**A)** I think those links should probably be replaced with a reference to
	
[`https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-focus-within#browser`](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-focus-within#browser)

Since [`postcss-focus-within/src/browser.js`](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-focus-within/src/browser.js) contains already a polyfill, correct?

**OR** *if I understood this wrong*

**B)** The links from **Library** and **Polyfill** should be referenced at the top of the [README of postcss-focus-within](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-focus-within/README.md) in a similar way that [the README of postcss-focus-visible](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-focus-visible/README.md) mentions its polyfill:

> PostCSS Focus Visible lets you use the :focus-visible pseudo-class in CSS, following the Selectors Level 4 specification.
> 
> **It is the companion to the [focus-visible polyfill](https://github.com/WICG/focus-visible).** Note that this plugin alone is not sufficient to polyfill for :focus-visible and that you need the browser's polyfill as well.

### Actual Description in READMEs

In the quote the last paragraph `browsers's polyfill` confused me as well. Does this refer to:

- some kind of browser part of the csstool/postcss-plugin that I always need in order to get the postcss plugin running

**OR**

[...]

PR #943, opened 2023-04-10
clarify browser polyfills

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/914

I am hoping that by listing and mentioning the two required steps right at the top, that it will be clearer for users.


Antonio-Laguna (approved):
LGTM

PR #947, opened 2023-04-13
logical properties : automatic test

romainmenke:
`@webref/css` now adds a marker to properties that are part of a "logical property group".
This makes it possible to generate a list of related properties of which some are logical properties. (`height` and `block-size` are part of the same group)

This gives us a good signal if new logical properties are added to CSS specs and gives us extra coverage for all logical related plugins.

### DIFF
--- a/plugin-packs/postcss-preset-env/package.json
+++ b/plugin-packs/postcss-preset-env/package.json
@@ -110,11 +110,12 @@
 	},
 	"devDependencies": {
 		"@csstools/postcss-tape": "*",
+		"@webref/css": "^6.5.1",
 		"postcss-simple-vars": "^7.0.1"
 	},
 	"scripts": {
 		"prebuild": "node ./scripts/generate-plugins-data.mjs && eslint --fix ./src/plugins/*.mjs",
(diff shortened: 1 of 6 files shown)


plugin-packs/postcss-preset-env/package.json
```diff
@@ -110,11 +110,12 @@
 	},
 	"devDependencies": {
 		"@csstools/postcss-tape": "*",
+		"@webref/css": "^6.5.1",
 		"postcss-simple-vars": "^7.0.1"
 	},
 	"scripts": {
 		"prebuild": "node ./scripts/generate-plugins-data.mjs && eslint --fix ./src/plugins/*.mjs",
-		"build": "rollup -c ../../rollup/default.mjs",
+		"build": "rollup -c ../../rollup/default.mjs && node ./scripts/update-logical-properties-test.mjs",
```
romainmenke, line 118: Technically related to tests but I don't like it that test tasks modify files. This complicated CI.

Antonio-Laguna (approved):
Super nice! LGTM!

PR #951, opened 2023-04-17
Update adjust-selector-specificity.ts

nmn:
Uses a smaller `:not` rule to bump specificty.

### DIFF
--- a/plugins/postcss-cascade-layers/src/adjust-selector-specificity.ts
+++ b/plugins/postcss-cascade-layers/src/adjust-selector-specificity.ts
@@ -27,8 +27,8 @@ function generateNot(specificity: number) {
 
 	let list = '';
 	for (let i = 0; i < specificity; i++) {
-		list += ':not(#\\#)'; // something short but still very uncommon
+		list += '#\\#'; // something short but still very uncommon
 	}
 
-	return list;
+	return `:not(${list})`;
 }


romainmenke:
Hi @nmn, thank you for your contribution.

This change would break the cascade layers fallback for older browsers.
`:not()` with complex selector support is a recent feature.

Was there some issue in particular you are trying solve?

plugins/postcss-cascade-layers/src/adjust-selector-specificity.ts
```diff
@@ -27,8 +27,8 @@ function generateNot(specificity: number) {
 
 	let list = '';
 	for (let i = 0; i < specificity; i++) {
-		list += ':not(#\\#)'; // something short but still very uncommon
+		list += '#\\#'; // something short but still very uncommon
 	}
 
-	return list;
+	return `:not(${list})`;
```
romainmenke, line 33: breaks support for older browser versions : https://developer.mozilla.org/en-US/docs/Web/CSS/:not#browser_compatibility

romainmenke (changes requested):

romainmenke:
@nmn Going to close this pull requests because this change would limit the usefulness of the plugin, but please open an issue if there are any problems you are trying to solve.

nmn:
Looking at the MDN page, only selector “lists” seem to be a new feature. A repeated ID selector should work even in older browsers?

I will reopen once I can test support. Thanks!

romainmenke:
A repeated ID selector doesn't produce the right specificity, so that would definitely not work :)

You can test this by running `npm run build` and then `npm run test:browser`.
That will verify the plugin against the same sort of tests as Chrome, Safari, Firefox use to verify their native feature.

These tests will fail if you change the plugin to produce `:not(#\#, #\#)`

nmn:
@romainmenke My fix would have generated `:not(#\##\#)` and not `:not(#\#, #\#)` which has the right specificity.

However, I tested in Chrome 49 (as that is the oldest browser I need to support at work) and as stated, the double ID selector failed to work.

This is surprising to me because the new part of the specification is “selector lists” (many selectors separated by a comma). However, testing beats looking at specifications.

I guess the polyfill can’t be made any smaller!

Thanks for taking the time!

romainmenke:
The old version of `:not()` only supported a single simple selector, so only `:not(.foo)`, not `:not(.foo.bar)`, `:not(.foo, .bar)`, `:not(.foo > .bar)`, ...

It is true that the MDN labeling is confusing :)

PR #966, opened 2023-05-12
Fix indentation

igorbabko:


romainmenke:
Hi @igorbabko 

Can you elaborate a bit on what you are trying to solve?

For context, this example is actual output from the plugin and the plugin doesn't alter indents. So the generated output might look unformatted but it is what users of this plugin should expect.

romainmenke:
Closing because this is an unwanted change.

Feel free open an issue to discuss any problems you might be trying to solve here :)

igorbabko:
Hi @romainmenke !

Got it, thanks for the explanation 👍

PR #967, opened 2023-05-13
Update README.md: consistent indentation

caldwell-63:
Makes indentation consistent across the entire file

### DIFF
--- a/plugins/postcss-nesting/README.md
+++ b/plugins/postcss-nesting/README.md
@@ -9,19 +9,19 @@ you might want to use [PostCSS Nested] instead.
 
 ```pcss
 .foo {
-	color: red;
+  color: red;
 
-	&:hover {
-		color: green;
-	}
+  &:hover {
+    color: green;
+  }
 
-	> .bar {
-		color: blue;
-	}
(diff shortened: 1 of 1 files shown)


romainmenke:
Hi @igorbabko

Can you elaborate a bit on what you are trying to solve?

For context, these examples are actual output from the plugin and the plugin doesn't alter indents.

Why did you also change indents to spaces?

romainmenke:
Closing because this is an unwanted change.

Feel free open an issue to discuss any problems you might be trying to solve here :)

caldwell-63:
8-width tabs in readme look absolutely horrendous, therefore this pull request

romainmenke:
Hi @caldwell-63 

We always tabs exactly because anyone can configure their preferred tab size.
If yours has a width of 8 then it is because you are using a default or have set it to have a width of 8.

For GitHub you can find more info here: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference#

We will never go back to spaces :)

Issue #969, opened 2023-05-18
`postcss-preset-env` throws TypeError starting with v8


ERuseckas:
### Bug description

v7.8.3 is working fine, but all the v8 versions are giving me the same error:

```
[!] (plugin postcss) TypeError: o is not a function
src/global.pcss
TypeError: o is not a function
    at /node_modules/postcss-preset-env/dist/index.cjs:1:17553
    at Array.forEach (<anonymous>)
    at pluginIdHelp (/node_modules/postcss-preset-env/dist/index.cjs:1:17429)
    at OnceExit (//node_modules/postcss-preset-env/dist/index.cjs:1:18705)
    at LazyResult.runAsync (/node_modules/postcss/lib/lazy-result.js:433:21)
```

This error is thrown even if my used `src/global.pcss` file is completely empty.

### Can you reproduce it with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

None

### Extra config

my used postcss config
```js
                require('postcss-discard-comments')(),
		require('stylelint')({
			files: 'src/**/*.pcss',
			cache: true,
			cacheLocation: path.resolve(__dirname, 'node_modules/.cache/.stylelintcache')
		}),
		require('postcss-mixins')({
			mixinsFiles: path.resolve(__dirname, 'src/common.pcss')
		}),
		require('postcss-preset-env')({
			stage: 2,
			preserve: false,
			features: {
				'custom-media-queries': true,
				'custom-properties': false,
				'custom-selectors': true,
				'environment-variables': true
			}
		}),
		require('postcss-nested')(),
		require('postcss-discard-empty')(),
		require('postcss-reporter')({ clearReportedMessages: true })
```

versions:
```json
                "postcss": "^8.4.23",
		"postcss-discard-comments": "^6.0.0",
		"postcss-discard-empty": "^6.0.0",
		"postcss-mixins": "^9.0.4",
		"postcss-nested": "^6.0.1",
		"postcss-preset-env": "^8.3.2",
		"postcss-reporter": "^7.0.5",
		"stylelint": "^15.3.0"
```

it's running under `rollup` v2 by using latest `rollup-plugin-postcss`:
```js
        postcss({
		extract: path.resolve('dist/main.min.css'),
		minimize: true
	})
```

### What plugin are you experiencing this issue on?

PostCSS Preset Env

### Plugin version

8.*.*

### What OS are you experiencing this on?

macOS

### Node Version

18.3.0

### Validations

[...]

PR #970, opened 2023-05-18
postcss-preset-env: fix exceptions on unknown features

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/969


Antonio-Laguna (approved):
LGTM!

PR #972, opened 2023-05-19
`css-tokenizer` : latest specification

romainmenke:
- `unicode-range-token` : https://github.com/w3c/csswg-drafts/issues/8835#issuecomment-1555145650
- sign character in parsed values for numerics
- switch to codepoints and update the allowed codepoints that start an ident
- tokenize `-0` as plain zero, not as negative zero
- serialize `-0` as `calc(-1 * 0<unit>)`


packages/css-calc/test/additional/mod-rem-infinity.mjs
```diff
@@ -27,7 +27,7 @@ assert.strictEqual(
 );
 
 assert.strictEqual(
-	calc('mod(-0, infinity)'),
+	calc('mod(calc(-1 * 0), infinity)'),
```
romainmenke, line 30: true negative zero can only be the outcome of a calculation, it can never be the result of parsing.

PR #973, opened 2023-05-20
add `state` to `walk` methods

romainmenke:
This adds a more convenient way to switch logic based on what the parent nodes are in an AST.

We already have a helper to construct the list of ancestors of an AST node, and that can be used to calculate the same things but that is a lot less efficient.

Before this change :
- construct the hierarchy of ancestors
- for each node
   - list all ancestors
   - iterate over ancestors and rebuild state
   - switch logic for current node based on ancestor state

After this change :
- pass initial state (maybe an empty object)
- for each node
   - switch logic for current node based on ancestor state
   - set state for descendants

The state objects do a shallow clone, not a deep clone


Antonio-Laguna (approved):
This is great! Nice way to be able to flag things in entries!

PR #974, opened 2023-05-21
relative-color-syntax: blog post + demo

romainmenke:


sites/postcss-preset-env/blog/postcss-relative-color-syntax.md
```diff
@@ -0,0 +1,389 @@
...
+
+
+```css
+.swatch {
+	/* NOTE :
+		postcss-preset-env and other plugins 
+		can not generate fallback values when `var()` is used.
+
+		But using custom properties in this example
+		makes it clear which values change and which are static.
```
Antonio-Laguna, line 122: Love these bits!

Antonio-Laguna (approved):

Antonio-Laguna:
@romainmenke this looks AWESOME! 🙇🏻

Thanks a lot for doing this and for the great demo there!

Issue #978, opened 2023-05-26
postcss-color-functional-notation unexpectedly not working


satouriko:
### Bug description

Hi there, I found that with the default `postcss-preset-env`, case
```css
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87 107 149 / var(--tw-text-opacity));
}
```

will be downgrade by `custom-properties` to
```css
.text-link {
  --tw-text-opacity: 1;
  color: #576b95;
  color: rgb(87 107 149 / var(--tw-text-opacity));
}
```

and will not further downgrade by `postcss-color-functional-notation` to
```css
.text-link {
  --tw-text-opacity: 1;
  color: #576b95;
  color: rgba(87,107,149,var(--tw-text-opacity));
}
```

after postcss-color-functional-notation@5.0.2 because of https://github.com/csstools/postcss-plugins/pull/835

This makes users between the gap (e.g Chorme 49~64) cannot display color correctly, and there seems no option to override

https://caniuse.com/css-variables
https://caniuse.com/?search=space-separated

### Source CSS

```css
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87,107,149,var(--tw-text-opacity));
}
```

### Expected CSS

```css
.text-link {
  --tw-text-opacity: 1;
  color: #576b95;
  color: rgba(87,107,149,var(--tw-text-opacity));
}
```

### Actual CSS

```css
.text-link {
  --tw-text-opacity: 1;
  color: #576b95;
  color: rgb(87 107 149/var(--tw-text-opacity));
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

N/A

### What plugin are you experiencing this issue on?

PostCSS Color Functional Notation

### Plugin version

5.0.2

### Node Version

v16.15.1

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [ ] I'm willing to open a PR

romainmenke:
Hi @satouriko,  
Thank you for reaching out about this!

Can you share your PostCSS config?

What I see in the [playground](https://preset-env.cssdb.org/playground/) :

```css
/* source */
.text-link { 
  --tw-text-opacity: 1; 
  color: rgba(87,107,149,var(--tw-text-opacity));
[...]

PR #979, opened 2023-05-26
extend `progressive-custom-properties` to the `var()` function

romainmenke:
see : https://github.com/csstools/postcss-plugins/issues/978

```css
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87 107 149 / var(--tw-text-opacity));
}
```

There are two possible ways to generate fallbacks for this case :

A : resolve `var()` first

```
/* source */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87 107 149 / var(--tw-text-opacity));
}

/* step 1 */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87 107 149 / 1);
  color: rgba(87 107 149 / var(--tw-text-opacity));
}

/* step 2 */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87, 107, 149);
  color: rgba(87 107 149 / 1);
  color: rgba(87 107 149 / var(--tw-text-opacity));
}
```

B : resolve `functional notation` first

```
/* source */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87 107 149 / var(--tw-text-opacity));
}

/* step 1 */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87, 107, 149, var(--tw-text-opacity));
  color: rgba(87 107 149 / var(--tw-text-opacity));
}

/* step 2 */
.text-link {
  --tw-text-opacity: 1;
  color: rgba(87, 107, 149, 1);
  color: rgba(87, 107, 149, var(--tw-text-opacity));
  color: rgba(87 107 149 / var(--tw-text-opacity));
}
```

Option B is what we want in this case because it happens to cover a more interesting range of browsers correctly.

But it also makes it harder to first resolve `var()` so that we can fallback features with a more complex syntax. (e.g. `color(var(--space) 0 0 0)`)

The correct solution to both issues is to extend the functionality of `progressive-custom-properties` so that any value that requires fallbacks **and** contains a `var()` function is placed in a `@supports` check for both that feature and `var()`.


plugin-packs/postcss-preset-env/.tape.mjs
```diff
@@ -18,7 +18,7 @@ orderDetectionPlugin.postcss = true
 
 postcssTape(plugin)({
 	'basic': {
-		message: 'supports basic usage'
+		message: 'supports basic usage',
```
romainmenke, line 21:
```suggestion
		message: 'supports basic usage'
```

romainmenke:
thank you for reviewing all this 🙇

PR #994, opened 2023-06-01
site : shareable playground

romainmenke:
This change makes it possible to share the playground state.
Which is handy when showcasing problems or fixes.

Screenshot 2023-06-01 at 17 54 06

### DIFF
--- a/sites/postcss-preset-env/src/static/js/playground.js
+++ b/sites/postcss-preset-env/src/static/js/playground.js
@@ -17,7 +17,7 @@ const stateAtLoad = (() => {
 			return {};
 		}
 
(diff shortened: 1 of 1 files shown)


ehoogeveen-medweb:
Might make sense to base64-encode the text in the url - that's what e.g. the [TypeScript](https://www.typescriptlang.org/play) and [babel](https://babeljs.io/repl) playgrounds do.

romainmenke:
Thanks @ehoogeveen-medweb 🙇 

I think I added base64 in there correctly, but it's been a while since I tried to store a bunch of random data in a url :D

Antonio-Laguna (approved):

Antonio-Laguna:
Great idea!

PR #995, opened 2023-06-04
Fix value calculation for `hue` components in relative color syntax.

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/8839#issuecomment-1573840429

We were already normalizing all percentages to numbers but I overlooked that degrees in `hue` components also should be resolved to numbers.


Antonio-Laguna:
LGTM! Interesting to know this!

PR #996, opened 2023-06-04
nesting : latest specification

romainmenke:
- Add warning when using the deprecated `@nest` rule
- Add support for nested selectors that that begin with a letter


plugin-packs/postcss-preset-env/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes to PostCSS Preset Env
 
+### Unreleased (minor)
```
romainmenke, line 3: Made this a minor so that people won't be surprised by the warning.

plugins/postcss-nesting/test/at-nest.css
romainmenke, line 1: One big massive file with all the tests that have `@nest`.

In the source test files I removed the `@nest` prefix.
This makes it possible to keep testing the old behavior while making it easier to cleanup in the future.

plugins/postcss-nesting/src/index.ts
```diff
@@ -6,13 +6,16 @@ import walk from './lib/walk.js';
 export type pluginOptions = {
 	/** Avoid the `:is()` pseudo class as much as possible. default: false */
 	noIsPseudoSelector?: boolean,
+	/** Silence the `@nest` warning. */
+	silenceAtNestWarning?: boolean,
```
Antonio-Laguna, line 10: Might also be worth to surface on the CHANGELOG? Users seem to look for these

Antonio-Laguna (approved):
This LGTM!

Thanks for making all of the changes and the amount of tests!!

plugins/postcss-nesting/src/index.ts
romainmenke: Good catch! Thank you 🙇 

I've added it in the CHANGELOG and general README.
I also linked to the Stylelint plugin.

PR #997, opened 2023-06-04
`plugins-stylelint/no-at-nest-rule`

romainmenke:
A Stylelint plugin to change nested css with `@nest` to css without.
Effectively migrating to the latest specification.

I am not going to bother supporting the non-standard nesting that Chrome and Safari shipped.
i.e. I am not going to wrap stuff with `:is()`.

The most important thing for us is that our users stop using `@nest`.

-----

@Antonio-Laguna We discussed at some point that we would also create a monorepo for the Stylelint plugins under `@csstools`. I think this is still worthwhile but separate from this change.

If any Stylelint plugins are created in this repo they should be tightly coupled to a PostCSS plugin.

Let me know if you disagree :)


.github/bin/license/allowed.mjs
```diff
@@ -147,6 +147,11 @@ export const osiApprovedWithCCO = [
 ];
 
 const licenseExceptions = [
+	{
+		// because ... irony? 🤷
```
Antonio-Laguna, line 151: 😂

Antonio-Laguna (approved):
@romainmenke I agree completely!

I like how nice the tests are. 
Does this need any special handling from the release script?

romainmenke:
> I like how nice the tests are.
> Does this need any special handling from the release script?

I ran the dry run thingy and that indicated that it would just work.
If I recall correctly I made the release script fairly agnostic of package contents. Because it are all workspaces under the same root it should just work 🤞

PR #1003, opened 2023-06-15
Fix weird indentation in postcss-nesting README

BD103:
This makes the CSS snippet in the nesting plugin's README correctly formatted and more readable.

Please note that this is my first contribution to this project. Please comment if you would like me to do anything else!

> _Side Note:_ The code oddly enough uses tabs instead of spaces. Would you like me to fix that too, or keep it how it is?

### DIFF
--- a/plugins/postcss-nesting/README.md
+++ b/plugins/postcss-nesting/README.md
@@ -30,16 +30,16 @@ you might want to use [PostCSS Nested] instead.
 	color: red;
 }
 .foo:hover {
-		color: green;
-	}
+	color: green;
+}
 .foo > .bar {
-		color: blue;
-	}
+	color: blue;
+}
 @media (prefers-color-scheme: dark) {
 	.foo {
 		color: cyan;
-}
 	}
+}
 ```
 
 ## Usage


romainmenke:
Hi @BD103,

A similar change was proposed in the past and we explained in detail why the current example is correct and will not be changed : 

https://github.com/csstools/postcss-plugins/pull/967

> these examples are actual output from the plugin and the plugin doesn't alter indents.
>

> We always use tabs exactly because anyone can configure their preferred tab size.
> If yours has a width of 8 then it is because you are using a default or have set it to have a width of 8.
> 
> For GitHub you can find more info here: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-personal-account-settings/managing-your-tab-size-rendering-preference#
> 
> We will never go back to spaces :)

BD103:
@romainmenke Could you go a bit further into why this is an unwanted change? The linked PR fails to fully explain your perspective on it, all I can see is miscommunication.

For instance, why do you prefer tabs over spaces? Should a note be added under the example explaining why the CSS is not formatted? Wouldn't it be clearer for the reader if the example was formatted but a note was added explaining that the actual output is not?

I respect that you prefer tabs over spaces, I just want to know why. Thank you.

romainmenke:
tabs vs. spaces is an accessibility issue.
There are many blog posts about it, this is one : https://adamtuttle.codes/blog/2021/tabs-vs-spaces-its-an-accessibility-issue/

Tabs are better for accessibility. So we always use tabs, never spaces.
We value being inclusive towards any and all developers :)

------

The formatting of the example is exactly the way it is actual output from the plugin.
You can see this live in the playground : https://preset-env.cssdb.org/playground/#eyJzb3VyY2UiOiIuZm9vIHtcblx0Y29sb3I6IHJlZDtcblxuXHQmOmhvdmVyIHtcblx0XHRjb2xvcjogZ3JlZW47XG5cdH1cblxuXHQ%2BIC5iYXIge1xuXHRcdGNvbG9yOiBibHVlO1xuXHR9XG5cblx0QG1lZGlhIChwcmVmZXJzLWNvbG9yLXNjaGVtZTogZGFyaykge1xuXHRcdGNvbG9yOiBjeWFuO1xuXHR9XG59IiwiY29uZmlnIjp7ImJyb3dzZXJzIjpbIj4gMC4yJSBhbmQgbm90IGRlYWQiXSwibWluaW11bVZlbmRvckltcGxlbWVudGF0aW9ucyI6MCwic3RhZ2UiOjIsImxvZ2ljYWwiOnsiaW5saW5lRGlyZWN0aW9uIjoibGVmdC10by1yaWdodCIsImJsb2NrRGlyZWN0aW9uIjoidG9wLXRvLWJvdHRvbSJ9fX0%3D

Most of the examples are generated so that we have a guarantee that they are correct and accurate.

If a plugin produces poorly formatted CSS then the example will have poorly formatted CSS.

The exact formatting is never important in this context, these plugins aren't formatters or linters, they are build tools. Just like you wouldn't look at the output from babel and give feedback on the indenting there :)

BD103:
Thank you for the lengthy explanation, it explains a lot more now. While I would recommend adding a note near the example, I'll leave that up to you.

Thank you and enjoy your weekend! :)

romainmenke:
And you enjoy yours 🏖️

PR #1004, opened 2023-06-16
Fix building postcss-preset-env on Windows

ehoogeveen-medweb:
Since paths use backslashes on Windows but import statements need forward slashes, using `path.join` generates an invalid `plugins-options.ts` file.

By using `path.posix.join` instead, it outputs forward slashes regardless of the host platform, the generated file matches the committed file and the build works as expected (with `Git Bash` or similar).


romainmenke (approved):
LGTM!

Thank you for this contribution @ehoogeveen-medweb

PR #1005, opened 2023-06-16
On Windows, spawn npm with shell: true

ehoogeveen-medweb:
I hit some more errors trying to run the release plan on Windows!

On Windows, `npm` goes through a few layers of indirection and the initial entry point is `npm.cmd` in the Node.js installation folder (e.g. `C:\Program Files\nodejs`).

So to spawn `npm` from Node.js on Windows, it needs a shell. But not on Linux (or any platform other than Windows, probably), so pass the platform check into `child_process.spawn` as a boolean :)

### DIFF
--- a/.github/bin/release-plan/commit.mjs
+++ b/.github/bin/release-plan/commit.mjs
@@ -1,4 +1,5 @@
 import { spawn } from 'child_process';
+import { platform } from 'process';
 
 export async function commitAfterPackageRelease(newVersion, packageDirectory, packageName) {
 	await new Promise((resolve, reject) => {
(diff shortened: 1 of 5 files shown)


romainmenke (approved):
Awesome!
I didn't know about this.

And thank you for testing all this on Windows.
We value that all these scripts work cross platform 🙇

ehoogeveen-medweb:
@romainmenke No problem! But I think the callsite you added is wrong (or at least unneeded): that callsite spawns `git` which (unlike `npm`) is just an executable (usually located at `C:\Program Files\git\cmd\git.exe`).

romainmenke:
That makes sense, thank you for calling that out.
I've reverted that change : https://github.com/csstools/postcss-plugins/commit/caafc13cc79c71a53693c0b1f7cde73fcdc75c2f

PR #1006, opened 2023-06-21
media queries : fix parsing of range queries with double equal signs

romainmenke:


packages/media-query-list-parser/src/nodes/media-feature-comparison.ts
```diff
@@ -30,8 +30,7 @@ export function matchesComparison(componentValues: Array<ComponentValue>): false
 						return [firstTokenIndex, i];
 					}
 
-					firstTokenIndex = i;
-					continue;
+					return [i, i];
```
romainmenke, line 33: For `==` it would continue here and then return on line `30` with `==` as the operator.

This however is invalid.
This function must always return when encountering `=`.

Either a lone `=` or `>=` | `<=`

Antonio-Laguna (approved):
LGTM!

Issue #1002, opened 2023-06-14
Link to changelog of dependencies from changelogs


ehoogeveen-medweb:
### What would you want to propose?

Dependency updates are already listed, with the version number and the severity of the change (not sure if that's the right word), but if you want to know what the changes are you have to look the dependencies up manually.

### Suggested solution

Turn each listed dependency into a link to that dependency, and the version number into a link to the changelog for that dependency.

For example, in the latest postcss-preset-env release:
* Updated [`@csstools/postcss-relative-color-syntax`](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-relative-color-syntax) to [`1.0.2`](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-relative-color-syntax/CHANGELOG.md) (patch)

### Additional context

Aside from just curiosity, I have a habit of including a brief summary of changes whenever I update a tool or library we use. This would save me a bunch of time, especially for larger releases :)

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
Thank you @ehoogeveen-medweb for opening an issue about this 🎉 

These entries are already largely generated, so should be straightforward to add :

https://github.com/csstools/postcss-plugins/blob/de99109c5a2881864892849e9d255a4f8ebfe7bf/.github/bin/release-plan/release-plan.mjs#L127-L147

Please feel free to open a pull request if you have time.

ehoogeveen-medweb:
@romainmenke How do you actually run it in a way that causes it to update the changelog?

I tried checking out https://github.com/csstools/postcss-plugins/commit/2009d84a1d0030374c392b81590a338792032a09 which comes right before https://github.com/csstools/postcss-plugins/commit/b8294e221db9690e41c541afc2a5305cd0b747f7 (which updates the `postcss-preset-env` changelog), then running `npm run release-plan`, but it just says `Nothing to release`. Is there something I need to do first?

romainmenke:
🤔 yes, we should document this better.

----

You can add this in any plugin Changelog:

```
### Unreleased (patch)

- dummy change
```

ehoogeveen-medweb:
[...]

PR #1008, opened 2023-06-21
Link to dependencies from changelogs

ehoogeveen-medweb:
Thanks for the assistance! This seems to do the trick.

I made the links relative with a leading slash: When rendering Markdown blobs, GitHub resolves those relative to the root of the repository and the branch being viewed, so the links work across forks and branches.

Here are some examples from a test branch on my fork:
* [css-calc changelog](https://github.com/ehoogeveen-medweb/postcss-plugins/blob/linkify-dependencies-test/packages/css-calc/CHANGELOG.md)
* [postcss-custom-properties changelog](https://github.com/ehoogeveen-medweb/postcss-plugins/blob/linkify-dependencies-test/plugins/postcss-custom-properties/CHANGELOG.md)

Fixes https://github.com/csstools/postcss-plugins/issues/1002.

### DIFF
--- a/.github/bin/release-plan/release-plan.mjs
+++ b/.github/bin/release-plan/release-plan.mjs
@@ -117,6 +117,10 @@ for (const workspace of notReleasableNow.values()) {
 		if (needsRelease.has(dependency)) {
 			const updated = needsRelease.get(dependency);
 
+			const dependencyLink = `/${updated.path.replaceAll('\\', '/')}`;
+			const nameAsLink = `[\`${updated.name}\`](${dependencyLink})`;
+			const versionAsLink = `[\`${updated.newVersion}\`](${dependencyLink}/CHANGELOG.md)`;
+
 			if (
 				packageInfo.dependencies &&
 				packageInfo.dependencies[updated.name] &&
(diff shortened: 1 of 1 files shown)


ehoogeveen-medweb:
Hmm, I guess it should also be possible to link to the section for the new version, but it might be a bit finicky. For example, the test change has heading ID [#212-june-21-2023](https://github.com/ehoogeveen-medweb/postcss-plugins/blob/linkify-dependencies-test/packages/css-tokenizer/CHANGELOG.md#212-june-21-2023).

Adding a [custom ID for each heading](https://www.markdownguide.org/extended-syntax/#heading-ids) might work.

romainmenke (approved):
This is awesome!
Very nice to see it in action with the examples you provided 🎉 

Thank you!

-----

Custom ID headings don't seem supported everywhere, and most importantly not on GitHub : https://github.github.com/gfm/#atx-headings

A technically simpler solution would be to change the change log format :

```md
### <semver version>

_<release date>_
```

```md
### 2.1.1

_June 1, 2023_
```

Then the heading id should be very predictable. `version.replaceAll('.', '-')`

I think we can do this in a follow up PR.
_I have time to update all the change log files in the next few days._

Thoughts?

ehoogeveen-medweb:
Yeah that sounds good, I don't think there should be any issue with headings that are simply version numbers.

PR #1021, opened 2023-06-27
postcss-src

romainmenke:
see : https://github.com/google/postcss-src/issues/7

Todo : 
- [ ] add to `cssdb`
- [ ] update tests after `cssdb` release
- [x] wait for `postcss-preset-env` `v9.0.0` to be released and to settle
- [ ] check that this doesn't break the site, it's a commonjs module.


romainmenke:
going to close this for now as we are slowly moving away from commonjs and adding a plugin that is only commonjs would become a hurdle later

PR #1024, opened 2023-06-29
document prefered plugin order

romainmenke:


plugin-packs/postcss-preset-env/README.md
```diff
@@ -159,6 +159,34 @@ postcss([
...
+				stage: 2,
+			},
+		],
+		// ...
+		// maybe a minifier?
+	],
+};
+```
+
+You can also use the [`insertBefore` / `insertAfter`](#insertbefore--insertafter) plugin options for more fine grained control.
```
yoriiis, line 188: What's the difference between declaring plugins directly in the `plugins` array (in the correct order) and doing it through these helpers? I feel like it's the same

plugin-packs/postcss-preset-env/README.md
```diff
@@ -159,6 +159,34 @@ postcss([
...
+[PostCSS Preset Env] is unable to handle non-standard features and syntactic sugar.
+
+If you also have PostCSS plugins for non-standard features and syntactic sugar,  
+you should place these first in the PostCSS config list.
+
+```js
+module.exports = {
+	plugins: [
+		"postcss-syntactic-sugar",
+		"postcss-non-standard",
```
yoriiis, line 174: Could we mention `postcss-nested` maybe ? Because it's clearly a non-standard syntax. Or another if you have an example

plugin-packs/postcss-preset-env/README.md
```diff
@@ -159,6 +159,34 @@ postcss([
 - [Gulp](INSTALL.md#gulp)
 - [Grunt](INSTALL.md#grunt)
 
+### Position in your PostCSS config list
```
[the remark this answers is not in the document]
yoriiis: I think we should mention the order of the plugins

```suggestion
### Order of plugins in PostCSS configuration
```

plugin-packs/postcss-preset-env/README.md
romainmenke: With these helpers you can "insert" a plugin that isn't part of `postcss-preset-env` in between plugins bundled within `postcss-preset-env`.

With the `plugins` array you can only insert it before or after all of `postcss-preset-env`

```
- your plugin
- all of postcss-preset-env
- your plugin
```

vs.

```
- postcss-preset-env
  - bundled plugin A
  - your plugin
  - bindled plugin B
```

yoriiis: Okay thanks for the clarification

plugin-packs/postcss-preset-env/README.md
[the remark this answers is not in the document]
romainmenke: A position in a list implies ordering :)

The subject of this documentation is `postcss-preset-env`.  
This section is for the position of `postcss-preset-env` in your PostCSS config.

Not a general guidance for the order of all plugins in your PostCSS config.

plugin-packs/postcss-preset-env/README.md
romainmenke: I didn't want to call out any plugin in particular because then it would seem as if only that specific plugin needs to be placed at a specific position.

plugin-packs/postcss-preset-env/README.md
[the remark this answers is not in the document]
yoriiis: Alright, so maybe this one fits better

```suggestion
### Position in your PostCSS plugins list
```

plugin-packs/postcss-preset-env/README.md
```diff
@@ -159,6 +159,34 @@ postcss([
 - [Gulp](INSTALL.md#gulp)
 - [Grunt](INSTALL.md#grunt)
 
+### Position in your PostCSS plugins list
+
+[PostCSS Preset Env] only contains polyfills and fallbacks for standard CSS features.  
+[PostCSS Preset Env] is unable to handle non-standard features and syntactic sugar.
```
romainmenke, line 165:
```suggestion
[PostCSS Preset Env] only contains polyfills and fallbacks for standard CSS features and is unable to handle non-standard features and syntactic sugar.
```

plugin-packs/postcss-preset-env/README.md
```diff
@@ -159,6 +159,34 @@ postcss([
...
+
+If you also have PostCSS plugins for non-standard features and syntactic sugar,  
+you should place these first in the PostCSS config list.
```
romainmenke, line 168:
```suggestion
you should place these first in the PostCSS plugins list.
```

romainmenke:
Thank you for the feedback here @yoriiis 🙇

PR #1025, opened 2023-06-29
allow number values in `hwb` and use number values for RCS in `hwb` and `hsl`

romainmenke:
this was an addition in [css-color-5](https://drafts.csswg.org/css-color-5/#relative-HWB)

---

I've created a base branch for both this change and the changes for powerless components.


packages/css-color-parser/test/wpt/color-invalid-relative-color.mjs
```diff
@@ -24,12 +24,6 @@ const tests = [
 
 	// hsl(from ...)
 
-	// Testing invalid permutation (types don't match).
-	['hsl(from rebeccapurple s h l)'],
-	['hsl(from rebeccapurple s s s / s)'],
-	['hsl(from rgb(10%, 20%, 30%, 40%) s h l)'],
-	['hsl(from rgb(10%, 20%, 30%, 40%) s s s / s)'],
-
 	// Testing invalid values.
```
romainmenke, line 27: These tests change for invalid, unparseable color notations.
With the specification change, the color is no longer invalid.

Antonio-Laguna (approved):

PR #1033, opened 2023-07-07
update powerless component handling to latest specification text

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/8609

---

I've created a base branch for both this change and the changes to `hwb`.


packages/css-color-parser/src/color-data.ts
```diff
@@ -254,7 +254,7 @@ export function colorDataTo(colorData: ColorData, toNotation: ColorNotation): Co
 						break;
 					case ColorNotation.Lab:
 					case ColorNotation.OKLab:
-						outputColorData.channels = carryForwardMissingComponents(colorData.channels, [0], outputColorData.channels, [0]);
+						outputColorData.channels = carryForwardMissingComponents(colorData.channels, [0, 1, 2], outputColorData.channels, [0, 1, 2]);
```
romainmenke, line 257: `a` and `b` components are analogous

packages/css-color-parser/src/color-data.ts
```diff
@@ -294,6 +294,44 @@ export function colorDataTo(colorData: ColorData, toNotation: ColorNotation): Co
...
+			if (reducePrecision(out[1], 4) <= 0) {
+				out[2] = NaN;
+			}
+
+			break;
+		case ColorNotation.OKLCH:
+			if (reducePrecision(out[1], 6) <= 0) {
+				out[2] = NaN;
+			}
+
```
romainmenke, line 325: Only `hue` is powerless and only when there is no meaningful chroma.

packages/css-color-parser/src/color-data.ts
```diff
@@ -490,7 +528,10 @@ function reducePrecision(x: number, precision = 7): number {
 }
 
 export function colorDataFitsRGB_Gamut(x: ColorData): boolean {
-	const srgb = colorDataTo(x, ColorNotation.RGB);
+	const copy = JSON.parse(JSON.stringify(x)) as ColorData;
```
romainmenke, line 531: We don't want to mutate `x`

packages/css-color-parser/src/functions/color-normalize-channel-values.ts
```diff
@@ -65,7 +65,7 @@ export function normalize_Color_ChannelValues(tokens: Array<CSSToken>, colorData
 				token[2],
 				token[3],
 				{
-					value: token[4].value,
+					value: value,
```
romainmenke, line 68: These were typo's

packages/css-color-parser/src/serialize/p3.ts
```diff
@@ -9,7 +9,7 @@ import { toPrecision } from './to-precision';
 import { XYZ_D50_to_P3_Gamut } from '../gamut-mapping/p3';
 
 export function serializeP3(color: ColorData, gamutMapping = true): FunctionNode {
-	color.channels = convertPowerlessComponentsToMissingComponents(color.channels, color.colorNotation);
+	color.channels = convertPowerlessComponentsToZeroValuesForDisplay(color.channels, color.colorNotation);
```
romainmenke, line 12: "for display" is taken from the specification.

In our context it means that this color will be serialized to a string, and not further processed.

This makes it possible to give true white in `rgb` for `lab(100 0 10)`.
Which would otherwise become something non-white due to gamut mapping.

packages/css-color-parser/test/wpt/color-computed-lab.mjs
```diff
@@ -79,7 +79,7 @@ const tests = [
 	['lch(10 20 740deg)', 'rgb(51, 15, 19)', 'color(display-p3 0.18359 0.06826 0.0765)'],
 	['lch(10 20 -700deg)', 'rgb(51, 15, 19)', 'color(display-p3 0.18359 0.06826 0.0765)'],
 	['lch(-40 0 0)', 'rgb(0, 0, 0)', 'color(display-p3 0 0 0)'],
-	['lch(20 -20 0)', 'rgb(0, 56, 48)', 'color(display-p3 0.07361 0.21544 0.18776)'],
+	['lch(20 -20 0)', 'rgb(48, 48, 48)', 'color(display-p3 0.18938 0.18938 0.18938)'],
```
romainmenke, line 82: This was incorrect and was a result of the typo's

plugins/postcss-color-mix-function/test/basic.expect.css
```diff
@@ -40,11 +40,11 @@
...
 
 .spec-example-6 {
-	color-1: rgb(166, 158, 255);
-	color-1: color(display-p3 0.64126 0.60728 1);
+	color-1: rgb(174, 138, 255);
+	color-1: color(display-p3 0.66273 0.5437 0.98509);
 	/* TODO : https://github.com/w3c/csswg-drafts/issues/8609 */
-	color-2: rgb(116, 163, 255);
-	color-2: color(display-p3 0.49247 0.63302 1);
+	color-2: rgb(119, 163, 255);
```
romainmenke, line 46: Slightly different but close enough to not be breaking.
These are green or red in Chrome/Safari/...

Ours is blue before and after, just a slightly different tint.

packages/css-color-parser/src/color-data.ts
Antonio-Laguna: Good catch!

packages/css-color-parser/src/functions/three-channel-space-separated.ts
```diff
@@ -134,11 +134,19 @@ export function threeChannelSpaceSeparated(
...
+				// Removing this gives unexpected results for us and we don't have a good solution for it yet.
+				// Maybe we are holding it wrong.
+				// The results we get for HSL/HWB are good at this time, so keeping the code as is until we receive a bug report about this.
 				relativeToColor.channels = conversions.sRGB_to_HSL(XYZ_D50_to_sRGB_Gamut(colorData_to_XYZ_D50(relativeToColor).channels));
 			} else if (
 				colorNotation === ColorNotation.HWB
-			) { // https://github.com/w3c/csswg-drafts/issues/8444
+			) {
+				// https://github.com/w3c/csswg-drafts/issues/8444
+				// Removing this gives unexpected resultsfor us and we don't have a good solution for it yet.
```
Antonio-Laguna, line 147:
```suggestion
				// Removing this gives unexpected results for us and we don't have a good solution for it yet.
```

Really minor typo

Antonio-Laguna (approved):
This looks great to me!

Issue #1041, opened 2023-07-11
postcss-preset-env behaves differently than in the "interactive playground"


diesieben07:
### Bug description

`postcss-preset-env` behaves different when I use it locally compared to the [interactive playground](https://preset-env.cssdb.org/playground/), specifically in terns of the `custom-properties` feature.
I want to explicitly support only browsers that already support this feature. Therefor I set my browserslist configuration as `defaults and supports css-variables`.
In the interactive playground this has the desired effect and custom properties are not transformed (see [here](https://preset-env.cssdb.org/playground/#eyJzb3VyY2UiOiJib2R5IHtcbiAgLS1jb2xvcjogcmVkO1xuICBjb2xvcjogdmFyKC0tY29sb3IpO1xufSIsImNvbmZpZyI6eyJicm93c2VycyI6WyJkZWZhdWx0cyBhbmQgc3VwcG9ydHMgY3NzLXZhcmlhYmxlcyJdLCJtaW5pbXVtVmVuZG9ySW1wbGVtZW50YXRpb25zIjowLCJzdGFnZSI6MiwibG9naWNhbCI6eyJpbmxpbmVEaXJlY3Rpb24iOiJsZWZ0LXRvLXJpZ2h0IiwiYmxvY2tEaXJlY3Rpb24iOiJ0b3AtdG8tYm90dG9tIn19fQ%3D%3D)).

Locally however, with the exact same postcss.config.mjs file, they are replaced, seen in [this repository](https://github.com/diesieben07/postcss-reproduction).

### Source CSS

```css
body {
    --color: red;
    color: var(--color);
}
```

### Expected CSS

```css
body {
    --color: red;
    color: var(--color);
}
```

### Actual CSS

```css
body {
    --color: red;
    color: red;
    color: var(--color);
}
```

### Debug output

```
⚠  Using features from Stage 2 (default)
  prefers-color-scheme-query has been disabled by "enableClientSidePolyfills: false".
  custom-selectors with stage 1 has been disabled
  focus-visible-pseudo-class has been disabled by "enableClientSidePolyfills: false".
  focus-within-pseudo-class has been disabled by "enableClientSidePolyfills: false".
  blank-pseudo-class has been disabled by "enableClientSidePolyfills: false".
  has-pseudo-class has been disabled by "enableClientSidePolyfills: false".
Disabling :is on "nesting-rules" due to lack of browser support.
Adding area[href] fallbacks for ":any-link" support in Edge and IE.
[...]

PR #1042, opened 2023-07-11
fix browserslist support queries

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1041

On the site :

Screenshot 2023-07-11 at 19 00 25

I didn't really want to provide UI to communicate error messages.
`alert()` should be fine for now 😄


plugin-packs/postcss-preset-env/src/lib/browsers-with-supports-stats.mjs
```diff
@@ -0,0 +1,16 @@
...
+	'edge',
+	'firefox',
+	'ie',
+	'ios_saf',
+	'op_mini',
+	'op_mob',
+	'opera',
+	'safari',
+	'samsung',
+];
```
romainmenke, line 16: If this includes something that we don't have data for, like `baidu` and the browserslist query of the user also includes one of these then the internal logic in `postcss-preset-env` will always transpile everything.

I've moved this to a separate file and I am not using this a filter consistently throughout the codebase.

diesieben07:
Wow, thank you so much for the blazing-fast turn around! Good work! :rocket:

romainmenke:
We still have to release this, but we will do so as soon as possible 🎉

PR #1048, opened 2023-07-17
add `postcss-exponential-functions` to `postcss-preset-env`

romainmenke:


romainmenke:
@Antonio-Laguna I think it would also be good to have a cssdb release before this one.
That would be a nice addition to `9.1.0`, then we have a nice mix of updated data, new options, bug fixes and a new plugin :)

PR #1050, opened 2023-07-21
improve support for open-props

romainmenke:
[`open-props`](https://open-props.style) is something that users will want to combine with `postcss-global-data`.

However I encountered two issues and then discovered 1 more :
- referencing npm packages didn't work in `postcss-global-data`
- `open-props` uses the zero specificity `:where(html)`
- cascade layers support for `postcss-custom-properties` had a typo and was broken


plugins/postcss-custom-media/src/cascade-layers.ts
```diff
@@ -103,19 +103,19 @@ export function collectCascadeLayerOrder(root: Root) {
...
-			return -1;
+			return 0;
 		}
 
-		return layers.get(node.parent);
+		return layers.get(node.parent)! + 1;
 	}
 
-	return Infinity;
+	return 10_000_000;
```
romainmenke, line 118: `Infinity` worked, but limited what was possible.
You can't do `Infinity - 100` or `Infinity * 10`.

plugins/postcss-custom-properties/src/get-custom-properties-from-root.ts
```diff
@@ -28,29 +27,21 @@ export default function getCustomPropertiesFromRoot(root: Root): Map<string, val
...
+				specificity = 0;
+			} else if (HTML_SELECTOR_REGEXP.test(selector)) {
+				specificity = 1;
+			} else if (ROOT_SELECTOR_REGEXP.test(selector)) {
+				specificity = 2;
+			} else {
+				return;
+			}
 
-				const thisCascadeLayer = cascadeLayerNumberForNode(decl, cascadeLayersOrder);
```
romainmenke, line 45: `cascadeLayerNumberForNode(decl, cascadeLayersOrder);`

This looks at the immediate parent, in this case a rule with a selector is the parent of the `decl`, it will never find a layer rule.

This typo was introduced when updating all three `postcss-custom-*` plugins. The other two are for at-rules where the immediate parent can be a layer rule.

Changed to : ``cascadeLayerNumberForNode(rule, cascadeLayersOrder);``

PR #1059, opened 2023-07-25
nesting: `@starting-style`

romainmenke:


plugins/postcss-nesting/src/lib/valid-atrules.ts
```diff
@@ -1 +1 @@
-export default ['container', 'document', 'media', 'supports', 'layer'];
+export default ['container', 'document', 'media', 'supports', 'layer', 'starting-style'];
```
Antonio-Laguna, line 1: Such an easy change!

Antonio-Laguna (approved):
LGTM!

PR #1060, opened 2023-07-25
`postcss-bundler`

romainmenke:
A plugin pack for bundling CSS


romainmenke:
@Antonio-Laguna I've split some parts of this out into separate pull requests.

±20k lines and 600 files is a bit much.
The actual code is a tiny fraction of that.

Part of this depends on those other plugins, so I can't split it fully, but I will make more chunks after the other pull requests have been merged.

Issue #1061, opened 2023-07-28
Unexpected `Unknown feature` warning


equinusocio:
Hello, I'm **disabling** the [`oklab-function`](https://preset-env.cssdb.org/features/#oklab-function) in my config, and recently I'm getting a warning that the feature doesn't exist. It's suggesting to use [`lab-function`](https://preset-env.cssdb.org/features/#lab-function) which is a different one.

```
[next:dev] (1:1) postcss-preset-env: Unknown feature: "postcss-oklab-function", did you mean: "lab-function"
```

My postcss config:

```js
import postcssMixins from 'postcss-mixins';

{
  plugins: {
    'postcss-import': {},
    'postcss-replace': {
      pattern: /token\(.*?--([^\s]+?)\)/gi,
      data: tokens,
    },
    'postcss-preset-env': {
      stage: 0,
      features: {
        'logical-properties-and-values': false,
        'prefers-color-scheme-query': false,
        'postcss-oklab-function': false,
        'gap-properties': false,
        'cascade-layers': false,
      },
      insertAfter: {
        'custom-media-queries': postcssMixins,
      },
    },
    'postcss-mixins': {},
    cssnano: {
      preset: [
        'default',
        {
          discardComments: {
            removeAll: true,
          },
        },
      ],
    },
  }
}
```
I'm getting this warning in every file where I use `oklch` colors

![CleanShot 2023-07-28 at 15 47 36](https://github.com/csstools/postcss-plugins/assets/10454741/982f65ee-dff5-4fee-aeaa-3c3a7fb3e5c6)

romainmenke:
Hi @equinusocio,

You can find the feature name here : https://github.com/csstools/postcss-plugins/blob/main/plugin-packs/postcss-preset-env/FEATURES.md

The feature name is : `oklab-function`

_I will need to check if we can make it suggest that correctly._

-----


You should move `'postcss-mixins': {},` up in your config.

It's better to have non-standard features first, even with PostCSS 8 and later.

-----

`prefers-color-scheme-query` This won't be enabled by default in recent versions of `postcss-preset-env`, so you can remove this override.

romainmenke:
@equinusocio,

I am also a bit curious why you are disabling the `oklch` plugin.

Is there a specific issue you encountered?

equinusocio:
Because we don't want the following transformation:

```css
.test-oklab {
	color: oklab(40% 0.001236 0.0039);
}

.test-oklch {
[...]

PR #1062, opened 2023-07-28
display multiple suggestions when a config has an unknown feature

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1061

We use a Levenshtein distance to find the most similar real feature but we also use the npm package names to find entries. This makes it possible to encounter this edge case :

- user enters `postcss-oklab-function`
- the package name is `@csstools/postcss-oklab-function`
- the feature name is `oklab-function`
- most similar result is `postcss-lab-function`

This is the most similar result because only two characters need to be altered.
Whereas the right result requires at least 8 changes.

```
user entry:
          postcss-oklab-function

candidates:
            postcss-lab-function
@csstools/postcss-oklab-function
                  oklab-function
```

This can not be fixed in a general, yet simple way.

----------

Instead of showing only the top hit we will now show the 3 most likely features.

This makes it possible for users to pick the correct option even when a Levenshtein distance gives an unexpected result.

```
[postcss-preset-env]: Unknown feature: "postcss-oklab-function", did you mean one of: "lab-function", "oklab-function", "image-set-function"
```


plugin-packs/postcss-preset-env/.tape.mjs
```diff
@@ -18,7 +18,7 @@ orderDetectionPlugin.postcss = true
 
 postcssTape(plugin)({
 	'basic': {
-		message: 'supports basic usage'
+		message: 'supports basic usage',
```
romainmenke, line 21:
```suggestion
		message: 'supports basic usage'
```

PR #1066, opened 2023-08-01
Ensure `env` parameter is checked when building default options

BPScott:
Fix a bug in the `env` option that was introduced to preset-env in #1012.

It did not take the `env` option into account when building the options that are passed to individual plugins. This resulted in generating plugin options based upon the default browserslist environment ('production') instead of the one specified in preset-env's `env` option.

This PR ensures that the env is read when determining plugin options, and updates the e2e test to demonstrate that. 

---

In the updated e2e test we test this using [`any-link-pseudo-class`'s `areaHrefNeedsFixing` option being set when edge or ie are in the targeted browsers](https://github.com/csstools/postcss-plugins/blob/cd3f9fcf134a485c8f0240ccef9f7974c21d5e7b/plugin-packs/postcss-preset-env/src/lib/get-options-for-browsers-by-feature.mjs#L34-L48), which results in an `area[href]` selector being added to the output.

The production environment does not target ie/edge, and thus the `area[href]` selector is absent.
The legacy-edge environment does target ie/edge, and thus the  `area[href]` selector is present.
Without this fix, the selector is erroneously absent from the legacy-edge environment.


plugin-packs/postcss-preset-env/src/lib/get-options-for-browsers-by-feature.mjs
```diff
@@ -3,7 +3,8 @@ import { getUnsupportedBrowsersByFeature } from './get-unsupported-browsers-by-f
 
 // add extra options for certain browsers by feature
 export default function getOptionsForBrowsersByFeature(browsers, feature, cssdbList, options, logger) {
-	const supportedBrowsers = browserslist(browsers, { ignoreUnknownVersions: true });
+	const env = browsers ? undefined : options.env; // When `browsers` is set, ignore `env`
```
BPScott, line 6: Instead of this, I could have added an `env` parameter to `getOptionsForBrowsersByFeature` and it's calling function `formatStagedFeature`.

That would match how we pass browsers through the functions but felt slightly clunky - these functions have enough parameters already so I didn't feel great about adding another one given that the env is already present in the options.

This logic matches what is in https://github.com/csstools/postcss-plugins/blob/cd3f9fcf134a485c8f0240ccef9f7974c21d5e7b/plugin-packs/postcss-preset-env/src/lib/list-features.mjs#L17

romainmenke (commented):
@BPScott Thank you for catching this and for suggesting a fix 🎉 

I think the main issue was that we were doing browserslist queries for the same query in two places. This will always be error prone.

I think it will be easier to maintain if we remove one and communicate the resulting list of browsers.

Reducing the number of calls to `browserslist` should also make everything a little bit faster.

Thoughts?

-------

I've also added a changelog entry and mentioned you there.

BPScott:
> I think it will be easier to maintain if we remove one and communicate the resulting list of browsers.

I like that! I didn't want to do much refactoring as I'm not familiar with the codebase but making that looks like it was very straight forward. Thank you for adding it for me.

"don't call browserslist in a loop" refactorings along the way are also super nice.

---

Thank you for the review and improvements. No outstanding thoughts for me!

Antonio-Laguna (approved):
This LGTM!

Thanks @BPScott !

romainmenke:
This has been released, thank you again for your contribution 🙇

PR #1067, opened 2023-08-01
add access checking to release planner

romainmenke:
There are still a few plugins I don't have access to.
This makes it harder for me to release new versions if one of the pending releases if for something I can't release.

This change adds access checks so that a release plan can be created that will complete without error.

Without this change a plan would be started and would error half way through, leaving the repo in a weird state.

### DIFF
--- a/.github/bin/release-plan/release-plan.mjs
+++ b/.github/bin/release-plan/release-plan.mjs
@@ -10,6 +10,7 @@ import { npmPublish } from './npm-publish.mjs';
 import { npmVersion } from './npm-version.mjs';
 import { updateDocumentation } from './docs.mjs';
 import { currentVersion } from './current-version.mjs';
+import { canPublish } from './npm-can-publish.mjs';
 
 const workspaces = await listWorkspaces();
 // Things to release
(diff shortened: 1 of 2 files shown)


.github/bin/release-plan/release-plan.mjs
```diff
@@ -44,6 +45,13 @@ for (const workspace of workspaces) {
 
 	let changelog = (await fs.readFile(path.join(workspace.path, 'CHANGELOG.md'))).toString();
 	if (changelog.includes('Unreleased')) {
+		const canPublishPackage = await canPublish(workspace.name);
```
Antonio-Laguna, line 48: Given you're rejecting errors this will halt the execution if one of them happens? You might want to `catch` these or wrap them with a try / catch

romainmenke: I think we want to halt execution in case of an error.  
This would occur very early, when building the list of packages to be released.  
So there isn't any chance of stopping halfway through the release process.

If we did add a try / catch it would also do a print of the error message and `process.exit(1)` :)

I did see that the error messages were incorrect and I've fixed those.

romainmenke: I do think there is some opportunity for better code here : https://github.com/csstools/postcss-plugins/issues/1068

Antonio-Laguna: Ahh I see...

Just realised that the errors were showstoppers

Antonio-Laguna (approved):

.github/bin/release-plan/release-plan.mjs
romainmenke: Indeed showstoppers, but it's a slippery slope :D
By not catching anything it quickly becomes a mess.
So thank you for pointing this out 🙇

PR #1082, opened 2023-08-15
`postcss-rebase-url`

romainmenke:
_part of https://github.com/csstools/postcss-plugins/pull/1060_

----

```pcss
/* when used with a bundler like `postcss-import` */

/* test/examples/example.css */
@import url("imports/basic.css");

/* test/examples/imports/basic.css */
.foo {
	background: url('../../images/green.png');
}

/* becomes */

/* test/examples/example.expect.css */
.foo {
	background: url("../images/green.png");
}
```

The benefit of a url rebaser is that you can use working file paths in your CSS.

In this example you can click through on the path and your code editor will take you to that file. This only works when you use a url rebaser.
Screenshot 2023-08-15 at 11 42 49


plugins/postcss-rebase-url/README.md
```diff
@@ -0,0 +1,66 @@
+# PostCSS Rebase URL [<img src="https://postcss.github.io/postcss/logo.svg" alt="PostCSS Logo" width="90" height="90" align="right">][PostCSS]
+
+[<img alt="npm version" src="https://img.shields.io/npm/v/@csstools/postcss-rebase-url.svg" height="20">][npm-url] [<img alt="Build Status" src="https://github.com/csstools/postcss-plugins/workflows/test/badge.svg" height="20">][cli-url] [<img alt="Discord" src="https://shields.io/badge/Discord-5865F2?logo=discord&logoColor=white">][discord]
+
+[PostCSS Rebase URL] rebases `url()` functions when transforming CSS.
+
+This plugin is only intended to help with bundling CSS and only in a way that you author CSS as if there was no bundling or url rebasing.
+
+If you need something with more knobs and dials, please checkout [`postcss-url`](https://www.npmjs.com/package/postcss-url)
+
```
Antonio-Laguna, line 10: Checking this and the example below it's not immediately clear what is going on. I wonder if it should get a better explanation of what's the problem and how is it being solved. I think the comments themselves can be confusing.

Something in these lines:

> It is normal to separate CSS files across several ones to ease maintenance and/or to create more modular approaches. However, when it comes to importing relative files, code editors can't really work out routes resources in `url` (such as images) because the path you need to use is not exactly relative to the file but rather to where the built file will be. 
> This plugin offers a simplistic/naive approach to solving that.

Thoughts?

plugins/postcss-rebase-url/package.json
```diff
@@ -0,0 +1,85 @@
+{
+	"name": "@csstools/postcss-rebase-url",
+	"description": "Rebase url() functions when transforming CSS",
+	"version": "0.1.0-alpha.0",
```
Antonio-Laguna, line 4: Is this intended? Not really against it but just double checking since we don't usually go this path.

plugins/postcss-rebase-url/src/rebase.mjs
```diff
@@ -0,0 +1,42 @@
+import path from 'path';
```
Antonio-Laguna, line 1: Is there any reason for this file not to be `.ts`?

plugins/postcss-rebase-url/src/normalized-dir.ts
```diff
@@ -0,0 +1,5 @@
+import path from 'path';
+
+export function normalizedDir(x: string): string {
```
Antonio-Laguna, line 3: Should we add some basic docs here (and serialize-string) as with `rebase`?

Antonio-Laguna (approved):
@romainmenke thanks for this!

This looks really good!

I left some questions and things that could improve but I don't see anything worth stopping from merging this!

Thanks!

plugins/postcss-rebase-url/package.json
romainmenke: I published an alpha version so that I could verify everything in : https://github.com/romainmenke/css-import-tests

plugins/postcss-rebase-url/README.md
romainmenke: Yeah, it's confusing.

I wrote that from my perspective when creating this.
Should be written from the perspective of a CSS authors who is using it.

> When bundling CSS the location of the final stylesheet file will be different than the individual source files.
> This plugin rewrites the contents of `url()` functions so that relative paths continue to work.
>
> Instead of mentally mapping where the files will all be in the final output you can use this plugin and simply use the relative paths of the source files.

plugins/postcss-rebase-url/src/normalized-dir.ts
romainmenke: done

plugins/postcss-rebase-url/README.md
romainmenke: I've made a few tweaks : https://github.com/csstools/postcss-plugins/blob/postcss-rebase-url--ambitious-whippet-ee3940f7aa/plugins/postcss-rebase-url/README.md

PR #1083, opened 2023-08-15
`@import` linter for bundlers

romainmenke:
This covers the main issues authors can encounter when bundling CSS :
- bad ordering of remote (`https://`) and local imports
- ambiguous notations that have different results when viewed as file paths or urls
- case sensitivity
- query params
- url anchors

We might add some more later, but I think this is already a good base set of issues it can detect.


romainmenke:
todo : urls must be lowercase because file systems can be case sensitive or case insensitive and not everyone on a given team might have the same type of file system

plugins-stylelint/no-invalid-at-import-rules-when-bundling/index.js
```diff
@@ -0,0 +1,138 @@
...
+
+	return index;
+}
+
+function parseAsURL(uri) {
+	const localProtocols = ['npm', 'node_modules'];
+
+	try {
+		const url = new URL(uri, 'https://454c178c-dadf-429e-8d9c-88e032826008.com');
+		if (url.host && url.host !== '454c178c-dadf-429e-8d9c-88e032826008.com') {
```
Antonio-Laguna, line 116: Why this domain? Clever though!

Antonio-Laguna (approved):
This LGTM! Nice to be able to lint these!

plugins-stylelint/no-invalid-at-import-rules-when-bundling/index.js
romainmenke: I wanted something that is extremely unlikely to be used in a real url and generating a new uuid is easy :)

_I really which there was some builtin API to detect if a given string is a full url or only a relative path._

PR #1084, opened 2023-08-15
remove support for `postcss-html`

romainmenke:
`postcss-html` never progressed to a point where we could actively add support for it in complex cases.

Dropping support entirely for now in `postcss-tape`.

This will prevent us from working around quirks that are unique to `postcss-html`, like having broken sourcemaps.


romainmenke:
superseded by : https://github.com/csstools/postcss-plugins/commit/c398a4a648cbb64cff39ca8c4b22308dd2678c5c

I found a smaller change.

PR #1090, opened 2023-08-21
`postcss-bundler`

romainmenke:
_part of https://github.com/csstools/postcss-plugins/pull/1060_

----

## A plugin pack for bundling CSS

The goal is to pass [these tests](https://github.com/romainmenke/css-import-tests) and to do for bundling CSS what `postcss-preset-env` does for fallbacks.

It uses :
- an embedded hard fork of `postcss-import`
- a new plugin to rebase `url()` values

I don't intend to include things best left to other tools :
- doing things with SVG's
- copying files to a `/dist` folder
- doing things with images
- ...

It is better if this tool adheres to these principles :
- zero config is a good config
- CSS authors don't have to re-learn how to write CSS
- clicking a file link in CSS in a code editor takes you to that file

### `postcss-import`

I initially hoped to be able to use `postcss-import` directly, but even after fully rewriting that plugin I am still not where I need/want to be. We would still need to override any and every possible thing to make it compatible with the goals I laid out here.

I think it is better to do a hard fork and strip out all the parts we don't intend to use.

After doing this work, I found it was much easier to correctly fix the last remaining issues to pass all tests.


Antonio-Laguna (approved):
This looks great!

Can't wait to start using this!

romainmenke:
Thank you for reviewing all this 🙇

PR #1092, opened 2023-08-24
`postcss-minify` : make it a bit faster and the output a bit smaller

romainmenke:
Changes : 
- remove a bit of extra whitespace around, but not within meaningful bits
- use a cache to reduce the amount of processing we need to do

This is now both the worst and the best JavaScript minifier when benchmarking.

Output is always the largest, but a significant chunk of this is the license and copyright comments that are being removed by other minifiers.

For very large sources this is about 2-3 times slower than esbuild/lightningcss.
But it will be faster in practice when a given stack is already running PostCSS anyway.

It is faster than any other JavaScript based minifier.
This is apples and oranges because those other packages are actually css optimizers and this plugin doesn't do any of those transforms.

I think it now strikes a good balance between :
- correctness
- compliance
- fast enough
- small enough

Screenshot 2023-08-24 at 18 42 16


plugins/postcss-minify/src/index.ts
```diff
@@ -75,86 +85,107 @@ function removeEmptyNodes(node: Container | Document): boolean {
 	return false;
 }
 
+function setSemicolon(node: AtRule | Rule) {
+	if (node.raws.semicolon) {
+		const last = node.last;
+		if (last?.type !== 'decl' || !last.variable) {
+			node.raws.semicolon = false;
+		}
```
romainmenke, line 93: Technically the last semicolon is always safe to omit, even for variables.
But out of an abundance of caution I am skipping them.

plugins/postcss-minify/src/index.ts
```diff
@@ -75,86 +85,107 @@ function removeEmptyNodes(node: Container | Document): boolean {
...
-						return;
-					}
+						if (node.type === 'rule') {
 
-					node.raws = {
-						before: minify(node.raws?.before),
-						after: minify(node.raws?.after),
-						between: minify(node.raws?.between),
-						semicolons: node.raws?.semicolons,
-					};
```
romainmenke, line 138: Assigning a completely new object to `raws` wasn't ideal.
- more memory allocations
- erasing unknown future fields

Especially the second part would have backfired. If PostCSS added a `foo` field in `raws` tomorrow we would always erase it and we would need to ship a new release.

plugins/postcss-minify/src/index.ts
Antonio-Laguna: Curious what made you go this way. I'll always agree with extra caution just curious about what could be the factor!

Antonio-Laguna (approved):
LGTM!

Loved the comparison table and the explanation around it!

PR #1093, opened 2023-08-27
contrast-color-function

romainmenke:
Add experimental support for `contrast-color()`.

I would like to use apca for this, but the license is too restrictive : https://github.com/Myndex/apca-w3/blob/master/LICENSE.md

I've asked about this here : https://github.com/Myndex/apca-w3/discussions/12

We might have to look into something like WCAG 2.1 instead


romainmenke:
Chrome wants to experiment with return values that aren't pure black/white.

If we create a fallback early people will use it and Chrome will get less feedback on their experiments.

We should wait until they are done experimenting.

PR #1098, opened 2023-08-29
css-color-parser : correctly clip values during parsing

romainmenke:
This was changed and I missed that.
Explains a bunch of weirdness I encountered and had been fighting against.

This is now valid after the spec changes :
- `hsl(50deg -10% 50%)`
- `hsl(50deg 200% 50%)`

There is no practical use for these and no one will be writing `hsl` like that.
But it affects computed values.

Doing this correctly allows us to finally defer gamut mapping to serialization.
Before this change we still had weird places where we had to do gamut mapping half way a computation to get the right result.


packages/css-color-parser/test/wpt/color-computed-relative-color.mjs
```diff
@@ -164,7 +175,7 @@ const tests = [
 	['hsl(from rebeccapurple h alpha l / s)', 'rgba(102, 101, 103)'],
 	['hsl(from rebeccapurple h l l / l)', 'rgba(102, 61, 143)'],
 	['hsl(from rebeccapurple h alpha alpha / alpha)', 'rgb(3, 3, 3)'],
-	['hsl(from rgb(20%, 40%, 60%, 80%) h l s)', 'rgba(77, 127, 178, 0.8)'],
+	['hsl(from rgb(20%, 40%, 60%, 80%) h l s)', 'rgba(77, 128, 179, 0.8)'],
```
romainmenke, line 178: Rounding is different, but better. This now matches WPT.

plugins/postcss-gradients-interpolation-method/test/basic.expect.css
```diff
@@ -383,7 +383,7 @@ to-clone {
 }
 
 .wpt-srgb-gradient {
-	background: linear-gradient(to right, rgb(255, 0, 0) 0%, rgb(230, 26, 0), rgb(204, 51, 0), rgb(179, 77, 0), rgb(153, 102, 0), rgb(128, 128, 0), rgb(102, 153, 0), rgb(77, 179, 0), rgb(51, 204, 0), rgb(26, 230, 0), rgb(0, 255, 0) 100%);
+	background: linear-gradient(to right, rgb(255, 0, 0) 0%, rgb(230, 26, 0), rgb(204, 51, 0), rgb(179, 77, 0), rgb(153, 102, 0), rgb(128, 128, 0), rgb(102, 153, 0), rgb(77, 179, 0), rgb(51, 204, 0), rgb(25, 230, 0), rgb(0, 255, 0) 100%);
```
romainmenke, line 386: different rounding, this is fine

PR #1099, opened 2023-09-01
numeric separators

romainmenke:


romainmenke:
CSSWG decided against this change : https://github.com/w3c/csswg-drafts/issues/9293#issuecomment-1709545683

PR #1118, opened 2023-09-12
improve docs

romainmenke:
This change makes the npm package scope and name more prominent by placing it higher in the readme.

see : https://github.com/csstools/postcss-plugins/discussions/1115


Antonio-Laguna (approved):
LGTM!

PR #1121, opened 2023-09-16
`postcss-initial`

romainmenke:
Was reported on Discord that the original plugin wasn't maintained and had some serious issues. On inspection I agree that the issues are sufficiently important to do a fork.


plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
```diff
@@ -174,7 +173,7 @@
 	min-width: 0;
 	opacity: 1;
 	orphans: 2;
-	outline: medium none invert;
+	outline: medium none currentcolor;
```
romainmenke, line 176: see : https://github.com/maximkoretskiy/postcss-initial/issues/21

plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
```diff
@@ -200,7 +199,6 @@
 	transform-origin: 50% 50% 0;
 	transform-style: flat;
 	transition: none 0s ease 0s;
-	unicode-bidi: normal;
```
romainmenke, line 203: see : https://github.com/maximkoretskiy/postcss-initial/issues/53

plugin-packs/postcss-preset-env/test/basic.autoprefixer.expect.css
```diff
@@ -144,14 +144,13 @@
...
 	column-gap: normal;
-	column-rule: medium none currentColor;
+	column-rule: medium none currentcolor;
 	column-span: 1;
 	column-width: auto;
 	content: normal;
 	counter-increment: none;
 	counter-reset: none;
 	cursor: auto;
-	direction: ltr;
```
romainmenke, line 154: see : https://github.com/maximkoretskiy/postcss-initial/issues/53

plugin-packs/postcss-preset-env/test/layers-basic.preserve.true.expect.css
```diff
@@ -785,6 +784,8 @@ h1.test-custom-selectors:not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):n
 }
 
 .unset:not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#):not(#\#) {
+	clip: auto;
+	clip: initial;
```
romainmenke, line 788: No idea why this changed in this way.
But it looks more correct.

I don't really understand where `clip: initial` is suddenly coming from. It is the correct fallback for `unset`, but I didn't make any change to add this. It should have always been there.

Maybe a bug in `postcss-initial` where it removes nodes and somehow I don't have the same bug?

plugins/postcss-initial/test/multivalue.expect.css
```diff
@@ -0,0 +1,7 @@
+div {
+	background-repeat: initial initial;
+}
+
+div {
+	background-repeat: initial no-repeat;
+}
```
romainmenke, line 7: `initial` is a css wide keyword.

Using this in a shorthand together with other values is just invalid and must be ignored.

The old plugin didn't ignore these and did weird other things.

I am following the spec and so ignoring as expected.

plugins/postcss-initial/test/negative.expect.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+c {
+	/* not ignored */
+	background: transparent none repeat 0 0 / auto auto padding-box border-box scroll;
+	background: initial;
+
+	/* ignored */
+	background: url(initial);
+}
```
romainmenke, line 21: see : https://github.com/maximkoretskiy/postcss-initial/issues/52

plugins/postcss-initial/src/properties.ts
```diff
@@ -0,0 +1,2 @@
...line-color','currentcolor'],['outline-style','none'],['outline-width','medium'],['overflow','visible'],['overflow-x','visible'],['overflow-y','visible'],['padding','0'],['padding-bottom','0'],['padding-left','0'],['padding-right','0'],['padding-top','0'],['page-break-after','auto'],['page-break-before','auto'],['page-break-inside','auto'],['perspective','none'],['perspective-origin','50% 50%'],['position','static'],['quotes','“ ” ‘ ’'],['right','auto'],['tab-size','8'],['table-layout','auto'],['text-align','left'],['text-align-last','auto'],['text-decoration','none'],['text-decoration-color','inherited'],['text-decoration-color','currentcolor'],['text-decoration-line','none'],['text-decoration-style','solid'],['text-indent','0'],['text-shadow','none'],['text-transform','none'],['top','auto'],['transform','none'],['transform-origin','50% 50% 0'],['transform-style','flat'],['transition','none 0s ease 0s'],['transition-delay','0s'],['transition-duration','0s'],['transition-property','none'],['transition-timing-function','ease'],['vertical-align','baseline'],['visibility','visible'],['white-space','normal'],['widows','2'],['width','auto'],['word-spacing','normal'],['z-index','auto']]);
```
romainmenke, line 1: The old plugin did a lot of templating/formatting at runtime.

This is now all pre-computed, so the plugin is smaller, faster and much simpler.

Because I am not using weird file/folder names this bug is also gone : https://github.com/maximkoretskiy/postcss-initial/issues/51

plugins/postcss-initial/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+	);
+
+	return {
+		postcssPlugin: 'postcss-initial',
+		Declaration(decl) {
+			if (decl.variable) {
+				return;
+			}
+
+			if (!IS_INITIAL.test(decl.value)) {
```
romainmenke, line 42: `initial` is a css wide keyword.

It must be the only token in a value.

There can be whitespace in front or after but nothing else.
regexp should be sufficient and faster than tokenizing again.

Antonio-Laguna (approved):
Looking great as usual!

I'm wondering if `plugins/postcss-initial/src/properties.ts` should be better formatted for ease of maintenance since it's a very long list as it stands. However, I don't see those values changing too much (if any) so feel free to ignore!

Thanks for this!! 🙇

romainmenke:
> I'm wondering if `plugins/postcss-initial/src/properties.ts` should be better formatted for ease of maintenance since it's a very long list as it stands.

Good point!

Makes it much easier to review mutations :)
I also spotted an error after making it readable, `quotes` was broken, so immediate profits 🎉

PR #1137, opened 2023-10-02
postcss-bundler : fix parser

romainmenke:


plugin-packs/postcss-bundler/src/postcss-import/lib/apply-conditions.ts
```diff
@@ -19,7 +19,7 @@ export function applyConditions(bundle: Array<Statement>, atRule: (defaults?: At
 					stmt.conditions[0].media,
 					stmt.conditions[0].supports,
 				)}`;
-			} else if (stmt.conditions.length > 1) {
+			} else {
```
romainmenke, line 22: This was a useless condition, ln 11 already guards against this.

PR #1140, opened 2023-10-02
postcss-nesting : spec compliance

romainmenke:
A bit too soon maybe, but we really need to drop the old way of transforming nesting.
People will start to encounter issues because of the default config in `postcss-preset-env`, which gives different results from what shipped in browsers.


romainmenke:
superseded by : https://github.com/csstools/postcss-plugins/pull/1305

Issue #1049, opened 2023-07-17
`@media (color-gamut: ...)` checks for color fallbacks


romainmenke:
```css
.foo {
  color: oklch(...)
}

/* becomes */

.foo {
  color: rgb(...);
  color: color(display-p3 ...);
}
```

But this doesn't take into account the hardware capabilities of the current device.

Ideal output :

```css
.foo {
  color: oklch(...)
}

/* becomes */

.foo {
  color: rgb(...);
}

@media (color-gamut: display-p3) {
  .foo {
    color: color(display-p3 ...);
  }
}
```

----

This requires similar tricks as progressive custom properties.

ai:
We can think about it as a polyfill, because current browsers violate CSSWG spec and do gamut correction in a wrong way.

So we will fix compatibility issue.

romainmenke:
Exactly :)

For me this is very similar to [`postcss-progressive-custom-properties`](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-progressive-custom-properties#postcss-progressive-custom-properties-)

```css
:root {
	/* fallback */
	--a-color: red;
	/* progressive enhancement */
	--a-color: oklch(40% 0.234 0.39 / var(--opacity-50));
}

/* becomes */

:root {
	--a-color: red;
}

@supports (color: oklch(0% 0 0)) {
	:root {
		--a-color: oklch(40% 0.234 0.39 / var(--opacity-50));
	}
}
```

But then not for CSS variables and with `@supports` but for all wide gamut colors and with `@media`.

romainmenke:
Hi @ai,

This shipped in `postcss-preset-env` `v9.2.0` :

https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-gamut-mapping#readme

[...]

PR #1144, opened 2023-10-08
postcss-gamut-mapping

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1049

```pcss
p {
	background-color: oklch(80% 0.05 0.39 / 0.5);
	color: oklch(20% 0.234 0.39 / 0.5);
	border-color: color(display-p3 0 1 0);
}

/* becomes */

p {
	background-color: oklch(80% 0.05 0.39 / 0.5);
	color: rgba(50, 0, 20, 0.5);
	border-color: rgb(0, 249, 66);
}

@media (color-gamut: rec2020) {
p {
	color: oklch(20% 0.234 0.39 / 0.5);
}
}

@media (color-gamut: p3) {
p {
	border-color: color(display-p3 0 1 0);
}
}
```

-----------

This does not "fix" massively out of gamut source colors so that they fit within `display-p3` or `rec2020`. Only a single `srgb` fallback is provided.

Providing fallbacks for each wide gamut option would be slow and would needlessly bloat the output.

I am making the assumption that authors have the tools to choose and write colors that fit within either `display-p3` or `rec2020` and that it is sufficient to provide an `srgb` fallback so that clipping is avoided. Clipping is the most harmful effect as it radically alters the color, away from the intention of the author.

I don't think we need to provide a tool to enforce that wide gamut colors are real colors or  that these fit within `rec2020`. Linters are better suited to that specific task.

_`rec2020` is currently the widest gamut expected to be supported by displays. Authoring colors that do not fit within `rec2020` doesn't make sense today. The chance that hardware beyond `rec2020` will become wide spread before browsers ship gamut mapping is small. If that happens we can adjust how this plugin works._

-----------

This also better integrates `postcss-progressive-custom-properties` into `postcss-preset-env`.

It is now part of the determined plugin order.

There will be some duplication in the output if a source requires many transforms.
The current form is the best balance between duplicate/redundant output and complexity in multiple plugins. 

_The only way to reduce duplication further is by bringing both gamut mapping fallbacks and progressive props to each and every plugin that might be related to colors. I don't see the point in doing that. A little bit of duplication is short term cost, eventually users will require fewer fallbacks._


plugin-packs/postcss-preset-env/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 # Changes to PostCSS Preset Env
 
-### Unreleased (patch)
+### Unreleased (minor)
 
-- Updated [`cssdb`](https://github.com/csstools/cssdb) to [`7.7.3`](https://github.com/csstools/cssdb/blob/main/CHANGELOG.md#773-october-05-2023) (patch)
+- Added `@csstools/postcss-gamut-mapping` [Check the plugin README](https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-gamut-mapping#readme) for usage details.
```
Antonio-Laguna, line 5: Should there be any mention to the progressive custom properties?

plugins/postcss-color-function/README.md
```diff
@@ -90,7 +90,7 @@ postcssColorFunction({ preserve: true })
 	--a-color: rgb(164, 49, 43);
 }
 
-@supports (color: color(srgb 0 0 0)) {
+@supports (color: color(display-p3 0 0 0)) {
```
Antonio-Laguna, line 93: Assuming this is intentional but looks off?

plugins/postcss-gamut-mapping/README.md
```diff
@@ -0,0 +1,76 @@
+# PostCSS Gamut Mapping [<img src="https://postcss.github.io/postcss/logo.svg" alt="PostCSS Logo" width="90" height="90" align="right">][PostCSS]
+
+[<img alt="npm version" src="https://img.shields.io/npm/v/@csstools/postcss-gamut-mapping.svg" height="20">][npm-url] [<img alt="CSS Standard Status" src="https://cssdb.org/images/badges/gamut-mapping.svg" height="20">][css-url] [<img alt="Build Status" src="https://github.com/csstools/postcss-plugins/workflows/test/badge.svg" height="20">][cli-url] [<img alt="Discord" src="https://shields.io/badge/Discord-5865F2?logo=discord&logoColor=white">][discord]
+
+```bash
+npm install @csstools/postcss-gamut-mapping --save-dev
+```
+
+[PostCSS Gamut Mapping] lets you use wide gamut colors with gamut mapping for specific displays following the [CSS Color 4 Specification].
```
Antonio-Laguna, line 9: Wondering if we should explain why this is important? Something in this line:

> This plugin ensures that colors get progressively enhanced at the closest gamut that the device supports.

By looking at the examples (unless you really know what's going on) it's somewhat difficult to understand why this would be useful. Thoughts?

Antonio-Laguna (approved):
@romainmenke this looks great to me! Left a few minor questions but great to go!

plugin-packs/postcss-preset-env/CHANGELOG.md
romainmenke: A changelog entry for `progressive custom properties` will be generated by the release planner.

I don't think we need to add more detail here because users of `postcss-preset-env` are unlikely to know that this plugin is part of the inner workings :)

plugins/postcss-color-function/README.md
romainmenke: Yes this is correct :)

It's more optimal to group the checks a bit instead of having a unique supports check per color space.

plugins/postcss-gamut-mapping/README.md
romainmenke: Good point!

I've added a bit more of an explainer to the docs.
I hope this is clear?

romainmenke: https://github.com/csstools/postcss-plugins/blob/1e3febcf007058aaba75d77987a62792f34f1cfd/plugins/postcss-gamut-mapping/README.md

Antonio-Laguna: This looks lovely!

PR #1148, opened 2023-10-10
postcss-cascade-layers : performance improvements

romainmenke:
This reduces the number of entire AST walks from 4 to 1 before the first early abort when users do not actually have any cascade layers in their source code.

Also speeds up the plugin a little bit when users do use cascade layers.


Antonio-Laguna (approved):
LGTM!

PR #1161, opened 2023-10-28
postcss-bundler : do not shift up `@import` statements 

romainmenke:
`postcss-import` had multiple ways in which it re-ordered CSS.
This change removes the final instance in which this could occur.

context :
- `@layer` statements can precede `@import` statements
- declaration order of cascade layers is important
- cascade layers can be conditional (e.g. only when `@media screen`)
- `@layer` statements that precede `@import` statements can not be written as conditional statements, the CSS syntax/grammer does not allow this.

This issue is irrelevant when all imported stylesheets are local resources.
In that case all imports will be resolved and `@media screen { @layer a; }` works just fine.
It will never be followed by an `@import` statement.

----------

However when some imported stylesheets are external this stops working :

```css
@media screen { @layer a; }

@import 'https://example.com/style.css'; /* will never be loaded */
```

----------

`postcss-import` changes the order to :

```css
@import 'https://example.com/style.css';

@media screen { @layer a; }
```

But that changes the order of declaration and potentially the order of cascade layers.

----------

This change fixes this issue by base64 encoding these `@layer` statements and re-writing them as data url imports.

```css
/* ./a.css */
@import 'data:text/css;base64,QGxheWVyIGIsIGE=' print;
@import url("http://localhost:8080/green.css") screen;
```

-----------

This is an edge case of an edge case :
- very few people use external stylesheets
- very few people use conditional imports
- very few people use cascade layers on imports

"no one" uses all three together in the specific combo that triggers this issue.

-------------

The real benefit is that by fixing this issue we can fully eliminate the re-ordering that `postcss-import` does.

External resources must always be imported first when bundling.
We must never do any magic that enables users to mix external and local resources because we can never implement this correctly.


plugin-packs/postcss-bundler/.tape.mjs
romainmenke, line 1: Some new tests where brought over from `css-import-tests` while at the same time some warnings were removed from the plugin.

We just happen to end up with the same number of tests that do have a warning.

plugin-packs/postcss-bundler/src/postcss-import/lib/apply-styles.ts
```diff
@@ -1,16 +1,21 @@
...
+import { Stylesheet, isImportStatement, isNodesStatement } from './statement';
 
-export function applyStyles(bundle: Array<Statement>, styles: Root | Document) {
+export function applyStyles(stylesheet: Stylesheet, styles: Root | Document) {
 	styles.nodes = [];
 
+	if (stylesheet.charset) {
+		stylesheet.charset.parent = undefined;
+		styles.append(stylesheet.charset);
+	}
```
romainmenke, line 10: `@charset` must always be first.

It doesn't make sense to add it in a loop where theoretically it could be inserted later.

plugin-packs/postcss-bundler/src/postcss-import/lib/base64-encoded-import.ts
```diff
@@ -0,0 +1,24 @@
...
+			`@import ${params}`,
+		).toString('base64')}' ${formatImportPrelude(
+			condition.layer,
+			condition.media,
+			condition.supports,
+		)}`;
+	}
+
+	return params;
+}
```
romainmenke, line 29: pretty gnarly bit of magic.
Better to have it in its own file.

plugin-packs/postcss-bundler/src/postcss-import/lib/parse-styles.ts
```diff
@@ -40,42 +40,37 @@ export async function parseStyles(
...
-			bundle.push(stmt);
+
+			stylesheet.statements.splice(i, 1, ...stmt.stylesheet.statements);
+			i--;
+			continue;
 		}
-	});
+	}
 
-	return charset ? [charset, ...imports.concat(bundle)] : imports.concat(bundle);
```
romainmenke, line 73: This is what caused the re-ordering of nodes.

`imports` and `bundle` each contained nodes that should have been interleaved.

`postcss-import` assigns each to different "buckets" and then concats these, effectively shifting up all `@import` statements.

plugin-packs/postcss-bundler/src/postcss-import/lib/post-process.ts
romainmenke, line 1: This function base64 encodes the `@layer` statements that precede `@import` statements.

In the future more things might be allowed before `@import`.
If that happens we must handle these in the same way in this function.

plugin-packs/postcss-bundler/src/postcss-import/lib/base64-encoded-import.ts
Antonio-Laguna: Thanks for the thorough comments here!

Antonio-Laguna (approved):
This LGTM!

Thanks for tackling this!

romainmenke:
Thank you for reviewing 🙇

PR #1188, opened 2023-12-01
base-cli : make the build step uniform with other packages

romainmenke:
The `base-cli` package used `tsc` instead of rollup and had a different TypeScript config.

Having a uniform setup is easier to maintain.

----

see : https://github.com/csstools/postcss-plugins/pull/1170


Antonio-Laguna (approved):
This LGTM though I suspect this is part of the Housekeeping thing as well?

romainmenke:
> though I suspect this is part of the Housekeeping thing as well?

Yeah, I started splitting it up because that would make it easier to review :)

PR #1189, opened 2023-12-01
cli : add relative-color-syntax

romainmenke:
`@csstools/postcss-relative-color-syntax` was listed as a dependency in `package.json` but it wasn't actually wired up in code.


Antonio-Laguna (approved):
Nice catch, wasn't this included in housekeeping too?

Issue #1031, opened 2023-07-04
packages provide wrong types under node16 module resolution


not-my-profile:
## node16 from ESM: type errors

Using the packages from [Deno](https://deno.land/) with TypeScript leads to type errors, but these errors can also be observed with `tsc` and Node.JS when using the new node16 module resolution:

```
$ npm i postcss-nesting
$ echo '{"type": "module"}' > package.json
$ echo "import postcssNesting from 'postcss-nesting'; postcssNesting();" > example.ts
$ tsc example.ts --moduleResolution node16 --lib es2020
example.ts:1:47 - error TS2349: This expression is not callable.
  Type 'typeof import("/tmp/example/node_modules/postcss-nesting/dist/index")' has no call signatures.

1 import postcssNesting from 'postcss-nesting'; postcssNesting();
                                                ~~~~~~~~~~~~~~
```

## node16 from CJS: runtime error

```
$ npm i postcss-nesting
$ echo '{"type": "commonjs"}' > package.json
$ echo "import postcssNesting from 'postcss-nesting'; postcssNesting();" > example.ts
$ tsc example.ts --moduleResolution node16 --lib es2020
$ node example.js
/tmp/example/example.js:4
(0, postcss_nesting_1.default)();
                              ^

TypeError: (0 , postcss_nesting_1.default) is not a function
    at Object.<anonymous> (/tmp/example/example.js:4:31)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.14.2
```

## What to do

As [Are the types wrong? explains](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md):

> A golden rule of declaration files is that if they represent a module—that is, if they use import or export at the top level—they must represent exactly one JavaScript file. They especially cannot represent JavaScript files of two different module formats.

So the `exports` like the following are wrong since they provide one `types` for both ESM and CJS:

https://github.com/csstools/postcss-plugins/blob/aaf88074689f64fdca1db391fffc5f3c337ef2da/plugins/postcss-nesting/package.json#L39-L46

[...]

PR #1197, opened 2023-12-04
fix `esm` vs. `cjs` typings

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1031

------

I am not sure that I like this yet.
It adds a lot of complexity.

General approach :
- bundle and treeshake the type definitions
- do a string find/replace to convert the `esm` typings to `cjs`.

Bundling and treeshaking the types is critical to make this work.
All types for a given plugin need to be in a known place and shape to do the string find/replace.

This obviously is very brittle and depends on a bunch more 3rd party software.


romainmenke:
nope, this remains unsolvable.
No matter which way you start you always end up with single bit that is broken.

If you follow this route to it's end you find that there is just no way to express default exports in the correct way in both commonjs and esmodules and do this automated, not with manually crafted types.

I don't think we can have correct types as long as we dual ship esm and cjs.

PR #1199, opened 2023-12-05
release plan : smaller published footprint

romainmenke:
A significant portion of each package is information that isn't useful to running the plugins.

- dev tooling in `package.json`
- metadata in `package.json`
- older `CHANGELOG.md` entries

Screenshot 2023-12-05 at 17 43 19

Especially the `CHANGELOG.md` is something that annoys me.
We have really nice changelog entries now but they are verbose and these files keep growing over time.

Only including the last few entries when publishing to npm ensures that our packages have a somewhat stable size.

I have no strong preference for the number of entries.
I picked 3 because that seemed like a sane number.

A link is added to the full changelog on GitHub.
So anyone who needs to see the details of more releases can do so easily.

### DIFF
--- a/.github/bin/release-plan/npm-prepublish-scripts.mjs
+++ b/.github/bin/release-plan/npm-prepublish-scripts.mjs
@@ -1,0 +1,35 @@
+import { spawn } from 'child_process';
+import { platform } from 'process';
+
+export async function npmPrepublishScripts(packageDirectory, packageName) {
+	await new Promise((resolve, reject) => {
+		if (process.env.DEBUG) {
+			resolve(true);
+
+			return;
+		}
+
+		const publishCmd = spawn(
+			'npm',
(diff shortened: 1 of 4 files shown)


.github/bin/release-plan/npm-prepublish-scripts.mjs
```diff
@@ -0,0 +1,35 @@
+import { spawn } from 'child_process';
+import { platform } from 'process';
+
+export async function npmPrepublishScripts(packageDirectory, packageName) {
```
romainmenke, line 4: After deleting `scripts` from `package.json` the pre publish scripts will be gone.

So we need to run these manually before we minify `package.json` for publishing.

Antonio-Laguna (approved):
This is brilliant! LGTM!

PR #1200, opened 2023-12-05
fix types

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1031

------

This change makes a different tradeoff.

Instead of trying to have correct types for all users and failing to do so, we could choose to only provide types for part of our users.

In practice our current typings and the entries in `package.json` are broken.
I think they only work for those who write es module syntax while configuring their TypeScript project for very old node.

I think it's better to make sure it works correctly for modern TypeScript in es modules.
I am making the assumption that those who value typings in build tools are much more likely to be also using es modules for build scripts.

Focussing on es modules, modern TypeScript and modern node has the benefit of being future proof.

-------

The general idea remains the same :
- the code is bundled, so the types need to be bundled
- https://api-extractor.com seems to do the trick and this is created by Microsoft, making it a bit safer to add as an extra dependency
- only the public facing types are exposed

-------------

I don't think it is needed to review this change in it's entirety, just too many generated files changed.

example change :

`color-helpers`:
- main : https://github.com/csstools/postcss-plugins/tree/main/packages/color-helpers/dist
  - unbundled typings
  - contains internal typings
  - the same typings are advertised as commonjs and as es module typings https://github.com/csstools/postcss-plugins/blob/main/packages/color-helpers/package.json#L30-L40
- this : https://github.com/csstools/postcss-plugins/tree/fix-types--persistent-puffin-26bee81a68/packages/color-helpers/dist
  - bundled typings
  - only contains public typings
  - typings are only advertised for es modules https://github.com/csstools/postcss-plugins/blob/fix-types--persistent-puffin-26bee81a68/packages/color-helpers/package.json#L30-L43


.github/bin/generate-docs/api-documenter.mjs
romainmenke, line 1: This file is unused at this time, but I am hoping to use it to resolve : https://github.com/csstools/postcss-plugins/issues/1201

Actually running it generates hundreds of markdown files with API docs.
So I want to save that for a future PR.

Antonio-Laguna (approved):
LGTM! I did a general overview as this became nasty to review!

romainmenke:
> I did a general overview as this became nasty to review!

I feared as much :/
Luckily no test expect files changed and our e2e tests are more accurate now.

Thank you for checking all this 🙇

Issue #1202, opened 2023-12-07
Bug: replaceComponentValues() continues iterating replaced values


nex3:
```js
const {tokenize} = require("@csstools/css-tokenizer"); // @csstools/css-tokenizer is a peer dependency. 
var parser = require("@csstools/css-parser-algorithms")

const value = parser.parseComponentValue(tokenize({css: "foo(bar(baz))"}));
parser.replaceComponentValues([[value]], node => {
  console.log(node);
  if (parser.FunctionNode.isFunctionNode(node) && node.getName() === 'bar') {
    return new parser.TokenNode(['ident-token', 'x', node.value[2], node.value[3], {value: 'x'}]);
  }
});
```

([Runkit](https://runkit.com/embed/bowwkrpt693e))

In line with other AST-walking and -replacement APIs, I would expect that if I replace a parent node in the AST, the replacement algorithm will stop traversing any of that node's children. (Note that this is how PostCSS's `walk()` and `replaceWith()` interact.) Instead, `replaceComponentValues()` continues traversing the removed nodes. This is extra unnecessary work and can even cause data corruption if, for example, information is being stored about nodes encountered during the traversal.

romainmenke:
Hi @nex3,

Thank you for reaching out about this.

I think the root cause is how we `walk` AST's.
This is used in `parseComponentValue`

The current `walk` function implementations roughly work like this :
1. iterate child nodes
2. call the walker callback for each child node
3. check if a child node is walkable
4. walk each child node

The issue you are encountering is between `2.` and `4.`.
We don't check that the child node is still a child node of the current AST tree after `2.`

I think we can trivially check in `3.` that a child node is not only walkable, but also still part of the same parent node on which `walk` was last called.

In your case it would no longer be part of the same parent and `baz` would not be visited.

----------

But it also introduces another question.


For this tree :

```
A
  - 1
  - 2
    - I
  - 9
```

If we replace `A.2` with this tree :

```
3
  - III
```

We form :

```
A
  - 1
  - 3
    - II
  - 9
```

We can prevent visiting `A.2.I` with a new check, but is it expected that `A.3` and/or `A.3.II` are visited?

There is no plugin system here as there is in PostCSS.
So I assume that there is always only a single walker callback and that the AST being inserted is always final and doesn't need further manipulation in the same AST walk.

So I think it's fine to assume that we don't want to visit `A.3` and `A.3.II` and that the next callback would be on `A.9`?

[...]

PR #1203, opened 2023-12-07
parsers : do not walk sub trees of removed nodes and make mutations during AST walks safe

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1202
fixes : https://github.com/csstools/postcss-plugins/issues/1204


Antonio-Laguna (approved):
LGTM!

great stuff!

PR #1212, opened 2023-12-11
update cssdb

romainmenke:


plugins/postcss-slow-plugins/.tape.mjs
```diff
@@ -1,13 +1,26 @@
...
+const variableSpeedCreator = (delay) => {
+	return {
+		postcssPlugin: `postcss-${delay}ms`,
+		Once: async () => {
+			await new Promise((resolve) => setTimeout(resolve, delay));
+		}
+	}
+}
+
+variableSpeedCreator.postcss = true;
```
romainmenke, line 13: `postcss-preset-env` is a good case to test the slow plugin detection.
But `postcss-preset-env` also changes frequently.

We don't want to have to update the test expect files for the slow plugins feature each time `postcss-preset-env` changes.

Using abstract plugins with an arbitrary delay is also a good way to make sure it works.

PR #1237, opened 2024-01-01
`contrast-color()`

romainmenke:
https://drafts.csswg.org/css-color-5/#contrast-color

`contrast-color( <color> max )`

The `max` keyword indicates that the most contrasting color should be returned. Either white or black.

Omitting that keyword is allowed in the specification but it is purposefully left vague what the behavior should be in that case. This leaves room for browser engines to experiment with this.

I don't want to muddy the waters by providing a polyfill.
If our users want to use `contrast-color()` without `max` then they should get the real behavior so that they can also give feedback to browsers.

Another reason not to implement this, is because I would like to tweak the outcome a bit depending on the matching `prefers-contrast` media query. That can only be done in a PostCSS plugin, not in the color parser package.

-------

> UAs are advised to not simply use the WCAG 2.1 section 1.4.3 Contrast (Minimum) contrast ratio algorithm to decide between light and dark colors, as it has several known issues. However, colors returned by this function should still meet the WCAG 2.1 section 1.4.3 Contrast (Minimum) for AA large text, as many authors need to meet legal requirements that mandate this.

This is tricky because there currently isn't anything better.
There are many candidates but there isn't consensus.
Things like APCA are really promising but we really cannot use that because of the attached license. 

So for now we are stuck with WCAG 2.1


romainmenke:
There is now a specification for this, so we can proceed with work on this : https://drafts.csswg.org/css-color-5/#contrast-color

PR #1267, opened 2024-01-31
Add `flattenNestedSelector`

romainmenke:
See : https://github.com/stylelint/stylelint/issues/6234

Stylelint has a bunch of rules which analyse selectors in a more analytical and less functional way.
For these rules it isn't relevant that the final selector is valid, or an equivalent.
It is much more important that it contains the right parts and the correct number of duplications. Just wrapping everything with `:is()` does not work there.

-------------

Name needs bikeshedding.

I want this to communicate that the outcome is the combination of the current and parent selector by using `&` as insertion points.

While at the same time making it clear that the result is not valid or equivalent.

Maybe :
- `glueNestedSelector`
- `insertParentSelector`
- `combineNestedSelector`
- ...


Antonio-Laguna (approved):
This LGTM! I think I'd go with either combine or flatten maybe?

romainmenke:
I like `flatter`!

ybiquitous:
Hi, I love this work! 😄 

For naming. The [CSS Nesting spec](https://drafts.csswg.org/css-nesting/#nest-selector) uses "desugar" for `:is()`, too. Is this confusing?

romainmenke:
> For naming. The [CSS Nesting spec](https://drafts.csswg.org/css-nesting/#nest-selector) uses "desugar" for :is(), too. Is this confusing?

Yup, that is confusing, and also why I am looking for a better name :)

ybiquitous:
I see. For now, +1 `flattern`. There may be a better name, though. 😅

ehoogeveen-medweb:
`desugar` is usually used to mean "translate to equivalent, older syntax" so I agree that it's confusing/wrong.

`flatten` sounds okay to me - it doesn't necessarily imply equivalence, since similar functions like `Array.prototype.flat` in JavaScript also aren't reversible (in that case, you throw away information about the size of the nested arrays). The downside is that it isn't clear from the name what information gets retained.

Other options that come to mind for me (not necessarily better than `flatten`):
* `concat` (or `concatenate`) - I don't think this makes it clear how `&` is handled.
* `substitute` - since IIUC you're essentially recursively substituting parent rules into the `&` selectors of their children.
* `resolve` - this sounds too much like it implies equivalence to me.

romainmenke:
`flatten` it is!

Thank you everyone for the feedback 🙇

PR #1273, opened 2024-02-09
fix: missing "types" in package.json for css-tokenizer

amertak:
Description:

This fixes missing typings when installing the package using npm, yarn, pnpm, etc...
Types are there, they are just not visible.

### DIFF
--- a/packages/css-tokenizer/package.json
+++ b/packages/css-tokenizer/package.json
@@ -30,6 +30,7 @@
 	"type": "module",
 	"main": "dist/index.cjs",
 	"module": "dist/index.mjs",
+	"types": "dist/index.d.ts",
 	"exports": {
 		".": {
 			"import": {


romainmenke:
Hi @amertak,

That `types` is left empty is by design.
We do not support typings for common js, only for es modules.

romainmenke:
@amertak Can you confirm that types are only missing if your project is a commonjs project?

amertak:
@romainmenke Trying to make it work but with no success.
Our project is esm.
I have `"type": "module"` in package.json.

romainmenke:
Can you create a minimal reproduction as a github repo?

amertak:
We have a problem in `moduleResolution` set to `node` in tsconfig. If changed to `bundler` types work.

romainmenke:
Can you also try with `node16` ?

`node` (if I recall correctly) is actually commonjs

romainmenke:
Always happy to help 🙇

PR #1280, opened 2024-02-14
`postcss-is-pseudo-class` Process :matches in addition to :is

carlosjeurissen:
Considering:
- :matches() was the predecessor to :is()
- :matches() was implemented in Safari
- :matches() is still supported in Safari
- it is fairly easy to add support for :matches()

This PR makes sure `postcss-is-pseudo-class` plugin processes selectors which include `:matches()` as if they are `:is()`.

No plans to add `:any()`, `:-moz-any()` or `:-webkit-any()` as they have a different specificity handling compared to :is() and :matches().

### DIFF
--- a/plugins/postcss-is-pseudo-class/src/index.ts
+++ b/plugins/postcss-is-pseudo-class/src/index.ts
@@ -24,7 +24,7 @@ export type pluginOptions = {
 	specificityMatchingName?: string
 };
 
-const HAS_IS_PSEUDO_REGEX = /:is\(/i;
+const HAS_IS_PSEUDO_REGEX = /:(is|matches)\(/i;
 
 const creator: PluginCreator<pluginOptions> = (opts?: pluginOptions) => {
 	const options = {
--- a/plugins/postcss-is-pseudo-class/src/split-selectors/complex.ts
+++ b/plugins/postcss-is-pseudo-class/src/split-selectors/complex.ts
@@ -7,7 +7,7 @@ import { isPseudoInFirstCompound } from './complex/is-pseudo-in-first-compound';
 
 export default function complexSelectors(selectors: Array<string>, pluginOptions: { onComplexSelector?: 'warning' }, warnOnComplexSelector: () => void, warnOnPseudoElements: () => void) {
 	return selectors.flatMap((selector) => {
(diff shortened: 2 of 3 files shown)


romainmenke:
Hi @carlosjeurissen,

Why would we want to support `:matches`?
Isn't it a dead end?

From our perspective it makes more sense that CSS authors do a find/replace in their codebase for `:matches` -> `:is`.

Then we don't have to keep supporting `:matches` indefinitely here :)

carlosjeurissen:
@romainmenke Thanks for your quick reply! I fully understand and agree we should all leave `:matches` behind and use `:is` always.

The reasoning is to be able to pass legacy / not yet upgraded code without having to have a developer deal with :matches and other legacy aspects first.

As dealing with :matches within this plugin seems to the most straight forward approach and supporting it seems fairly effortless.

However if dealing with `:matches()` is not in line with this plugin and/or the cost is too high I fully respect that!

romainmenke:
@carlosjeurissen thank you for your thoughtful reply.

Our focus is indeed purely on following the specifications and `:matches()` is no longer a part of that. Removing a plugin when all browsers support a feature must always be possible. And when you write `:matches()` and remove this plugin it would suddenly and unexpectedly break your project in non-Webkit browsers.

We are however sympathetic to your example :)
Maybe you can create a new plugin that only replaces `:is()` with `:matches()`?

Breaking changes are always very complex for us because CSS is so hard to test in general.
So when we remove or alter things it often causes production issues. This is why I don't think it is a good idea to add something that we already know must be removed again in the future.

carlosjeurissen:
@romainmenke Sure, totally understand and respect that especially considering the amount of projects relying on postcss-preset-env!

Creating a plugin for this was indeed a plan B. Which can be found here:
https://www.npmjs.com/package/postcss-matches-is-pseudo-class

To get the same behaviour as in this PR, do the following in your postcss config:
```js
[
  require('postcss-matches-is-pseudo-class')({ preserve: false }),
  require('@csstools/postcss-is-pseudo-class')
]
```

Make sure `postcss-matches-is-pseudo-class` is run before `postcss-is-pseudo-class`.

Closing this PR.

PR #1282, opened 2024-02-14
postcss-tape : type checker

romainmenke:
_Not sure about this one yet, but it seemed useful._

It's relatively easy to mess up a single prop assignment that isn't immediately visible to us in our own tests.

```js
rule.selector = ['.foo', '.bar'];
```

That would still serialize fine because an array of strings happens to concat with `,` by default in JavaScript.

But another plugin doing `selector.replace(...)` would get an on such a node.

see : https://github.com/postcss/postcss-import/issues/557


romainmenke:
Going to close for now, we already have TypeScript to guard against the same underlying issue.

Issue #1143, opened 2023-10-07
Plugin for `light-dark()`


MrHBS:
### What would you want to propose?

A PostCSS plugin for `light-dark()` function added in `css-color-5`

### Suggested solution

```css
:root {
  —foo: light-dark(bar, baz);
}
```
becomes:
```css
:root {
  —foo: bar;
}

@media (prefers-color-scheme: dark) {
  :root {
    —foo: baz;
  }
}
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this feature?

- [ ] I'm willing to open a PR

romainmenke:
Hi @MrHBS,

Thank you for requesting this feature.

If I recall correctly the `light-dark()` function is tied to the `color-scheme` property, not the `prefers-color-scheme` media query : https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

We might be able to do this with a space toggle.

romainmenke:
[You can see the difference between the property and media query in this ](https://drafts.csswg.org/css-color-adjust-1/#used-color-scheme)

> To determine the used color scheme of an element:
> If the user’s [preferred color scheme](https://drafts.csswg.org/css-color-adjust-1/#preferred-color-scheme), as indicated by the prefers-color-scheme media feature, is present among the listed [color schemes](https://drafts.csswg.org/css-color-adjust-1/#color-scheme), and is supported by the user agent, that’s the element’s [used color scheme](https://drafts.csswg.org/css-color-adjust-1/#used-color-scheme).
> 
> Otherwise, if the user has indicated an overriding preference for their chosen color scheme, and the [only](https://drafts.csswg.org/css-color-adjust-1/#valdef-color-scheme-only) keyword is not present in [color-scheme](https://drafts.csswg.org/css-color-adjust-1/#propdef-color-scheme) for the element, the user agent must [override the color scheme](https://drafts.csswg.org/css-color-adjust-1/#override-the-color-scheme) with the user’s [preferred color scheme](https://drafts.csswg.org/css-color-adjust-1/#preferred-color-scheme). See [§ 2.3 Overriding the Color Scheme](https://drafts.csswg.org/css-color-adjust-1/#color-scheme-override).
> 
[...]

PR #1283, opened 2024-02-15
postcss-light-dark-function

romainmenke:
fixes : https://github.com/csstools/postcss-plugins/issues/1143
readme : https://github.com/csstools/postcss-plugins/blob/3324ac0501ce1ed13733d2eec91b4e0b1d7ca6c9/plugins/postcss-light-dark-function/README.md

This plugin is quite involved because it needs to expose the value of `color-scheme` as a toggle for the `light-dark()` function elsewhere. Because it covers multiple runtime features and isn't purely syntactic sugar I also added puppeteer tests to make sure our fallback matches WPT.

EDIT : 
~~It seems that there are some issues with inheritance, investigating further...~~

Edit 2:

Inheritance with custom props now works as expected.


plugins/postcss-light-dark-function/src/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+						if (decl.variable && decl.parent) {
+							const variableInheritanceRule = rule({
+								selector: '& *',
+								source: decl.source,
+							});
+
+							variableInheritanceRule.append(decl.clone({ value: modified }));
+							decl.parent.append(variableInheritanceRule);
+						}
```
romainmenke, line 102: With nesting the parent might be an atrule, so we can't just clone that and manipulate its selector.

But because of nesting we can just insert a `& *` rule and it will be fine.

First time we get to use nesting to provide fallbacks instead of providing fallbacks for nesting itself 🎉

Antonio-Laguna (approved):
This LGTM!

Awesome work here! Specially love the browser tests!

romainmenke:
Thank you for reviewing this @Antonio-Laguna 🙇

PR #1286, opened 2024-02-17
utilities

romainmenke:
We discussed a few times before how we are duplicating specific bits quite a lot.

These parts do not form a coherent package on themselves that could be generally useful, but they are specific to how we write a lot of our PostCSS plugins. 

I chose to keep this really abstract and vague with `@csstools/utilities`.

-----

Were there any other functions we have should move that you can think of?


Antonio-Laguna (approved):
LGTM!

Nice to get this refactored and also great to have those Regexes with the `\b`!

PR #1287, opened 2024-02-18
Fix specificity calculation of any pseudo classes

carlosjeurissen:
The specificity of `:-moz-any()` and `:-webkit-any()` is always the same as b-level / class-level specificity of 1.

Since `:any()` is never implemented this can not be tested, yet if it is used at all, is used as equivalent of `:-moz-any()` and `:-webkit-any()`. So treating it as having the same specificity calculations.


romainmenke:
Hi @carlosjeurissen,

Do you have references for this change?

For Gecko for example it seems that the rollout of `:-moz-any()` was gradual.
It first shipped with the specificity of one class selector and then the behavior of the current `:is()` selector in regard to specificity was discussed but that never shipped for `:-moz-any`.

[However `:-moz-any()` still exists today and is an alias of `:is()` and it now also has the same specificity properties as `:is()`.](https://bugzilla.mozilla.org/show_bug.cgi?id=561154#c5)

You can test this with https://codepen.io/romainmenke/pen/XWGGNOG?editors=1100

`:-webkit-any()` does still appear to have the old specificity mechanic.
Maybe this change should be limited to `:-webkit-any()` and optionally `:any()`?

carlosjeurissen:
@romainmenke good point and thanks for double checking! Also saw the bug and was under the impression they didn't push the change yet. My bad.

Tested in Safari and seems it shows the same behaviour as in Chrome. Updated the PR to reflect the Firefox exception.

Now we are talking about this, seems the specificity of an empty `:is` is always 0. Is there a reference to this? It seems to result into an invalid selector so seems irrelevant anyway.

romainmenke:
I chose to use `0` for an empty `:is()` because of this statement:
> The specificity of the :is() pseudo-class is replaced by the specificity of its most specific argument.

https://drafts.csswg.org/selectors/#matches

romainmenke (approved):
Thank you for this @carlosjeurissen 🙇 

Awesome to have some feedback on the results of this package.

Issue #1297, opened 2024-02-23
[postcss-nesting] Nested mixins declaration order


pciarach:
### Bug description

Hello!

We have recently upgraded `postcss` and its plugins (including `postcss-nesting`) and noticed unexpected style orders in some places.
After some digging and debugging, it turned out that the differences come from `postcss-nesting` and nested `@mixin` rules interaction implemented in the `groupDeclarations` function.

Our previous assumption was that all styles from mixins would be placed exactly where the mixin was used. However, this behavior has been affected by the new grouping and reordering mechanism.

I’ve attempted to address this issue in the forked repo, but before submitting a pull request, I’d like to confirm whether this behavior is intentional or unintended. If it’s intentional, any hints/details will be welcome 😃 
Link to my changes: https://github.com/csstools/postcss-plugins/compare/main...pciarach:postcss-plugins:fix/nested-mixins-reorder

### Source CSS

```
@define-mixin mixinToOverride {
	display: flex;
}

@define-mixin mixinWithDecl {
	color: blue;
}

@define-mixin nestedMixins {
	@mixin mixinWithDecl;
	@mixin mixinToOverride;

	&:disabled {
		color: white;
	}
}

a {
	color: yellow;

	& b {
		@mixin nestedMixins;
		display: inline-flex;
	}
}

```

### Expected CSS

```
a {
	color: yellow;
}

a b {
	color: blue;
	display: flex;
	display: inline-flex;
}

a b:disabled {
	color: white;
}
```

### Actual CSS

```
a {
	color: yellow;
}

a b {
	color: blue;
	display: inline-flex;
	display: flex;
}

a b:disabled {
	color: white;
}
```

### Does it happen with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

None

### What plugin are you experiencing this issue on?

PostCSS Nesting

### Plugin version

12.0.3

### What OS are you experiencing this on?

Windows

### Node Version

20.11.0

### Validations

- [X] Follow our [Code of Conduct](https://github.com/csstools/postcss-plugins/blob/master/CODE_OF_CONDUCT.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

### Would you like to open a PR for this bug?

- [X] I'm willing to open a PR

romainmenke:
Hi @pciarach,

I think you are right that this is a bug.
It's always fine to open a pull request by the way.

[...]

PR #1298, opened 2024-02-23
Fix/nested mixins reorder

pciarach:
Fixes: #1297

This is an attempt to preserve the same order of style declarations in the case of nested mixins.

I think that the previous assumption that the mixin after rules or at-rules will resolve to more rules or at-rules is mostly valid because of the CSS nesting module draft (https://drafts.csswg.org/css-nesting/#mixing). However, this draft does not cover non-standard `mixin` rules and thus should be excluded from the grouping logic.


pciarach:
Thank you @romainmenke 

It looks really nice. I've also added a minor change to improve readability a bit.

When it comes to the additional test - you are right, the similar thing is covered by other mixin tests.

romainmenke (approved):
Thank you so much for this @pciarach,
This is a complex feature and the test cases and pull request really helped to make it better 🎉

pciarach:
@romainmenke Thanks for the help and such a quick responses!

PR #1304, opened 2024-02-26
`postcss-preset-env` : array options

romainmenke:
I've often run into issues when providing options to plugins.
Each time you set any plugin options the plugin will also always run.
This largely defeats the purpose of `postcss-preset-env`.

Although we try very hard to make the default settings also the best settings we sometimes get it wrong or we don't yet want to change the default because that would be a breaking change.

I've taken inspiration from stylelint as they also have an array structure for plugin options.

Any feedback would be more than welcome!
Is there a better way or another parallel? (maybe babel, core-js)


Antonio-Laguna (approved):
This LGTM! Nice work! Having so many tests makes these features go smoother! 🙇

romainmenke:
Thank you for reviewing this 🙇

PR #1305, opened 2024-02-28
nesting editions

romainmenke:


Antonio-Laguna (approved):
Awesome idea and great execution! Love this path forward!

plugins/postcss-nesting/src/editions/2024-02/index.ts
```diff
@@ -4,24 +4,11 @@ import walk from './lib/walk.js';
...
-		// Default options
-		{
-			noIsPseudoSelector: false,
-		},
-		// Provided options
-		opts,
-	);
-
-	if (options.noIsPseudoSelector) {
-		throw new Error('The `noIsPseudoSelector` option is no longer supported. Migrate your CSS to use the latest CSS nesting syntax.');
```
romainmenke: We can't throw here because `postcss-preset-env` sets this option when a user targets very old browsers.

That behavior already has a comment to remove it in a next major.

romainmenke:
Thank you reviewing @Antonio-Laguna 🚀

PR #1316, opened 2024-03-08
Fix typo in `postcss-nesting` documentation

meduzen:


romainmenke (approved):
Thank you @meduzen 🙇 
Good catch!

PR #1360, opened 2024-04-13
add raytrace gamut mapping method

romainmenke:
This changes the gamut mapping method to a newer algorithm developed by @facelessuser 
It will further reduce chroma and preserve lightness and hue better than the current version in the specification. This is favorable because the priorities for CSS in order are : `lightness > hue > chroma`.

It also has other properties which make it more interesting for browsers.

Given our unique context of generating fallback values we have more room to experiment as long as the result is never broken. This change, as far as I can tell, never produces colors that are visually radically different.

Rolling out this changes means that we can gather feedback on the new algorithm. If any unexpected results surface we can undo this change.


.github/bin/license/check-license.mjs
```diff
@@ -62,6 +62,11 @@ const allFiles = (await getFiles('./')).filter((file) => {
 					continue;
 				}
 
+				if (line.startsWith(' * @license MIT')) {
+					// W3C license is ok
```
romainmenke, line 66:
```suggestion
					// MIT license is ok
```

PR #1366, opened 2024-04-18
fix possibly undefined property accessor

### LINKED ISSUE #1365: HTML Select with "multiple" attribute leads to "Cannot read properties of undefined"
### Reproduction link

_No response_

### Bug description

In https://github.com/csstools/postcss-plugins/blob/main/plugins/css-blank-pseudo/src/browser.js#L52, the code assumes that `element.selectedIndex` is always the index of an element in the options array. 

When using a select field with attribute `multiple`, [this index can be `-1` though](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex). This leads to an `Uncaught TypeError: Cannot read properties of undefined (reading 'value')`

**Minimal Example**

```html
<!DOCTYPE html>
<html>
  <body>
    <select multiple="" required="">
      <option value="one">One</option>
      <option value="two">Two</option>
      <option value="three">Three</option>
    </select>

    <script src="https://unpkg.com/css-blank-pseudo@6.0.1/dist/browser-global.js"></script>
    <script>
      cssBlankPseudoInit();
    </script>
  </body>
</html>
```

### Actual Behavior

_No response_

### Expected Behavior

_No response_

### Can you reproduce it with `npx @csstools/csstools-cli <plugin-name> minimal-example.css`?

None

### `npx` Output

_No response_

### Extra config

_No response_

CSS Blank Pseudo

### Plugin version

6.0.1

_No response_

### Node Version

18

### Validations

HannesT117:
Fixes #1365


romainmenke (approved):
LGTM! 

Thank you so much for this @HannesT117 🎉

PR #1368, opened 2024-04-20
postcss-nesting: nested conditionals

romainmenke:
see : https://github.com/w3c/csswg-drafts/issues/8738#issuecomment-2061777236

```css
::before, ::after {
	color: blue;

	@media screen { color: cyan; }
}
```

Is now equivalent to :

```css
::before, ::after {
	color: blue;

	@media screen {
		@nest {
			color: cyan;
		}
	}
}
```

Where `@nest` represents the parent rule and its selector directly.
There is no wrapping in `:is()`.

`@nest` doesn't take any arguments/params.

[`@nest` is not sure to remain as its value as an author writeable rule is questionable.](https://github.com/w3c/csswg-drafts/issues/10234) Because of this I didn't want to add support for this new `@nest` just yet. We can transpile to unnested CSS without it.


plugins/postcss-nesting/CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changes to PostCSS Nesting
 
+### Unreleased (patch)
+
+- Edition `2024-02`:
+  - Do not rewrite selectors for declarations in conditional rules to a form with `:is()`
```
Antonio-Laguna, line 6: Is this indent intentional?

Also, I assume you want to patch this after the major?

romainmenke: Before, the major.

This is a minimally breaking change.
(Any change theoretically is a breaking change for someone.)

It is only breaking when having something like this :

```css
a, .link {
  @media (min-width: 64rem) {
    color: blue;
  }
}

a {
  color: red;
}
```

Before this change `blue` would have specificity `[0, 1, 0]` because of `:is()` and would overrule `red`.

After this change `blue` will have specificity `[0, 0, 1]` and `red` will have win because it appears later.

So it only affects declarations in nested conditionals that became more specific as a result of `:is()` **and** had lower specificity declarations appearing after it.

User that do depend on the `:is()` behavior would need to opt-in by writing :

```css
a, .link {
  @media (min-width: 64rem) {
    & {
      color: blue;
    }
  }
}

a {
  color: red;
}
```

Antonio-Laguna: Ahh, makes sense!

PR #1369, opened 2024-04-20
`postcss-preset-env` v10

romainmenke:
- [x] remove custom handling of `noIsPseudoSelector` for `postcss-nesting`. This option is no longer relevant for the latest CSS specification.
- [x] https://github.com/csstools/postcss-plugins/pull/1371
- [x] https://github.com/csstools/postcss-plugins/pull/1370
- [x] Improve performance of various helper packages


romainmenke:
@Antonio-Laguna This has been open for some time and I haven't encountered any other spec changes that require a breaking change.

This change would be limited to the changes in nesting and a bump of our minimum node version.

Is there any thing you would still like to change?

If not, I suggest we ship this next week 🚀 
We shipped `9.0.0` around `2023-07`, so that gives us a nice interval.

------

A good explanation for the aspect of nesting that was most recently changed can be found here: https://drafts.csswg.org/css-nesting/#nested-declarations-rule%E2%91%A0

We eagerly added support for this in the `2024-02` edition, even when it had not yet landed in the spec text (https://github.com/csstools/postcss-plugins/pull/1305)

> I am choosing not to re-order declarations and nested rules in this new edition because we actually know that this is a major pain point and I want to give users hands on experience with what it could be if https://github.com/w3c/csswg-drafts/issues/8738 is adopted.

With the recent spec edits our `2024-02` editions aligns with https://drafts.csswg.org/css-nesting/

PR #1370, opened 2024-04-20
`postcss-nesting`: make edition `2024-02` the default

romainmenke:
- make `2024-02` the default in `postcss-nesting`
- update test expect files everywhere
- update the playground as this no longer needs this extra flag


plugins/postcss-nesting/README.md
```diff
@@ -80,43 +82,28 @@ instructions for:
...
-	silenceAtNestWarning: true
-})
-```
-
 ## Options
 
 ### edition
 
-The default behavior is to transpile CSS following an older version of the CSS nesting specification.
+The default behavior is to transpile CSS following the version of the CSS nesting specification as it existed during the last major version of this plugin.
```
Antonio-Laguna, line 89: I'm not sure this sentence is clear enough.

Is there more recent work to the spec? While I know we don't need to explain the specs I feel it doesn't hurt to be more explicit here.

plugins/postcss-nesting/README.md
```diff
@@ -80,43 +82,28 @@ instructions for:
...
 
-If you want to already use the latest version you can set the `edition` option to `2024-02`.
+If you want to use an older or even more recent version you can set the `edition` option.
```
Antonio-Laguna, line 91: We don't have the option right now to do "even more recent"?

Antonio-Laguna (approved):
This LGTM!

I left minor comments around the README for a bit of discussion but this is looking good! Great thing the tests were already separated and they weren't like `_tape-default.mjs`, nice forward thinking!

plugins/postcss-nesting/README.md
romainmenke: I was being vague on purpose but I see know that we need to be clearer on why this is vague.

I didn't want to have to change this sentence each time we ship a new edition or change the default.

Maybe better to explain why we have this mechanic instead.

romainmenke: @Antonio-Laguna I've rewritten this section which more focus on what it is and when users should set this option. Hopefully an improvement :)

https://github.com/csstools/postcss-plugins/blob/postcss-nesting-make-2024-the-default--sympathetic-spadefoot-toad-3bea5e0b0d/plugins/postcss-nesting/README.md#edition

plugins/postcss-nesting/README.md
romainmenke: Correct! I was again trying to future proof these statements, but I agree that this is mostly making it more confusing :)

PR #1421, opened 2024-07-06
Log what browsers require polyfills in preset-env

clarfonthey:
The debug feature will mention which plugins are disabled, but not why certain ones were enabled. I found this particularly useful when deciding how to configure my browserslist, and think it's worth including since it's gated behind a debug mode.

I decided arbitrarily to not print more than five browsers in the list, since it felt like a good "small" number. But I'm open to changing it if desired.


romainmenke (approved):
Thank you @clarfonthey!

I think it's better to log all the browsers.
The first 5 doesn't actually mean anything because the list isn't sorted in a helpful way.

Tests are failing and source wasn't build, so I will merge this into a second branch to fix those things.

romainmenke:
This has been published

PR #1432, opened 2024-07-13
update dependencies

romainmenke:


plugin-packs/postcss-preset-env/src/test/lib/list-features/client-side.mjs
```diff
@@ -1,135 +1,118 @@
...
-			'    ios_saf 15.6-15.8',
-			'    op_mini all',
-			'    op_mob 80',
-			'    opera 111',
-			'    opera 110',
-			'    opera 109',
-			'    safari 17.5',
-			'    safari 17.4',
-			'    samsung 25',
-			'    samsung 24',
```
romainmenke, line 129: This list changes with each update to `browserslist`.
I made the checks a bit for fuzzy so that they only check that something resembling this is present in the logs, but not this exact output.

PR #1555, opened 2025-01-23
Process all declarations when preserve: false

depoulo:
When `{ preserve: false }` is used, there's no use case for disabling this plugin for a single ruleset by providing a manual fallback. This change should not make a difference for almost anyone, except for those who mistakenly (e.g. through preprocessors) declared any duplicate color declarations. Before this change, this plugin would effectively skip any ruleset with duplicate color declarations when `{ preserve: false }` was used.


plugins/postcss-contrast-color-function/test/basic.preserve-false.expect.css
```diff
@@ -0,0 +1,169 @@
...
+}
+
+.max-darkgreen {
+	color: rgb(255, 255, 255);
+}
+
+.manual-fallback {
+	color: white;
+	color: rgb(255, 255, 255);
+}
```
depoulo, line 107: Without the change to index.ts, this would read

```css
.manual-fallback {
	color: white;
	color: contrast-color(darkgreen max);
}
```

which would be quite an unexpected result, as it effectively skips processing.

romainmenke:
Hi @depoulo

Thank you for taking the time to open a pull request.
Unfortunately I am not going to apply this change.
The current behavior is intentional and correct.

When a user adds their own fallback or when there is a relevant `@support` rule we see this as a signal that the user does not want this plugin to transform their CSS.
Especially when `preserve: false` is this important.

Feel free to open an issue if you want to make a case for this change :)

depoulo:
Oooh so your use case is to implicitly opt out of `{ preserve: false }` for a single rule by providing a manual fallback! Might make sense to extend the examples accordingly. In any case, nevermind, and thank you for the plugin!

PR #1587, opened 2025-03-28
rewrite-url docs: Use languageOptions in stylelint config

ehoogeveen-medweb:
Now that Stylelint 16.17.0 is out with support for `declaration-property-value-no-unknown` in `languageOptions`, the suggested config for `postcss-rewrite-url` can be generalized. This commit makes the following changes:

* Use `languageOptions` instead of overriding `declaration-property-value-no-unknown` rule config
* Make syntax consistent with JSON config: Only use double quotes, no trailing commas

Preview: https://github.com/ehoogeveen-medweb/postcss-plugins/blob/rewrite-url-update-docs/plugins/postcss-rewrite-url/README.md#stylelint

I considered adding a fallback for Stylelint <16.17.0 but given that it's a minor version bump I expect people will update (and if they're looking at `postcss-rewrite-url` for the first time, they're probably in a good position to update `stylelint` anyway).

Using the JSON style for the config is a choice I don't feel strongly about - I'm still using it that way in my own projects, but here I just kept it because it seemed consistent with the quoted keys (although the markdown block is technically using language `js`).


ehoogeveen-medweb:
Ah, maybe the preceding text could use an update too, as it currently links to and talks about `declaration-property-value-no-unknown` specifically.

Also, I wonder if it would make sense to do this:
```js
	"types": {
		"rewrite-url()": "rewrite-url( <string> <url-modifier>* )",
		"url": "| <rewrite-url()>"
	}
```

romainmenke:
Thank you for this @ehoogeveen-medweb 🙇 

I think the current change is fine for now :)
Making the type reusable is indeed better but also requires the reader to know a bit more about the syntax definitions.

romainmenke (approved):
Thank you. LGTM 👍

PR #1640, opened 2025-06-25
fix being unable to ignore warnings in postcss-cascade-layers

Fox32:
I tired to ignore the `onRevertLayerKeyword` warnings because they aren't relevant in my case, however setting it to `false` did not disable them. Looking at the code I would have to disable `onImportLayerRule` at the same time, however then `hasAnyLayer` is not evaluated and the plugin exits early.

I modified the code to evaluate both options individually and added a test for it. Including the updated files from the dist folder is correct, right?


Fox32:
It was fun, test setup is nice!

PR #1775, opened 2026-02-04
fix: avoid adding duplicate text-decoration fallback when multiple declarations exist

lukhanina:
- Check if previous declaration has same value before adding fallback
- Skip adding -webkit prefix when following text-decoration exists
- Fixes formatting issues with consecutive text-decoration rules


romainmenke:
https://github.com/csstools/postcss-plugins/pull/1776 has been merged and an update has been released.
Thank you for you help in improving this plugin 🙇

PR #1776, opened 2026-02-04
`postcss-text-decoration-shorthand`: reduce redundant fallbacks

romainmenke:


lukhanina:
@romainmenke , thank you, it makes sense, but there is still 1 minor issue (it's edge case), that when there is a comment next to the css property, e.g.:

```
	text-decoration: underline; /* 2 */
	text-decoration: underline dotted;
```

the condition passed:

```
	if (!next || next.type !== 'decl' || next.prop.toLowerCase() !== 'text-decoration') {
		decl.cloneBefore({
			prop: '-webkit-text-decoration',
			value: nonShortHandValue,
		});
	}

```
because next.type !== 'decl', _next.type === 'comment'_
and webkit prefix is added, so in the end  we get:
```
	-webkit-text-decoration: underline;
	text-decoration: underline; /* 2 */
	-webkit-text-decoration: underline dotted;
	        text-decoration: underline dotted; /* 2 */
```
what is wrong, because without a comment it matches a condition next.prop.toLowerCase() !== 'text-decoration' and doesn't add -webkit-text-decoration: underline; 

so if next.type === 'comment' we need to check next.next().prop.toLowerCase() to make sure it's not text-decoration, before we do clone

PR #1860, opened 2026-06-30
split color conversion for RCS from interpolation

romainmenke:


packages/css-color-parser/src/color-data.ts
```diff
@@ -297,13 +297,6 @@ const predefinedRGB_or_XYZ_Spaces = new Set([
 ]);
 
 export function colorDataTo(colorData: ColorData, toNotation: ColorNotation): ColorData {
-	colorData = {
-		...colorData,
-	};
-
-	// Change any powerless components in src to missing components
-	colorData.channels = convertPowerlessComponentsToMissingComponents(colorData.channels, colorData.colorNotation);
```
romainmenke, line 305: If we only do powerless to none when converting to a different color space then we never have to do it before conversion.

